PR #144, opened 2020-10-14
Centralize docker build instructions into this repository

mcgibbon:
This repository can't update to newer version of fv3gfs-wrapper because of both openmpi and mpich being installed in the image. This PR fixes the issue by keeping all docker build instructions in this repository, instead of building based off of images defined by other repositories. This required a complete overhaul of the docker image structure.

The README changelog further elaborates on these changes and how they are used.

High level summary of changes:
- The build system is now self-contained, and does not use Dockerfiles from other repositories
- fv3core_wrapper now based on fv3core instead of fv3gfs-wrapper
- Feature sharing between fv3core and fv3core_wrapper image build systems
    - CUDA=y/n
    - USE_GT4PY_DEVELOP=y/n
    - dependency caching
    - image tagging
    - sarus integration
- When starting a container, the docker image will install "our" code (including gt4py).
    - For fv3core, this means being able run a command with the latest gt4py develop branch without re-building the image. 
    - For fv3gfs-wrapper, this means it automatically sets itself to use the source code you have saved on disk.
    - This causes a short (~5s for fv3core) / long (~1.5 min for fv3core_wrapper, if you haven't already built) delay when starting the docker container

Changes:
- the "environment" image is now replaced by a "serialbox" and "gt4py" image, and are two of multiple dependency images which can be built, pushed, and pulled
- Docker instructions for the images are now contained entirely in this repository, instructions from fv3gfs-wrapper and fv3gfs-fortran are not relied on for building images
- Only package versions from direct submodules are installed in the docker images
- Replaced `dev_docker.sh` with `make -C docker dev_wrapper` development environment which bind mounts machine sources into the image and performs a build of the sources when entering the image. This prevents the environment from being out-of-date with the files on disk when entering.
- Images now use the entrypoint script defined at `docker/entrypoint.sh` to ensure that up-to-date bind-mounted sources are installed when using the images with bind mounts
- Versions of pypi dependencies are now locked using constraints.txt generated by pip-compile, a tool provided by pip-tools. This automatically pins all dependencies and dependencies of dependencies, avoiding the need to manually update pins and synchronize pins between multiple requirements files.
- docker/Makefile is added which provides make targets for building, pushing, and pulling images, and new logic for image naming and tagging.


Makefile
```diff
@@ -1,9 +1,10 @@
+include docker/Makefile.image_names
+
 GCR_URL = us.gcr.io/vcm-ml
 REGRESSION_DATA_STORAGE_BUCKET = gs://vcm-fv3gfs-serialized-regression-data
 EXPERIMENT ?=c12_6ranks_standard
 FV3CORE_VERSION=0.1.0
```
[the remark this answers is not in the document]
rheacangeo: Thank you! I noticed it is updated in the docker/Makefile.image_names and not here -- we can remove this if it is no longer used. Though the top level Makefile might be where we expect to configure a version number? But the image_names is common to the Makefiles so also makes a lot of sense there.

Makefile
```diff
@@ -43,98 +42,73 @@ DEV_MOUNTS = '-v $(CWD)/$(FV3):/$(FV3)/$(FV3) -v $(CWD)/tests:/$(FV3)/tests -v $
...
+	$(MAKE) -C docker sarus_load_tar
 
 cleanup_remote:
-	gsutil rm $(CORE_BUCKET_LOC)
-	gcloud container images delete -q --force-delete-tags $(FV3_IMAGE)
+	$(MAKE) -C docker cleanup_remote
+
+# end of image build targets which have been moved to docker/Makefile
 
 tests: build
```
rheacangeo, line 113: Not sure what happened, but now using this Makefile gives an error:
 *** missing separator.  Stop
Must be a whitespace complaint of some sort?

docker/Makefile
```diff
@@ -0,0 +1,119 @@
...
+pull_deps_if_needed: ## pull images of any are missing
+	for IMAGE in $(DEPS_IMAGES); \
+	do \
+		if [ -z $(shell docker images -q $(IMAGE)) ]; then \
+			docker pull $$IMAGE; \
+		fi; \
+	done
+
+pull_core_deps_if_needed:
+	DEPS_IMAGES="$(CORE_DEPS_IMAGES)" make pull_deps_if_needed
```
rheacangeo, line 94: There is a formatting error with this one:
/bin/sh: line 0: [: too many arguments

docker/Makefile
```diff
@@ -0,0 +1,119 @@
...
+
+fv3core_wrapper_image: fortran_image
+	FV3CORE_TARGET=fv3core_wrapper FV3CORE_IMAGE=$(FV3CORE_WRAPPER_IMAGE) $(MAKE) fv3core_image
+
+build_deps: build_core_deps ## build container images of all dependencies
+	docker build -f $(DEPENDENCIES_DOCKERFILE) -t $(FMS_IMAGE) $(BUILD_ARGS) --target fv3gfs-fms ../external/fv3gfs-fortran
+	docker build -f $(DEPENDENCIES_DOCKERFILE) -t $(ESMF_IMAGE) $(BUILD_ARGS) --target fv3gfs-esmf ../external/fv3gfs-fortran
+
+build_core_deps: ## build container images of dependencies
+	docker build -f $(DEPENDENCIES_DOCKERFILE) -t $(ENVIRONMENT_IMAGE) $(BUILD_ARGS) --target fv3core-environment .
```
rheacangeo, line 63: I appreciate this being separated out! Although in theory docker caching should never rerun these when it has already been done, I keep running into cases where that happens, so appreciate this for now. We may decide later that pulling the image takes just as long as building and reverse our decision, but I like this here. Though, building fv3core still took 3 minutes on my machine :P. So perhaps it was more due to the pip installs than the apt-gets. I am ok with moving on from this for now, as it all seems quite reasonable.

rheacangeo (commented):
Thanks for being flexible and adjusting to my feedback! I appreciate it a lot. I think something is funky with the top level Makefile right now so am not able to retest everything... it's probably an easy fix.

Makefile
mcgibbon: Thanks! Don't know how I didn't see this.

docker/Makefile
mcgibbon: Fixed!

docker/Makefile
mcgibbon: I was glad to figure out a way that enabled doing it with a lot of make target re-use! Without it my main concern was too much copy-paste.

Yeah the pip installs do take a while. 3 minutes is similar to the fv3net image. I can put them in the environment image, but it means you have to re-build the environment image if you update the pip installs (including what version of fv3config and mpi4py are being used). I think the current approach will save time compared to that when local image caches are considered, but let me know if we should change it.

Makefile
```diff
@@ -43,98 +41,73 @@ DEV_MOUNTS = '-v $(CWD)/$(FV3):/$(FV3)/$(FV3) -v $(CWD)/tests:/$(FV3)/tests -v $
...
+	fi
+	$(MAKE) -C docker fv3core_wrapper_image
 
 pull_environment_if_needed:
-	if [ -z $(shell docker images -q $(FV3_INSTALL_IMAGE)) ]; then \
-		docker pull $(FV3_INSTALL_IMAGE); \
-	fi
+	$(MAKE) -C docker pull_core_deps_if_needed
 
 pull_wrapped_environment_if_needed:
```
rheacangeo, line 82: this one now has an error:
/bin/bash: line 0: [: too many arguments ...

mcgibbon: Fixed, thanks!

rheacangeo:
Ok great, I think things are working. One thing I haven't tried to do though is run the wrapped baroclinic case using these new images. @oelbert have you tried it?

rheacangeo:
Ok, so sorry, but dev_wrapper seemed to have an issue just now. I'll cleanup and try again in case it's due to images leftover from a previous test, or building gt4py when it should be pulled from an existing accepted image, but dev_wrapper seemed to hang for a long time, then had the error
    raise VersionConflict(dist, req).with_context(dependent_req)
pip._vendor.pkg_resources.ContextualVersionConflict: (attrs 20.2.0 (/usr/local/lib/python3.7/dist-packages), Requirement.parse('attrs~=19.1'), {'gt4py'})

rheacangeo:
Phew, ok starting fresh fixed that issue. The next one is that cmake is not in the the wrapped image, so we cannot build a different Serialbox version on the fly like the README indicates... That itself was a hack to get around the fact that the baroclinic test data was generated using serialbox 2.6.1. This will be less of a problem if we generate data using the submodule in this repo. Though, it's always possible somebody will want to use some serialized data generated in some fashion with a mismatched version. 
To avoid that issue I tried running the fv3core_test script, but that may be missing some features we added to the baroclinic test case, it resulted in the error
  File "lib/_wrapper.pyx", line 1471, in fv3gfs.wrapper._wrapper.get_state
fv3gfs.util._exceptions.InvalidQuantityError: Quantity cloud_fraction does not exist - is there a typo?

I assumed master could run these, but it was not a requirement of getting the wrapper into the repo. Is the plan to then iterate on consolidating the run script and getting it working again after this PR is merged?

rheacangeo:
Yay! The baroclinic_test.py DOES work if I replace the baroclinic data with standard data. We only need the data for the grid anyways (unless we want to compare to the unit test regressions). So we have a workaround to the workaround if needed. But, it might be worthwhile considering whether we want to install cmake in this image or not.

rheacangeo (approved):
Great job! Thanks for incorporating feedback!

mcgibbon:
And thanks @rheacangeo for all the legwork testing out the features! Hopefully we can add some tests soon for the wrapper to automate at least some of it.

PR #158, opened 2020-10-27
Implement additional GPU validation fixes

eddie-c-davis:


fv3core/stencils/a2b_ord4.py
```diff
@@ -290,8 +290,7 @@ def extrapolate_corner_qout(qin, qout, i, j, kstart, nk, corner):
         qin[i + i3a, j + j3a, kslice],
         qin[i + i3b, j + j3b, kslice],
     )
-    r3 = 1.0 / 3.0
-    qout[i, j, kslice] = (ec1 + ec2 + ec3) * r3
+    qout[i, j, kslice] = (ec1 + ec2 + ec3) * (1.0 / 3.0)
```
rheacangeo, line 293: That's great this works. It was written this way to get better validation matching the fortran (as it does changing the numbers ever so slightly). But if this is passing the tests, they A+.

fv3core/stencils/remap_profile.py
```diff
@@ -297,23 +294,17 @@ def set_inner_as_kord9(
...
-        elif extm and extm[0, 0, 1]:
-            a4_2 = a4_1
-            a4_3 = a4_1
-            a4_4 = 0.0
-        elif extm > 0.0 and (qmin > 0.0 and a4_1 < qmin):
+        if (
+            (extm != 0.0 and extm[0, 0, -1] != 0.0)
+            or (extm != 0.0 and extm[0, 0, 1] != 0.0)
+            or (extm > 0.0 and (qmin > 0.0 and a4_1 < qmin))
+        ):
```
rheacangeo, line 301: sweet!

fv3core/stencils/remap_profile.py
```diff
@@ -297,23 +294,17 @@ def set_inner_as_kord9(
...
             a4_2 = a4_1
             a4_3 = a4_1
             a4_4 = 0.0
         else:
-            diff_23 = a4_2 - a4_3
             a4_4 = 6.0 * a4_1 - 3.0 * (a4_2 + a4_3)
-            if abs(a4_4) > abs(diff_23):
+            if abs(a4_4) > abs(a4_2 - a4_3):
```
rheacangeo, line 307: I'm so glad we can do this now

tests/test_translate.py
```diff
@@ -18,6 +18,9 @@
 
 OUTDIR = os.path.join(os.path.dirname(os.path.realpath(__file__)), "output")
 GPU_MAX_ERR = 1e-10
+GPU_NEAR_ZERO = 1e-15
+
+_near_zero = 1e-18
```
rheacangeo, line 23: I think I'd be ok with changing the near0 for all of them to 1e-15. 1e-18 was somewhat arbitrary (was based on the differences I was seeing with the numpy backend), while we can reason about 1e-15 being at the floating point limit. But, it also makes sense we'd have different thresholds for GPU, so adding this capability is good.

tests/translate/translate_remap_profile_2d.py
```diff
@@ -24,6 +24,7 @@ def __init__(self, grid):
             "a4_3": {"serialname": "q4_3", "istart": 0, "iend": grid.ie - 2},
             "a4_4": {"serialname": "q4_4", "istart": 0, "iend": grid.ie - 2},
         }
+        self.ignore_near_zero_errors = {"q4_4": True}
```
rheacangeo, line 27: interesting that it's just one of them

rheacangeo (approved):
You haven't asked for a review yet, so I'm sorry for jumping the gun. But this looks like good stuff!

fv3core/stencils/a2b_ord4.py
eddie-c-davis: Thank you, was just trying to simplify the code.

fv3core/stencils/remap_profile.py
eddie-c-davis: Yes, the conditionals have come a long way!

tests/translate/translate_remap_profile_2d.py
eddie-c-davis: Agreed!

eddie-c-davis:
Thank you for the preemptive review! There will be more GPU fixes to come, but I wanted to get these ones in before they get lost in the shuffle.

PR #160, opened 2020-11-11
Fixes for compile-time regions

jdahm:
This is meant to be tested with a new `VCM/gt4py:develop` branch.

* Removes `splitters` in calls
* `__splitters__` are merged into `__externals__`
* Functions that use the axis offsets are now undecorated
* Axis offsets are set to None if not on the processor owned grid


rheacangeo (commented):
Looks good to me! as long as the tests pass!

rheacangeo (approved):
Looks great to me! As long as you get it passing in your VM, it's good to go!

PR #162, opened 2020-11-23
Reviving the FV3_TAG behavior 

rheacangeo:
Previously we used FV3_TAG so that different PRs and builds in jenkins would be distinctly identified in google storage and the GCR, and then would be cleaned up when tests passed. This was removed with the restructuring of the infrastructure to support running the wrapper and the full model, and now jenkins is deleting resources that are not specific to a PR and can collide with each other. This PR tries to get that behavior back while preserving the features that are needed.


rheacangeo:
launch jenkins

rheacangeo:
launch jenkins

rheacangeo:
launch jenkins

Makefile
```diff
@@ -3,7 +3,6 @@ include docker/Makefile.image_names
 GCR_URL = us.gcr.io/vcm-ml
 REGRESSION_DATA_STORAGE_BUCKET = gs://vcm-fv3gfs-serialized-regression-data
 EXPERIMENT ?=c12_6ranks_standard
-FV3CORE_VERSION=0.1.0
```
rheacangeo, line 6: This is no longer used anywhere so deleted

docker/Makefile.image_names
```diff
@@ -37,8 +37,9 @@ SERIALBOX_IMAGE ?= $(GCR_URL)/serialbox-build:$(TAG_NAME)
 ENVIRONMENT_IMAGE ?= $(GCR_URL)/fv3core-environment:$(TAG_NAME)
 
 # built image names
+FV3_TAG ?=
 FORTRAN_IMAGE = $(GCR_URL)/$(FORTRAN_NAME):$(TAG_NAME)
 WRAPPER_IMAGE = $(GCR_URL)/$(WRAPPER_NAME):$(TAG_NAME)
```
mcgibbon, line 42: These may also want the jenkins tag appended. If not, maybe a `#` comment as to why not.

mcgibbon (approved):
Looks good! Thanks for the fast fix! The one change I ask is that `FV3_TAG` be changed to `JENKINS_TAG`, since it's really a tag of the jenkins job. Otherwise it's unclear what the difference in purpose is between TAG_NAME and FV3_TAG are, just from reading the makefiles. There's also an optional change.

rheacangeo:
launch jenkins

rheacangeo:
trying again because got new strange daint error that seems unrelated to the code change
Failed to retrieve username for uid=21028 from "/apps/daint/system/opt/sarus/1.1.0/etc/passwd"

PR #163, opened 2020-11-30
quick fix to be able to build fv3core again after pip changes broke it

rheacangeo:
There have been deprecation warnings for months about breaking pip install changes, and today they finally happened and broke our build in several ways. Here is a quick step to getting this repo usable again. 
We can probably remove the legacy-resolver once we update gt4py to not require pytest5.2 and attrs19.1

Also added a step to the Makefile constraints.txt endpoint to remove the git+https lines from the constraints files (egg=name does not seem to fix the error). pip-compile will probably do this for us in the near future I am guessing, at which point we can remove this line. Or maybe we will learn another way to handle it. 

Also removed some Makefile endpoints that appeared to be duplicates, likely from some merging issue.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -51,14 +51,12 @@ update_submodules:
 
 constraints.txt: requirements.txt requirements_wrapper.txt requirements_lint.txt
 	pip-compile $^ --output-file constraints.txt
-
+	sed -i '' '/^git+https/d' constraints.txt
 # Image build instructions have moved to docker/Makefile but are kept here for backwards-compatibility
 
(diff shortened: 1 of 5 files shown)


mcgibbon (approved):
Looks good. Is `--use-deprecated=legacy-resolver` needed in the final PR or was this an intermediate solution you tried that can be removed now?

rheacangeo:
It's an intermediate solution until we use a new gt4py build, so was thinking removing it would be a separate PR.

PR #165, opened 2020-12-02
remove pip legacy-resolver

rheacangeo:
we do not need the legacy-resolver anymore now that gt4py pytest and attrs versions are updated. We did need to update attrs 20.2 to 20.3 so there was not a mismatch. We will likely encounter more version issues in the future unless we do more, but for this PR let's just get rid of the legacy-resolver so we at least find out next time this happens.


jdahm (approved):
Looks good to me!

PR #166, opened 2020-12-04
Enable axis offsets relative to endpoints

jdahm:
This makes fv3core compatible with the new develop branch. This includes:

* A rewrite of the decorator, removing our internal cache (which was broken: it did not take into account externals). This now returns an object instead.
* Requiring origin and domain on all stencil calls. This may be temporary.
* Expresses axis offset externals relative to the beginning and now also the end of axes.

Test: https://jenkins.ginko.ch/job/fv3core-build-for-daint_PR/436/

### DIFF
--- a/README.md
+++ b/README.md
@@ -125,7 +125,7 @@ common options for our tests, which you can add to `TEST_ARGS`:
 **NOTE:** FV3 is current assumed to be by default in a "development mode", where stencils are checked each time they execute for code changes (which can trigger regeneration). This process is somewhat expensive, so there is an option to put FV3 in a performance mode by telling it that stencils are immutable:
 
 ```shell
-$ export FV3_IMMUTABLE_STENCILS=True
+$ export FV3_STENCIL_REBUILD_FLAG=True
 ```
 
 ## Porting a new stencil
--- a/README.md
+++ b/README.md
@@ -69,7 +69,7 @@ $ make dev_tests_mpi
 ```
 These will mount your current code into the fv3core container and run it rather than the code that was built when `make build` ran.
 
-## Running tests  inside a container
+## Running tests inside a container
 
 If you to prefer to work interactively inside the fv3core container, get the test data and build the docker image:
 ```shell
@@ -122,6 +122,11 @@ common options for our tests, which you can add to `TEST_ARGS`:
 
 * `-m` - will let you run only certain groups of tests. For example, `-m=parallel` will run only parallel stencils, while `-m=sequential` will run only stencils that operate on one rank at a time.
 
+**NOTE:** FV3 is current assumed to be by default in a "development mode", where stencils are checked each time they execute for code changes (which can trigger regeneration). This process is somewhat expensive, so there is an option to put FV3 in a performance mode by telling it that stencils are immutable:
+
+```shell
+$ export FV3_IMMUTABLE_STENCILS=True
+```
 
 ## Porting a new stencil
 
--- a/README.md
+++ b/README.md
@@ -122,10 +122,10 @@ common options for our tests, which you can add to `TEST_ARGS`:
 
 * `-m` - will let you run only certain groups of tests. For example, `-m=parallel` will run only parallel stencils, while `-m=sequential` will run only stencils that operate on one rank at a time.
 
-**NOTE:** FV3 is current assumed to be by default in a "development mode", where stencils are checked each time they execute for code changes (which can trigger regeneration). This process is somewhat expensive, so there is an option to put FV3 in a performance mode by telling it that stencils are immutable:
+**NOTE:** FV3 is current assumed to be by default in a "development mode", where stencils are checked each time they execute for code changes (which can trigger regeneration). This process is somewhat expensive, so there is an option to put FV3 in a performance mode by telling it that stencils should not automatically be rebuilt:
 
 ```shell
-$ export FV3_STENCIL_REBUILD_FLAG=True
+$ export FV3_STENCIL_REBUILD_FLAG=False
 ```
 
 ## Porting a new stencil
--- a/fv3core/decorators.py
+++ b/fv3core/decorators.py
@@ -112,18 +112,15 @@ class StencilDataCache(collections.abc.Mapping):
         """In-memory cache of the data pickled to disk."""
 
     def _get_cache_filename(self, stencil: gt4py.StencilObject) -> str:
-        try:
-            pymodule_filename = stencil._file_name
-            return f"{os.path.splitext(pymodule_filename)[0]}_{self.extension}"
-        except AttributeError:
-            raise ValueError("_file_name attribute not included in stencil object")
+        pymodule_filename = stencil._file_name
+        return f"{os.path.splitext(pymodule_filename)[0]}_{self.extension}"
 
     def __getitem__(self, stencil: gt4py.StencilObject):
         key = hash(stencil)
         if key not in self.cache:
             filename = self._get_cache_filename(stencil)
             if not os.path.exists(filename):
-                raise ValueError(f"Cannot open cache file {filename}")
+                raise KeyError(f"Cache file {filename} does not exist")
             self.cache[key] = pickle.load(open(filename, mode="rb"))
         return self.cache[key]
 
@@ -165,7 +162,7 @@ class FV3StencilObject:
         self.backend_kwargs: Dict[str, Any] = kwargs
         """Remainder of the arguments are assumed to be gt4py compiler backend options."""
 
-        self._build_info_cache: StencilDataCache = StencilDataCache("build_info.py")
+        self._axis_offsets_cache: StencilDataCache = StencilDataCache("axis_offsets.p")
 
     @property
     def built(self) -> bool:
@@ -173,9 +170,9 @@ class FV3StencilObject:
         return self.stencil_object is not None
 
     @property
-    def build_info(self) -> Dict[str, Any]:
-        """The build_info dictionary."""
-        return self._build_info_cache[self.stencil_object]
+    def axis_offsets(self) -> Dict[str, Any]:
+        """AxisOffsets used in this stencil."""
+        return self._axis_offsets_cache[self.stencil_object]
 
     def __call__(self, *args, origin: Index3D, domain: Index3D, **kwargs) -> None:
         """Call the stencil, compiling the stencil if necessary.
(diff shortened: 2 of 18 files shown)


setup.cfg
```diff
@@ -17,3 +17,5 @@ use_parentheses = true
 lines_after_imports = 2
 default_section = THIRDPARTY
 sections = FUTURE,STDLIB,THIRDPARTY,FIRSTPARTY,LOCALFOLDER
+known_first_party = fv3core
+known_third_party = f90nml,pytest,xarray,numpy,mpi4py,gt4py
```
jdahm, line 21: @mcgibbon I'm not sure if we need all these. Can you check them?

jdahm: Some of those may only be used through sub-packages.

fv3core/utils/grid.py
```diff
@@ -457,3 +434,45 @@ def slice_data_k(self, ki):
         # update instance vars
         for k, v in self.data_fields.items():
             setattr(self, k, self.data_fields[k])
+
+
+def axis_offsets(
+    grid: Grid,
```
jdahm, line 440: Shall we rewrite this to use the partitioner in this PR?

fv3core/stencils/riem_solver3.py
```diff
@@ -129,9 +129,9 @@ def compute(
     shape = w.shape
     domain = (grid.nic, grid.njc, km + 2)
     riemorigin = (grid.is_, grid.js, 0)
-    dm = copy(delp)
-    cp3 = copy(cappa)
-    pe_init = copy(pe)
+    dm = copy(delp, origin=grid.default_origin())
+    cp3 = copy(cappa, origin=grid.default_origin())
+    pe_init = copy(pe, origin=grid.default_origin())
```
rheacangeo, line 134: default_origin is (0, 0, 0) and since you've modified the copy method to choose this if origin is not specified, you don't need this change. But, it doesn't hurt either.

fv3core/stencils/riem_solver_c.py
```diff
@@ -75,9 +75,9 @@ def compute(ms, dt2, akap, cappa, ptop, hs, w3, ptc, q_con, delpc, gz, pef, ws):
     shape = w3.shape
     domain = (spec.grid.nic + 2, grid.njc + 2, km + 2)
     riemorigin = (is1, js1, 0)
-    dm = copy(delpc)
-    cp3 = copy(cappa)
-    w = copy(w3)
+    dm = copy(delpc, origin=grid.default_origin())
+    cp3 = copy(cappa, origin=grid.default_origin())
+    w = copy(w3, origin=grid.default_origin())
```
rheacangeo, line 80: same as last comment

fv3core/decorators.py
```diff
@@ -90,120 +92,118 @@ def wrapped(state, *args, **kwargs):
...
-            else:
-                raise TypeError(
-                    f"Type of the origin argument was {type(origin_arg)}. Expected Sequence or Mapping."
-                )
-
-            return origin
+    @property
+    def def_ir(self) -> Optional[dict]:
+        """Current stencil definition IR if built, else None."""
+        return self.build_info.def_ir
```
rheacangeo, line 135: adding this for debugging purposes?

rheacangeo (commented):
Nice job sleuthing the problem and coming up with a solution. Looks pretty exciting! Looks like it relies on the gt4py side caching instead of adding out own? 

We'll want to make sure this doesn't disrupt any of the workflows that don't get tested automatically -- e.g. will this still work when integrated with the wrapper? I don't expect it wouldn't though, since these changes shouldn't impact that interface, I think. We should make sure changing rebuild, backend etc work and whatever else we can think of. 
Looks like it fails jenkins with a   TypeError: unsupported operand type(s) for +: '_AxisOffset' and 'int' for pk3_halo. I believe you made gt4py side changes that fix this. Rebuilding develop seems to have not worked to update that side for the tests here to work. The error is:
/bin/python3: No module named pytest
did something get borked in the merge to lose pytest?

jdahm:
> The error is:
> /bin/python3: No module named pytest
> did something get borked in the merge to lose pytest?

Thanks for catching that. It should be rebuilding a new develop branch now.

fv3core/decorators.py
jdahm: Yeah, that future data flow analysis!

cc: @eddie-c-davis

fv3core/stencils/riem_solver3.py
jdahm: Good point. I'll remove those!

setup.cfg
mcgibbon: Even if we never import them (we shouldn't import mpi4py directly so the package can run in serial without MPI, and you should probably remove that one), there's no real harm to adding extra packages to this list. From what I can tell, it's only used for isort.

I would consider `fv3gfs` (e.g. `fv3gfs.util`) a first-party package.

fv3core/stencils/basic_operations.py
```diff
@@ -20,18 +21,24 @@ def copy_stencil(q_in: sd, q_out: sd):
         q_out = q_in
 
 
-def copy(q_in, origin=(0, 0, 0), domain=None):
+def copy(q_in, origin=None, domain=None):
```
mcgibbon, line 24: `origin=(0, 0, 0)` is actually safe (since tuples are immutable) and clearer, so long as we're restricting q_in to be a 3-dimensional field. Domain on the other hand has to default to None since its default value is dynamic.

fv3core/decorators.py
```diff
@@ -90,120 +92,116 @@ def wrapped(state, *args, **kwargs):
...
+        new_build_info = {}
+        stencil_object = gtscript.stencil(
+            definition=self.func, build_info=new_build_info, **stencil_kwargs
+        )
+        if stencil_object != self.stencil_object:
+            self.build_info = new_build_info
+            self.stencil_object = stencil_object
+
+        # Call it
+        kwargs["validate_args"] = kwargs.get("validate_args", utils.validate_args)
```
mcgibbon: There's still room to optimize this in a later PR after more long-term design of compile time argument management. For example, if we store these in a container that keeps a boolean indicating whether the values have changed since last compilation or not, which gets updated when the values change, we can avoid re-creating these objects every call.

fv3core/decorators.py
```diff
@@ -90,120 +92,116 @@ def wrapped(state, *args, **kwargs):
...
+            **self.backend_kwargs,
+        }
 
-        return wrapped
+        # Generate stencil if out of date
+        new_build_info = {}
+        stencil_object = gtscript.stencil(
+            definition=self.func, build_info=new_build_info, **stencil_kwargs
+        )
+        if stencil_object != self.stencil_object:
```
mcgibbon, line 173: I assume this is a fairly lightweight operation which only involves computing a new hash?

mcgibbon (approved):
Looks good! Nothing critical. A few comments with suggested changes.

fv3core/decorators.py
jdahm: Type comparison on the generated stencils (each is unique).

jdahm:
launch jenkins

fv3core/utils/typing.py
```diff
@@ -41,3 +41,5 @@ def _FieldDescriptorMaker(dtype):
 FloatField = _FieldDescriptor(float_type)
 IntField = _FieldDescriptor(int_type)
 BoolField = _FieldDescriptor(bool_type)
+
+Int3 = Tuple[int, int, int]
```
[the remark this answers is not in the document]
mcgibbon: bump

setup.cfg
mcgibbon: `fv3gfs` still needs to be added to first party imports.

README.md
```diff
@@ -122,6 +122,11 @@ common options for our tests, which you can add to `TEST_ARGS`:
 
 * `-m` - will let you run only certain groups of tests. For example, `-m=parallel` will run only parallel stencils, while `-m=sequential` will run only stencils that operate on one rank at a time.
 
+**NOTE:** FV3 is current assumed to be by default in a "development mode", where stencils are checked each time they execute for code changes (which can trigger regeneration). This process is somewhat expensive, so there is an option to put FV3 in a performance mode by telling it that stencils are immutable:
+
+```shell
+$ export FV3_IMMUTABLE_STENCILS=True
```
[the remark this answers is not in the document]
mcgibbon: Is there a better descriptor than "Immutable stencils"? This implies to me stencils can't be changed once created. What it really means is load stencils from the cache without checking the code is unchanged since it was cached. Maybe `FV3_DISABLE_CACHE_VALIDATION` (or something shorter)? It also has a scarier name that helps make sure you understand why and when you might get silent errors.

fv3core/decorators.py
```diff
@@ -90,120 +94,197 @@ def wrapped(state, *args, **kwargs):
...
+
+    @property
+    def def_ir(self) -> gt_ir.StencilDefinition:
+        """Return the definition IR."""
+        return self.build_info["def_ir"]
+
+    @property
+    def impl_ir(self) -> gt_ir.StencilImplementation:
+        """Return the implementation IR."""
+        return self.build_info["iir"]
```
mcgibbon, line 197: Is this necessary when we already have e.g. `stencil.build_info["def_ir"]`?

mcgibbon (changes requested):
There seems to be some dead code in FV3StencilObject around handling the cache (it appears gt4py handles it instead). The API of FV3StencilObject appeared at first to have changed significantly, holding multiple gt4py stencils instead of one, but it looks like that's only because of the dead code and it should be fine once that's removed.

Run timing should be put in another PR and might be appropriate to unit test depending on what it's being used for (is the runtime nonzero, and is call_run time greater than run time).

jdahm:
> fv3gfs still needs to be added to first party imports.

That changes the imports in a lot of files. I changed that in the `enable-flake8` PR.

jdahm:
> Is this necessary when we already have e.g. `stencil.build_info["def_ir"]`?

I prefer nested namespaces to dicts, but true, using a dict matches the way gt4py is exposing it. I'll make the change.

mcgibbon (approved):
Approving since we went over the PR and you addressed all the comments.

jdahm:
This is getting manually tested here: https://jenkins.ginko.ch/job/fv3core-build-for-daint_PR/436/

PR #168, opened 2020-12-15
standalone from serialbox

twicki:
## Key Changes
The goal of this PR is to create a standalone application to run the dynamics for timing purposes bare metal on daint.
In order to facilitate that we have:
- A submit script on daint
- Timing and plotting infrastructure
- The standalone application

## Additional Changes:
- In the process a broken stencil was fixed
- We moved relevant pieces from the testing folders to expose them to the rest of the framework


rheacangeo (commented):
Looks like there is an empty 'nightly_bench' file here, is that on purpose?

examples/standalone/benchmarks/plotting.py
```diff
@@ -0,0 +1,52 @@
...
+        if filename.endswith(".json"):
+            print(filename)
+            fullpath = args.data_dir + "/" + filename
+            with open(fullpath) as f:
+                alldata.append(json.load(f))
+    alldata.sort(
+        key=lambda k: datetime.strptime(
+            k["setup"]["experiment time"], "%d/%m/%Y %H:%M:%S"
+        )
+    )
```
rheacangeo, line 32: nice

examples/standalone/benchmarks/plotting.py
```diff
@@ -0,0 +1,52 @@
...
+
+    # simple plots for the history
+    for data in ["main", "init", "cleanup", "total"]:
```
rheacangeo, line 35: is the plan to expand 'main' to include finer scale metrics matching what Oli added to the Fortran code?

examples/standalone/benchmarks/run_on_daint.sh
```diff
@@ -0,0 +1,90 @@
...
+
+exitError()
+{
+    echo "ERROR $1: $3" 1>&2
+    echo "ERROR     LOCATION=$0" 1>&2
+    echo "ERROR     LINE=$2" 1>&2
+    exit $1
+}
+
+SCRIPT=`realpath $0`
```
rheacangeo, line 21: This is for daint only (run_on_daint), so this is fine, just want to note this isn't present on Macs. But what would the point be of benchmarking on our macs? hahahha

examples/standalone/benchmarks/run_on_daint.sh
```diff
@@ -0,0 +1,90 @@
...
+  then
+    target_dir="$ROOT_DIR"
+fi
+
+data_path="$5"
+if [ -z "$5" ]
+  then
+    data_path="/project/s1053/fv3core_serialized_test_data/7.0.0/c12_6ranks_standard/"
+fi
+
```
rheacangeo, line 51: It might be helpful to echo this information so that it shows up in the log before the job is launched. Maybe it does already by the act of running it?

examples/standalone/benchmarks/run_on_daint.sh
```diff
@@ -0,0 +1,90 @@
...
+target_dir="$4"
+if [ -z "$4" ]
+  then
+    target_dir="$ROOT_DIR"
+fi
+
+data_path="$5"
+if [ -z "$5" ]
+  then
+    data_path="/project/s1053/fv3core_serialized_test_data/7.0.0/c12_6ranks_standard/"
```
rheacangeo, line 49: I didn't quite understand why we are using 7.0.0 rather than 7.2.3? You said the data volume is smaller, but it shouldn't have to load all of the data? But as a default data_path, that is fine, no objection.

examples/standalone/benchmarks/run_on_daint.sh
```diff
@@ -0,0 +1,90 @@
...
+cd $ROOT_DIR
+rm -rf vcm_1.0
+
+echo "copying in the venv"
+cp -r /project/s1053/install/venv/vcm_1.0/ .
+git submodule update --init --recursive
+echo "install requirements"
+vcm_1.0/bin/python -m pip install external/fv3gfs-util/
+vcm_1.0/bin/python -m pip install .
+vcm_1.0/bin/python -m pip install gitpython
```
rheacangeo, line 63: Hopefully we can shortly replace this with the daint_venv install script, but this works and shouldn't block this PR>

examples/standalone/benchmarks/run_on_daint.sh
```diff
@@ -0,0 +1,90 @@
...
+
+# set up the experiment data
+cp -r $data_path test_data
+tar -xf test_data/dat_files.tar.gz -C test_data
+cp test_data/*.yml test_data/input.yml
```
rheacangeo, line 68: the data is not already expanded?

examples/standalone/benchmarks/run_on_daint.sh
```diff
@@ -0,0 +1,90 @@
...
+# Adapt batch script:
+cp buildenv/submit.daint.slurm .
+sed s/\<NAME\>/standalone/g submit.daint.slurm -i
+sed s/\<NTASKS\>/$ranks/g submit.daint.slurm -i
+sed s/\<NTASKSPERNODE\>/$ranks/g submit.daint.slurm -i
+sed s/\<CPUSPERTASK\>/1/g submit.daint.slurm -i
+sed s/#SBATCH\ --output=\<OUTFILE\>//g submit.daint.slurm -i
+sed s/00:45:00/01:30:00/g submit.daint.slurm -i
+sed s/\<G2G\>//g submit.daint.slurm -i
+sed -i "s#<CMD>#export PYTHONPATH=/project/c14/install/daint/serialbox2_master/gnu/python:\$PYTHONPATH\nsrun vcm_1.0/bin/python examples/standalone/runfile/dynamics.py test_data/ $timesteps $backend#g" submit.daint.slurm
```
rheacangeo, line 84: project c14?

examples/standalone/runfile/dynamics.py
```diff
@@ -0,0 +1,142 @@
...
+    # get grid from serialized data
+    grid_savepoint = serializer.get_savepoint("Grid-Info")[0]
+    grid_data = {}
+    grid_fields = serializer.fields_at_savepoint(grid_savepoint)
+    for field in grid_fields:
+        grid_data[field] = serializer.read(field, grid_savepoint)
+        if len(grid_data[field].flatten()) == 1:
+            grid_data[field] = grid_data[field][0]
+    grid = fv3core.testing.TranslateGrid(grid_data, rank).python_grid()
+    spec.set_grid(grid)
```
rheacangeo, line 88: might be helpful to put into a separate method, but I can see arguments for keeping it inlined

fv3core/stencils/basic_operations.py
```diff
@@ -89,7 +89,11 @@ def subtract_term_stencil(in1: sd, out: sd):
 
 
 @gtstencil()
-def multiply_constant(in1: sd, in2: float, out: sd):
+def multiply_constant(
+    in1: sd,
+    out: sd,
+    in2: float,
+):
```
rheacangeo, line 96: good catch! thank you x86!

fv3core/testing/parallel_translate.py
```diff
@@ -0,0 +1,200 @@
+import copy
```
rheacangeo, line 1: oh yeah, basically just moved

rheacangeo (commented):
This is looking great. I may have commented on parts of the code you said you were working on, I apologize if so.

examples/standalone/benchmarks/README.md
```diff
@@ -0,0 +1,19 @@
+# How to get performance numbers
+
+## Daint
+The run_on_daint script gets arguments:
+- Timesteps
+- Ranks
+- backend
+- (target directory)
+- (data directory)
+it will output a json file with times into the target dir. The first timestep is counted in init, the rest in main loop.
```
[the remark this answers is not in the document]
twicki: done

examples/standalone/benchmarks/plotting.py
```diff
@@ -0,0 +1,52 @@
...
+
+
+if __name__ == "__main__":
+    usage = "usage: python %(prog)s <data_dir>"
+    parser = ArgumentParser(usage=usage)
+    parser.add_argument(
+        "data_dir",
+        type=str,
+        action="store",
+        help="directory containing data to run with",
```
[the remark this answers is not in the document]
twicki: clarified

examples/standalone/benchmarks/plotting.py
twicki: yeah, this is what @oelbert is working on

examples/standalone/benchmarks/run_on_daint.sh
twicki: good to know

examples/standalone/benchmarks/run_on_daint.sh
twicki: done

examples/standalone/benchmarks/run_on_daint.sh
twicki: The initial idea was that the jenkins plan passes the path in. We can change that here though if you'd prefer?

examples/standalone/benchmarks/run_on_daint.sh
twicki: sounds good as a follow up

examples/standalone/benchmarks/run_on_daint.sh
twicki: no, it is still in the archive

examples/standalone/benchmarks/run_on_daint.sh
twicki: I created a copy for us. Freezing it might be a good idea. Good catch

examples/standalone/runfile/dynamics.py
twicki: I don't really see a benefit of abstracting it away here

fv3core/stencils/basic_operations.py
twicki: :+1:

fv3core/testing/parallel_translate.py
twicki: yeah

examples/standalone/benchmarks/run_on_daint.sh
rheacangeo: that makes sense to me

examples/standalone/benchmarks/run_on_daint.sh
rheacangeo: Since it is not a huge number of files, does it make sense to just keep the expanded version and run from that? This in theory takes up less space though, so that's good.

rheacangeo (approved):
Looks great!

rheacangeo:
launch jenkins

examples/standalone/runfile/dynamics.py
mcgibbon: FWIW I think it's fine in a scenario like this (a runfile example) to have code in the long-block-of-text style. But I thought I should point out some of the benefits of what @rheacangeo suggested if it's not clear. These benefits come into play in most coding since it's related to the rule of thumb that you should keep functions short.

One main benefit is if you do it to all the code, it makes it easier to both scan the main block to tell what the script does, and understand the information flow between tasks. e.g. (of course this won't run):
```
args = parse_args()
config = load_config(args)
cube = get_communicator(config["layout"))
state, grid = initialize_state_from_savepoint(config["data_dir"])
fv_dynamics.fv_dynamics(state, ...)
...
```
Doing things like this would also make it very clear e.g. what information is actually necessary to define the grid? The answer is either data_dir or serializer, but it takes some time reading the code to figure that out. When you notice yourself writing a lot of comments to describe the parts of a function (or main block), it's a cue that you may want to move each of those blocks into another named function. This has the added benefit of making it less likely those comments will go out of date.

PR #169, opened 2020-12-16
Run black before flake8

jdahm:
Running black before flake8 fixes many of the errors that flake8 would have flagged, causing you to have to re-run the checks again.


rheacangeo (approved):
Ah, yeah would be good to keep black, didn't realize it had gone away

jdahm:
@rheacangeo It didn't go away, but it was run after flake8. Running it before fixes some errors that flake8 flags.

PR #170, opened 2020-12-16
Update docker images for CUDA compatibility

rheacangeo:
This does a couple of things to make working with the docker images easier, and to avoid the problem that will come up if we don't stop using ubuntu19 next time we 'full_build'/ 
Some of these changes might need to be changed again when we incorporate explicit versioning of the gt4py:develop builds, this is just an incremental step. 
 -- Update Python version to 3.8, using pyenv to make controlling the version easier
 -- Update base image to ubuntu 18.04. This means the only difference in the BASE_IMAGE between CUDA=y and CUDA=n is whether or not cuda is installed in the image. When running on a GCE GPU we need this for the fv3core-environment image and for the fv3core we build
-- Since we aren't building gt4py in the dependencies, we really don't need to rebuild everything every time that updates. Made a new jenkins action we can trigger with the gt4py:develop tests. As a follow on task, not for this PR, we can now run the gt4py gpu tests *within* the fv3core image rather than raw on the gce instance, so the jenkins plans can be updated after this. 
-- Makes use of the latest serialized data, 7.2.3, which was also generated using ubuntu18.04. It used gnu-7, so a future task may be to use gnu-8 but testing revealed no differences in test results using 7 vs 8. (and using the gtx86 and gtmc backends, the same tests failed with the same results). UPDATE -- we are at least using gnu-8 for the fv3core images because gt4py requires it.A new dataset using fv3gfs-fortran will need to be generated if we determine we can't validate and the mentioned tests didn't capture something. 
-- Reorders the dependencies.Dockerfile so that if somebody does not have DOCKER_BUILDKIT set to 1, and they are building main fv3core tasks, not having to do with the fortran images, it should still work. 
-- Removes the CUDA specific tags for all dependent images besides the environment and the main fv3core
-- Testing has not been done on 'build_wrapped' -- this was broken before these changes, and parallel work has been done in another branch to address some of the issues. A follow-on task is to get that updated and all usung ubuntu18 as well. 
-- Makes it so the gtcuda tests can run https://jenkins.ginko.ch/job/fv3core-master-gtcuda/152/backend=gtcuda,experiment=c12_6ranks_standard,runaction=run_regression_tests,slave=gce-1p100/


rheacangeo:
launch jenkins

rheacangeo:
This failed going through sarus on parallel tests... 
boost::filesystem::filesystem_error'\\\\\\\\n  what():  boost::filesystem::remove: Resource busy: \\\\\\\\\\\\\\\"/apps/daint/system/opt/sarus/1.1.0/var/OCIBundleDir/rootfs/lib/libgfortran.so.4\\

rheacangeo:
launch jenkins

Makefile
```diff
@@ -151,13 +153,12 @@ dev_tests_mpi_host:
...
-	$(CONTAINER_ENGINE) run $(RUN_FLAGS) $(VOLUMES) $(MOUNTS) \
+	$(CONTAINER_ENGINE) run $(RUN_FLAGS) $(VOLUMES) $(MOUNTS) $(CUDA_FLAGS) \
 	$(FV3_IMAGE) bash -c "pip list && pytest --data_path=$(TEST_DATA_CONTAINER) $(TEST_ARGS) /$(FV3)/tests"
 
 test_base_parallel:
-	$(CONTAINER_ENGINE) run $(RUN_FLAGS) $(VOLUMES) $(MOUNTS) $(FV3_IMAGE) \
-	$(MPIRUN_CALL) \
-	bash -c "pip list && pytest --data_path=$(TEST_DATA_CONTAINER) $(TEST_ARGS) -m parallel /$(FV3)/tests"
+	$(CONTAINER_ENGINE) run $(RUN_FLAGS) $(VOLUMES) $(MOUNTS) $(CUDA_FLAGS) $(FV3_IMAGE) \
+	bash -c "pip list && $(MPIRUN_CALL) pytest --data_path=$(TEST_DATA_CONTAINER) $(TEST_ARGS) -m parallel /$(FV3)/tests"
```
eddie-c-davis, line 161: Woo, parallel tests!

docker/Makefile.image_names
```diff
@@ -14,26 +15,39 @@ FV3CORE_DOCKERFILE = Dockerfile
...
 	GT4PY_OPTIONALS = ""
 	ENV_CUDA_PATH = ""
 else
-	BASE_IMAGE = nvidia/cuda:10.2-devel-ubuntu18.04
-	TAG_NAME = $(VERSION)-gnu8-mpich314-cuda102-develop
+	BASE_IMAGE_ENV = nvidia/cuda:10.2-devel-ubuntu18.04
+	TAG_NAME = $(VERSION)-gnu7-mpich314-cuda102-develop
 	GT4PY_OPTIONALS = "[cuda102]"
-	ENV_CUDA_PATH = /usr/local/cuda/10.2
+	ENV_CUDA_PATH = /usr/local/cuda/
```
eddie-c-davis, line 39: Nice catch!

docker/Makefile.image_names
```diff
@@ -14,26 +15,39 @@ FV3CORE_DOCKERFILE = Dockerfile
...
+    CUDA_FLAGS =
+endif
+
+
+# base environment images w/ or w/o CUDA
 ifeq ($(CUDA),n)
-	BASE_IMAGE = ubuntu:19.10
-	TAG_NAME = $(VERSION)-gnu9-mpich314-nocuda-develop
+	BASE_IMAGE_ENV = ubuntu:18.04
+	TAG_NAME = $(VERSION)-gnu7-mpich314-nocuda-develop
```
eddie-c-davis, line 32: Should this be `gnu8`?

requirements.txt
```diff
@@ -8,7 +8,7 @@ xarray
 numpy
 zarr
 cftime
-git+https://github.com/mpi4py/mpi4py.git@aac3d8f2a56f3d74a75ad32ac0554d63e7ef90ab
```
eddie-c-davis, line 11: We might need to stick with `mpi4py` master based on experiments, but since we still do not have halo exchanges working on daint, this is probably fine for now.

eddie-c-davis (approved):
Very impressive attention to detail! I learned a lot simply by reviewing, thank you. Some small questions/comments, but overall looks great to me.

rheacangeo:
launch jenkins

requirements.txt
rheacangeo: This was changed because master was breaking the docker build using the cuda image base ... but a lot has changed since then. I can try again to see if this works.

docker/Makefile.image_names
rheacangeo: yes good point, though gfortran-8 inside of the docker image seems to be breaking the parallel tests through sarus on daint... I am trying using only gfortran-7 and gcc-8, g++-8... if that still doesn't work, is it ok if I revert back to gnu-7 and leave updated to gnu 8 for the step that would replace our sarus tests with testing using your venv?

rheacangeo: Actually -- should I just turn off daint tests? since we've decided to not support containers there, there's no reason to keep running these there if we are going to very shortly start running tests 'on metal' using your venv

requirements.txt
rheacangeo: It works now yay!

rheacangeo:
launch jenkins

PR #172, opened 2020-12-21
speedup jenkins with python regression of FVDynamics

rheacangeo:
Running all the regression tests takes quite awhile, and can be quite the productivity drag to have to wait 30+ minutes for all the sequential and parallel serialized data regression tests to validate. 

Shifting to just running the main code of interest, that is FVDynamics for our automated testing could work, but its validation thresholds are so loose that making a code change based on it passing may obscure real code bugs on the smaller units. 

Now that we have validation, we can generate Python-based regressions, which we should be able to match exactly, per backend and platform. This PR introduces using a Python regression for just FVDynamics. We don't implement for every test because that would be a larger overhaul of the testing system and would not not provide immediate benefit.  The existing tests are still useful for digging into errors that crop up and potentially finding bugs that might still persist in our code, as well as potentially useful for performance metrics. 

ParallelTranslate classes now have the attribute 'python_regression,' which defaults to False, but for classes for which this is set to True, if '--python_regression' is included in the TEST_ARGS with a 'make tests_mpi' call, only classes with this attribute set to True will run, and instead of comparing to serialized Fortran data with a validation threshold, they'll be checked for an exact match of the SHA1 output created by the Python code. 
A new jenkins action run_python_regression.sh will run FVSubgridZ the normal way, then all the parallel tests with 'python_regressions=True,' which for the immediate future will just be FVDynamics. If this fails, then the sequential and parallel tests will run, and the output may help inform why you have a mismatch. Of course it is possible that all the sequential tests and parallel tests against the Fortran data pass, but the SHA1 changes with a code change. Those instances should be investigated, but the most likely outcome is that the regression data is updated. 

A new script tests/regenerate_python_regressions.sh can be used to remake the python regression data for all datasets. The line that overwrites what's in the cloud buckets is commented out.

This data regression relies on pytest-regressions, which actually hasn't been updated for Python 3.8, so there are some annoying warning messages. That piece of it is simple so we could implement our own code here if these warnings become fatals and this package isn't maintained. We seem to usually prefer to use functionality that's available over rewriting existing wheels. The functionality that it provides is checking answers against yaml regression files, and making the files if they don't exist yet.


eddie-c-davis (approved):
LGTM, very cool work!

twicki (commented):
lgtm

tests/test_translate.py
```diff
@@ -275,9 +293,12 @@ def test_parallel_savepoint(
     failure_stride,
     subtests,
     caplog,
+    python_regression,
     xy_indices=False,
 ):
     caplog.set_level(logging.DEBUG, logger="fv3core")
+    if python_regression and not testobj.python_regression:
+        pytest.xfail(f"python_regression not set for test {test_name}")
```
jdahm, line 301: I see this only in the parallel translate class, which mirrors the code here. Can regression tests be serial?

rheacangeo: For this PR they can only be parallel. But a future PR could extend this for serial tests if we decide to go that route. The motivation would be somewhat different from the purpose of this PR.

PR #173, opened 2021-01-04
Updating wind tendency names

oelbert:
Just revising the wind tendency names to be in line with the new names in fv3gfs-wrapper. This way they are more descriptive and specific.


rheacangeo:
Looks like the FVSubgridz test fails -- I think you just also need to update the name in the translate_fvsubgridz file (we do want to not have duplicate lists in the test code and stencil code at some point, but we have not reached that point yet)

rheacangeo (approved):
LGTM!

PR #174, opened 2021-01-04
Enable debug_mode on GT backends by default

jdahm:
Passes `debug_mode=True` to the stencil decorator calls when using GT backends unless `FV3_CPP_DEBUG_MODE=False` is set in the shell.

### DIFF
--- a/README.md
+++ b/README.md
@@ -130,7 +130,7 @@ common options for our tests, which you can add to `TEST_ARGS`:
 $ export FV3_STENCIL_REBUILD_FLAG=False
 ```
 
-FV3Core has certain backends that use a C++ compiler. GT4Py, by default, will instruct it to heavily optimize the generated code using and skip adding debug symbols. FV3Core on the other hand, by default will add debug symbols and will not use optimization. This means stencils using C++ backends will be ready to execute faster, but could be less efficient. If the goal is to test FV3Core performance, you will want to set:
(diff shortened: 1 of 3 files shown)


mcgibbon (approved):
LGTM. Pinging @rheacangeo since this may be relevant to the contributing documents or other documentation she has in a PR.

jdahm:
The red X is a Jenkins issue, not a test failure.
Screen Shot 2021-01-04 at 1 27 59 PM

eddie-c-davis (approved):
Approved.

README.md
```diff
@@ -122,12 +122,20 @@ common options for our tests, which you can add to `TEST_ARGS`:
...
 
+### Development and performance-related options
+
 **NOTE:** FV3 is current assumed to be by default in a "development mode", where stencils are checked each time they execute for code changes (which can trigger regeneration). This process is somewhat expensive, so there is an option to put FV3 in a performance mode by telling it that stencils should not automatically be rebuilt:
 
 ```shell
 $ export FV3_STENCIL_REBUILD_FLAG=False
 ```
 
+FV3Core has certain backends that use a C++ compiler. GT4Py, by default, will instruct it to heavily optimize the generated code using and skip adding debug symbols. FV3Core on the other hand, by default will add debug symbols and will not use optimization. This means stencils using C++ backends will be ready to execute faster, but could be less efficient. If the goal is to test FV3Core performance, you will want to set:
```
jdahm, line 133:
```suggestion
FV3Core has certain backends that use a C++ compiler. GT4Py, by default, will instruct it to heavily optimize the generated code, and will skip adding debug symbols. FV3Core on the other hand, defaults to a development mode, adding debug symbols disabling C++ optimizations. This means stencils using C++ backends will be ready to execute faster, but could be less efficient. If the goal is to test FV3Core performance, you will want to set:
```

jdahm:
Mikael reported that this does not speed up compilation after all. Let's skip doing this, at least for now.

PR #177, opened 2021-01-07
Add `memoryview` type check in gt4py_utils.asarray

eddie-c-davis:
This PR adds an additional to type check to the `gt4py_utils.asarray` method to convert `memoryview` objects to `cupy.ndarray`.


rheacangeo (approved):
Thank you for figuring this out!

rheacangeo:
launch jenkins

PR #180, opened 2021-01-12
Tests/gt4py in fv3core

rheacangeo:
When we update gt4py:develop, we first run the gt4py tests, and if those pass, run the fv3core tests. But, before the gt4py tests were run directly in a virtualenv on gcp and daint. and then the fv3core tests ran in the docker image. We wanted to do this anyways because we already had a difference in python being used (3.6 vs 3.8) in the environments and risk the gt4py not behaving correctly in the image even though all the tests pass. A cuda problem on GCP that doesn't appear to come up in the image motivated us to do this step now to unblock gt4py:develop progress. 
Here we add the option to run the gt4py tests (including the cuda ones) using the image and a change to the jenkins action to enable this. Some manual changes will need to happen after this is merged as well. This was tested (without the pushing of the new gt4py) here https://jenkins.ginko.ch/job/TEST-gt4py-in-fv3core-gpu/3/console

### DIFF
--- a/.jenkins/actions/update_gt4py.sh
+++ b/.jenkins/actions/update_gt4py.sh
@@ -8,5 +8,4 @@ make -C docker build_gt4py
 make gt4py_tests_gpu
 make tests
 make tests_mpi
-# TODO uncomment when verify the above works
 make -C docker push_gt4py
--- a/.jenkins/actions/update_gt4py.sh
+++ b/.jenkins/actions/update_gt4py.sh
@@ -5,7 +5,7 @@ set -x
 export CUDA=y
 make pull_environment
 make -C docker build_gt4py
-make container_gt4py_tests
+make gt4py_tests_gpu
 make tests
 make tests_mpi
 # TODO uncomment when verify the above works
--- a/.jenkins/actions/update_gt4py.sh
+++ b/.jenkins/actions/update_gt4py.sh
@@ -2,9 +2,11 @@
 set -e
 set -x
 # Run when we change the gt4py source code
-
+export CUDA=y
 make pull_environment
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -191,6 +191,9 @@ list_test_data_options:
 lint:
 	pre-commit run --all-files
 
+gt4py_tests_gpu:
+	CUDA=y make build && \
+        docker run --gpus all $(FV3_IMAGE) python3 -m pytest -x gt4py
```
jdahm, line 196: Could this be generalized to both GPUs and CPUs?

.jenkins/actions/update_gt4py.sh
```diff
@@ -2,9 +2,11 @@
...
 set -x
 # Run when we change the gt4py source code
-
+export CUDA=y
 make pull_environment
 make -C docker build_gt4py
+make gt4py_tests_gpu
 make tests
 make tests_mpi
+# TODO uncomment when verify the above works
```
jdahm, line 11: Reminder to us: remove before merging.

Makefile
rheacangeo: it could, but is there a need to run only the cpu ones when we can launch a gpu instance?

.jenkins/actions/update_gt4py.sh
rheacangeo: thanks! got it

Makefile
rheacangeo: meaning, before we use a new gt4py:develop, we probably want to require that both cpu and gpu tests work, and this runs both of them

jdahm (approved):
Looks good!

PR #181, opened 2021-01-13
Merge `Fillz` stencils and add 2D fields

eddie-c-davis:
This PR merges the `fix_top`, `fix_interior`, `fix_bottom`, and `final_check` stencil into a single `fix_tracer` stencil and adds 2D fields to remove the `numpy`/`cupy` code between the former stencils.


fv3core/stencils/fillz.py
```diff
@@ -61,101 +69,70 @@ def fix_interior(
...
-            dm,
-            zfix,
+            sum0,
+            sum1,
             fac,
-            origin=(i1, js, 1),
-            domain=(im, j_extent, km - 1),
+            origin=(i1, js, 0),
+            domain=(im, jext, km),
+            validate_args=False,
```
jdahm, line 136: Is there a reason `validate_args=False` here? We should make sure to remove it before merging.

eddie-c-davis: Because otherwise the domain computation fails.

eddie-c-davis:
launch jenkins

fv3core/stencils/fillz.py
eddie-c-davis: Removed!

rheacangeo:
launch jenkins

fv3core/utils/gt4py_utils.py
```diff
@@ -207,7 +207,7 @@ def make_storage_from_shape(
     *,
     dtype: DTypes = np.float64,
     init: bool = True,
-    mask: Tuple[bool, bool, bool] = (True, True, True),
+    mask: Optional[Tuple[bool, bool, bool]] = None,
```
rheacangeo, line 210: Is this because some are 2d fields now? Or rather, what is this change about?

tests/translate/translate_fillz.py
```diff
@@ -53,5 +53,5 @@ def compute(self, inputs):
         for varname, data in inputs["tracers"].items():
             index = utils.tracer_variables.index(varname)
             tracers[:, :, index] = np.squeeze(data[self.grid.slice_dict(ds)])
-        out = {"q2tracers": tracers}
-        return out
+        self.ignore_near_zero_errors = {"q2tracers": True}
```
rheacangeo, line 56: We usually have this in the init method by convention, but it does work here as well!

fv3core/stencils/fillz.py
```diff
@@ -4,39 +4,47 @@
...
+    q: FloatField,
+    dp: FloatField,
+    dm: FloatField,
+    dm_pos: FloatField,
+    zfix: IntFieldIJ,
+    upper_fix: FloatField,
+    lower_fix: FloatField,
+    sum0: FloatFieldIJ,
+    sum1: FloatFieldIJ,
+    fac: FloatFieldIJ,
```
rheacangeo, line 21: pretty exciting to see 2d fields!

rheacangeo (approved):
Impressive! That's a complicated function to refactor, and very cool it could be compressed that way with the help of 2d fields. I have a minor suggested change, but doesn't block approval.

rheacangeo (commented):
I have a slight concern that the ignore_near_zero_errors could be hiding something very subtle, as in the past this stencil and others like it sometimes did see actual bugs resulting in similar tiny errors.. e.g. where one pathway of a conditional made a variable 0 and the else was a very tiny number... but it doesn't seem to impact the validation of anything calling it (though did cause a python_regression shift), and it would be better to spend our time merging more stencils than chasing down errors in the noise.

fv3core/utils/gt4py_utils.py
eddie-c-davis: This is to force `mask` to be computed from the shape if it is not specifically supplied, with the goal of minimizing changes to existing code.

tests/translate/translate_fillz.py
eddie-c-davis: Removed, I will save this for the GPU tolerance PR so it can be set per backend.

eddie-c-davis:
launch jenkins

PR #182, opened 2021-01-13
remove the -e that was added at the last minute and untested

rheacangeo:


ofuhrer (approved):
LGTM, since this change was untested and this PR is a revert back to the original version since it was causing issues.

PR #183, opened 2021-01-13
Regions port: c2l_ord

oelbert:
Incorporating `region` syntax into c2l_ord and fusing the stencils into fewer calls.


jdahm:
There is only 1 call to `compute_cubed_to_latlon`, in `fv_dynamics.py`. Could you move the stencil and driver method `compute_cubed_to_latlon` there?

fv3core/stencils/c2l_ord.py
```diff
@@ -101,90 +85,15 @@ def compute_ord4(u, v, ua, va, comm, mode=1):
...
         comm.vector_halo_update(u, v, n_points=utils.halo)
 
-    utmp = utils.make_storage_from_shape(ua.shape, utils.origin)
-    vtmp = utils.make_storage_from_shape(va.shape, utils.origin)
-    j1 = grid.js + 1 if grid.south_edge else grid.js
-    j2 = grid.je - 1 if grid.north_edge else grid.je
-    i1 = grid.is_ + 1 if grid.west_edge else grid.is_
-    i2 = grid.ie - 1 if grid.east_edge else grid.ie
-    vector_tmp(
+    ord4_transform(
```
jdahm, line 88: This function isn't needed any more. Can you inline the stencil call into the driver `compute_cubed_to_latlon`?

jdahm (changes requested):
Can you add a docstring with in/out/inout specifications in the driver `compute_cubed_to_latlon`?

jdahm:
launch jenkins

jdahm:
@rheacangeo It seems like this passes, but it keeps getting blocked by other Jenkins tests.

fv3core/stencils/c2l_ord.py
```diff
@@ -28,172 +38,94 @@ def c2l_ord2(
...
+def compute_cubed_to_latlon(
+    u: Quantity,
+    v: Quantity,
+    ua: FloatField,
+    va: FloatField,
+    comm: CubedSphereCommunicator,
+    do_halo_update: bool,
+):
+    """
+    Interpolate D-grid to A-grid winds at latitude-longitude.
```
rheacangeo, line 84: 'at latitude-longitude coordinates' might make it more clear

fv3core/stencils/fv_dynamics.py
```diff
@@ -207,7 +208,7 @@ def post_remap(state, comm):
         )
 
 
-def wrapup(state, comm):
+def wrapup(state, cube: CubedSphereCommunicator):
```
rheacangeo, line 211: I would prefer the change from 'comm' to 'cube' be done to the whole stack in one separate PR, after discussing whether we want that change (sounded like you and Jeremy prefer it, I am less keen on it, but don't have a super strong objection).  But this is not a big deal, just introduces one more inconsistency in a sea of many. I think 'comm'  makes more sense because this model is not going to be anything but a cubed sphere, and cube would make more sense if we had multiple kinds of communicators present. You had suggested 'cube_comm', which is more clear it is a cubed sphere communicator, but that's a little long. I'm not going to block this PR on this point, just wanted to express my perspective. If 'cube' is popular I'm not going to stop it.

rheacangeo (commented):
Looks good, just a couple of documentation related comments.

rheacangeo (approved):
Looks great! approved conditional on the tests passing!

jdahm (approved):

PR #184, opened 2021-01-14
Rename parallel to horizontal and use max int offsets instead of None

jdahm:
Changes to be compatible with the regions in a new develop branch branch.


eddie-c-davis (approved):
LGTM!

PR #185, opened 2021-01-14
Add `format_source` option to disable `black` formatting

eddie-c-davis:
This PR adds a `format_source` option to the `global_config` and modifies the decorator to set it in `stencil_kwargs`. The default value is `False` to save time when building tests.

### DIFF
--- a/fv3core/decorators.py
+++ b/fv3core/decorators.py
@@ -217,6 +217,7 @@ class FV3StencilObject:
                     **axis_offsets,
                     **self._passed_externals,
                 },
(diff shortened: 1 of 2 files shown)


eddie-c-davis:
launch jenkins

jdahm (approved):
Other than the fact that the growth of these globals is worrisome, this is great! That isn't the fault of this PR, as that is the accepted approach at the moment.

eddie-c-davis:
Thank you for the review!

PR #187, opened 2021-01-19
adding option to override validation thresholds using yaml file

rheacangeo:


rheacangeo:
launch jenkins

tests/test_translate.py
```diff
@@ -112,6 +111,17 @@ def sample_wherefail(
     return "\n".join(return_strings)
 
 
+def process_override(threshold_overrides, testobj, test_name, backend):
```
eddie-c-davis, line 114: How would you feel about `platform` being optional? Something like,
```python
        plat = platform()
        if max_error is not None:
            if isinstance(max_error, dict) and plat in max_error:
                testobj.max_error = float(max_error[backend][plat])
            else:
                testobj.max_error = float(max_error[backend])
        if near_zero is not None:
            if isinstance(near_zero, dict) and plat in near_zero:
                testobj.near_zero = float(near_zero[backend][plat])
            else:
                testobj.near_zero = float(near_zero[backend])
```
Or moving it to the top of the hierarchy?
```python
    plat = platform()
    if plat in threshold_overrides:
        overrides = threshold_overrides[plat]
    else:
        overrides = threshold_overrides
    override = threshold_overrides.get(test_name, None)
    if override is not None:
        max_error = override.get("max_error", None)
        near_zero = override.get("near_zero", None)
        if max_error is not None:
            testobj.max_error = float(max_error[backend])
        if near_zero is not None:
            testobj.near_zero = float(near_zero[backend])
```
This way it will not need to be typed so often in the yaml file...

eddie-c-davis (changes requested):
Looks great! One small suggestion that might reduce the verbosity of the yaml files...

tests/test_translate.py
```diff
@@ -35,7 +35,6 @@ def success_array(computed_data, ref_data, eps, ignore_near_zero_errors):
     success = np.logical_or(
         np.logical_and(np.isnan(computed_data), np.isnan(ref_data)),
         compare_arr(computed_data, ref_data) < eps,
-        # np.isclose(computed_data, ref_data, rtol=eps * 1e-2, atol=eps * 1e-2),
```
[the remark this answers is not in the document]
rheacangeo: no worries! I just noticed it.

tests/test_translate.py
rheacangeo: I like that!

rheacangeo: alternatively we could use lists/dicts :
C_SW:
  max_error:
        - backend: numpy
          platform: docker
          value: 2e-10
      - backend: gtcuda
         value: 1e-10
XPPM:
      - platform: metal
        value: 1e-9 
   
The second array item would mean apply this to both platforms for gtcuda, the third would mean on the metal platform use this threshold with all backends for XPPM. 
What do you think of this?

rheacangeo: or actually, how about make the list at the test level:
C_SW:
 - max_error: 2e-10
   backend: numpy
  platform: docker 
- backend: gtcuda
  max_error: 1e-10
  near_zero: 1e-14
XPPM:
- platform: metal
   max_error: 1e-9

eddie-c-davis: Even better!

eddie-c-davis (approved):
Well done, works likes a charm!

.jenkins/jenkins.sh
```diff
@@ -97,7 +97,17 @@ if grep -q "parallel" <<< "${script}"; then
 	fi
     fi
 fi
-
+# set thresholds override file if it exists
+test_type=${optarg2##*_}
```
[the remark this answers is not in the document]
rheacangeo: for different actions they might be something else so generic helps with that, though we do have quite a bit of logic in this file expecting them to be what they are for the regression tests. The hope is when those aren't relevant they don't have a side effect... I agree it adds a layer of obscurity. I can check in with the plan Tobias is making for the standalone to see if he is using these, and whether they are the same thing.

PR #189, opened 2021-01-21
Add CUDA thresholds

eddie-c-davis:
This PR implements the following changes:

1. Converts the `near_zero` tolerance to a member of the translate classes, rather than a global variable in `test_translate`.
2. Adds the `near_zero` parameter to the `success` methods in `test_translate`.
3. Adds thresholds for the `gtcuda` backend to the `standard.yaml` overrides file.
4. Adds `near_zero` checking to the `Fillz` stencil.


eddie-c-davis:
launch jenkins

tests/test_translate.py
```diff
@@ -124,7 +127,13 @@ def process_override(threshold_overrides, testobj, test_name, backend):
...
             if spec["backend"] == backend and spec["platform"] == platform()
         ]
-        if len(matches) > 1:
+        if len(matches) == 1:
+            match = matches[0]
+            if "max_error" in match:
+                testobj.max_error = float(match["max_error"])
+            if "near_zero" in match:
+                testobj.near_zero = float(match["near_zero"])
+        elif len(matches) > 1:
```
rheacangeo, line 136: oops, my bad. thanks for catching this case, where there is an override specified, but 0 matches

tests/test_translate.py
```diff
@@ -337,11 +345,11 @@ def test_parallel_savepoint(
...
         pytest.xfail(f"no translate object available for savepoint {test_name}")
     # Reduce error threshold for GPU
-    if backend.endswith("cuda") and testobj.max_error < GPU_MAX_ERR:
-        testobj.max_error = GPU_MAX_ERR
+    if backend.endswith("cuda"):
+        testobj.max_error = max(testobj.max_error, GPU_MAX_ERR)
+        testobj.near_zero = max(testobj.near_zero, GPU_NEAR_ZERO)
     if threshold_overrides is not None:
         process_override(threshold_overrides, testobj, test_name, backend)
-    fv3core._config.set_grid(grid[0])
```
rheacangeo, line 344: ah I believe this is what is causing the tests to fail

eddie-c-davis: Thank you, I really must be more careful...

tests/test_translate.py
```diff
@@ -63,11 +63,14 @@ def sample_wherefail(
...
     test_name,
     ignore_near_zero_errors,
+    near_zero,
     xy_indices=False,
 ):
     found_indices = np.where(
         np.logical_not(
-            success_array(computed_data, ref_data, eps, ignore_near_zero_errors)
+            success_array(
+                computed_data, ref_data, eps, ignore_near_zero_errors, near_zero
```
rheacangeo, line 72: Looks like there is one other call to success_array in this method that needs the near_zero argument (it is an optional test argument that prints out the errors in a nice grid like layout), under xy_indices

tests/test_translate.py
```diff
@@ -165,9 +169,9 @@ def test_sequential_savepoint(
     if testobj is None:
         pytest.xfail(f"no translate object available for savepoint {test_name}")
     # Reduce error threshold for GPU
-    if backend.endswith("cuda") and testobj.max_error < GPU_MAX_ERR:
-        testobj.max_error = GPU_MAX_ERR
-        _near_zero = GPU_NEAR_ZERO
+    if backend.endswith("cuda"):
+        testobj.max_error = max(testobj.max_error, GPU_MAX_ERR)
+        testobj.near_zero = max(testobj.near_zero, GPU_NEAR_ZERO)
```
rheacangeo, line 174: nice

tests/test_translate.py
```diff
@@ -365,7 +374,11 @@ def test_parallel_savepoint(
         with subtests.test(varname=varname):
             failing_names.append(varname)
             assert success(
-                output[varname], ref_data[varname][0], testobj.max_error, near0
+                output[varname],
+                ref_data[varname][0],
+                testobj.max_error,
+                near0,
+                testobj.near_zero,
```
rheacangeo, line 381: it might make sense to change the near0 variable name (and above where it is assigned) to ignore_near0 so it is more clear that this is the boolean argument, while testobj.near_zero is the value. it is fine to leave as-is, however.

rheacangeo (approved):
Looks good to me! I made a couple of comments, one of which impacts an optional pytest feature.

tests/test_translate.py
eddie-c-davis: Nice catch!

PR #190, opened 2021-01-21
Infrastructure/daint virtualenv

rheacangeo:
This is part 1 of a 2 part rejiggering of the infrastructure to ensure we can run fv3core in a virtualenv on daint. Everytime a jenkins plan on daint runs, it will first install the virtualenv for that machine. We may not want to do this for every plan here so I am open to discussion. 

This is not sufficient to run a virtualenv on any machine. The next PR will update the Makefile and jenkins plans to continue to run through docker on gce, but use the virtualenv on daint for verification.


.jenkins/jenkins.sh
```diff
@@ -133,6 +133,11 @@ if [ ${container_engine} == "sarus" ]; then
     fi
 fi
 
+if [ ${host} == "daint" ]; then
+    daintenv=${SCRATCH}/vcm_env_${BUILD_TAG}
+    ${root}/../install_virtualenv.sh ${daintenv}
```
twicki, line 138: why is this parent of root?

install_virtualenv.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+virtualenv_path=$1
```
twicki, line 2: should we add a check that $1 is not empty?

install_virtualenv.sh
```diff
@@ -0,0 +1,10 @@
...
+
+root=`dirname $0`
+(cd ${root}/external/daint_venv && ./install.sh ${virtualenv_path})
```
twicki, line 5: is this working with a commit to daint_venv? currently `install.sh` always goes into `/project[...]`, see https://github.com/VulcanClimateModeling/daint_venv/blob/c15c1d93700a644aac7f5d7f32889a02b9fb8d82/install.sh#L5

twicki (commented):
I still have some understanding questions

.jenkins/jenkins.sh
rheacangeo: I was thinking it would be useful to have a script to install a fv3core virtualenv at the top level. It doesn't quite work currently because you still need to have a few other things installed, like gt4py, and other things that happen in the daint specific script. I am ok with putting it in the jenkins directory for now since it won't be used outside of daint yet.

install_virtualenv.sh
rheacangeo: sure that's a good idea

install_virtualenv.sh
rheacangeo: yeah right now the daint_env submodule is pointing at a branch. I will open a parallel PR there so it can point to master

.jenkins/jenkins.sh
twicki: either way is ok with me

install_virtualenv.sh
twicki: ah, I did not see that. Makes sense

install_virtualenv.sh
twicki: sweet

twicki (approved):
lgtm. merge-able once it passes

rheacangeo:
launch jenkins

rheacangeo:
launch jenkins

PR #192, opened 2021-01-22
Add 'ignore_near_zero_errors' fields to overrides/standard.yaml

eddie-c-davis:
This PR adds the option to set `ignore_near_zero_errors` field names in the `overrides/standard.yaml` file by updating `test_translate.process_overrides`, and adding the appropriate field names for the `gtcuda` backend.


rheacangeo (approved):
very cool!

PR #195, opened 2021-01-25
c_sw as a single stencil

jdahm:
Builds on #153.


jdahm:
This is moving to a follow-up PR.

PR #196, opened 2021-01-25
Add local axis offsets

jdahm:
These are located at `grid.is_, grid.ie, grid.js, grid.je` and are prefixed by `local_`.


eddie-c-davis (approved):
Looks good!

PR #197, opened 2021-01-26
Rename main to main_loop in timers

ofuhrer:
This renames the main timer (which alludes to the main() function, which it is not) to main_loop as specified in the timer specification document (https://paper.dropbox.com/doc/Performance-timers-Fortran-Python--BD4~XI4SNtfc~gBml74rb_XUAg-lSnQfJGiZ6LVfSLQpcnwT).


twicki (approved):
lgtm. What do we do about the old files? Rename the main key or delete them?

PR #199, opened 2021-01-26
change from median to mean

twicki:
## Major Changes
- Changing from plotting the median to plotting the mean as this is the only metric the Fortran code generates.

## Minor Changes
- Changed the labels in the project repo to reflect this change.


ofuhrer (approved):
Sure thing!

PR #200, opened 2021-01-26
adding a mask option to the translate class metadata

rheacangeo:


rheacangeo:
Actually, we need more than this to enable gt4py storages created from serialized inputs to be properly 2d/1d, so closing

PR #202, opened 2021-01-27
Fix 2D writes in fillz.fix_stencil

eddie-c-davis:
Coauthor: @twicki

This PR resolves the writes to 2D fields in parallel regions in `fillz.fix_stencil` that are errors in the latest `gt4py`.


rheacangeo (approved):
LGTM!

PR #204, opened 2021-01-28
Adding Timers to fv_dynamics and standardizing timings 

oelbert:
This PR adds further timing utility to fv3core:
- Added a timer argument to fv_dynamics, compute and do_dyn to enable timing of dynamics, remapping, and tracer advection.
- Added `print_and_write_global_timings` to dynamics.py that incorporates Tobias' get totals, gathers and `write_to_json` into the fv3utils.timer framework
- Switched timers inside dynamics.py to fv3utils.timer form


examples/standalone/benchmarks/run_on_daint.sh
```diff
@@ -51,10 +51,13 @@ fi
 cd $ROOT_DIR
 rm -rf vcm_1.0
 
+echo "copying in the venv"
 cp -r /project/s1053/install/venv/vcm_1.0/ .
 git submodule update --init --recursive
+echo "install requirements"
 vcm_1.0/bin/python -m pip install external/fv3gfs-util/
 vcm_1.0/bin/python -m pip install .
+vcm_1.0/bin/python -m pip install gitpython
```
twicki, line 60: should no longer be needed, see
https://github.com/VulcanClimateModeling/daint_venv/blob/5dcd1355a7b603b10b07278eb50e2642129668ed/install.sh#L35

examples/standalone/runfile/dynamics.py
```diff
@@ -1,120 +1,152 @@
...
+    filename = now.strftime("%Y-%m-%d-%H-%M-%S")
+    experiment["setup"] = {}
+    experiment["setup"]["experiment time"] = dt_string
+    experiment["setup"]["data set"] = experiment_name
+    experiment["setup"]["timesteps"] = time_step
+    experiment["setup"]["hash"] = sha
+    experiment["setup"]["version"] = "python/" + backend
+    experiment["times"] = {}
+    for name, value in timer.times.items():
+        if is_root:
```
twicki, line 40: can we move this down to the second `is_root` block (Line 49) to only have one?

tests/translate/__init__.py
```diff
@@ -76,3 +76,6 @@
 from .translate_xtp_u import TranslateXTP_U
 from .translate_yppm import TranslateYPPM, TranslateYPPM_2
 from .translate_ytp_v import TranslateYTP_V
+
+
+# from fv3core.testing import  TranslateGrid, TranslateFortranData2Py
```
twicki, line 81: we should be able to remove this

twicki (changes requested):

examples/standalone/runfile/dynamics.py
oelbert: I don't know if we can do that, since the other `if is_root` block is inside a loop over label/operation. Is it better to take it out of the if-statement and just have every rank set `experiment["times"][name] = {}` or to have the if-statement there?

twicki: you're absolutely right

examples/standalone/benchmarks/run_on_daint.sh
```diff
@@ -51,8 +51,10 @@ fi
 cd $ROOT_DIR
 rm -rf vcm_1.0
 
+echo "copying in the venv"
 cp -r /project/s1053/install/venv/vcm_1.0/ .
 git submodule update --init --recursive
+echo "install requirements"
```
mcgibbon, line 57:
```suggestion
echo "installing requirements..."
```

examples/standalone/benchmarks/run_on_daint.sh
```diff
@@ -51,8 +51,10 @@ fi
 cd $ROOT_DIR
 rm -rf vcm_1.0
 
+echo "copying in the venv"
```
mcgibbon, line 54:
```suggestion
echo "copying in the venv..."
```

examples/standalone/runfile/dynamics.py
```diff
@@ -1,120 +1,152 @@
...
 import fv3core
 import fv3core._config as spec
 import fv3core.stencils.fv_dynamics as fv_dynamics
 import fv3core.testing
 import fv3gfs.util as util
-from fv3core.testing import GlobalTimer, write_to_json
+from fv3gfs.util import Timer
+
+
+def print_and_write_global_timings(
```
mcgibbon, line 20: It would be worthwhile to split this function into more routines so that it contains no low-level code and only high-level function calls. It has several responsibilities right now:
- Generate a timestamp in two different formats
- generate an experiment timing report dictionary
- save the report dictionary to a json file
- print the report (in a format slightly different from what's saved to disk)
- gather statistics across ranks

The feedback below is just for feedback and not to ask you to change anything here.

This function contains IO nested into a function. In terms of rules of thumb, this ends up breaking the single responsibility principle - you end up with many parts of your code with the "how do I write data" responsibility. A better pattern would be to have a function return a timing report, and print and write it outside of that function. One example where this is necessary is if someone wants to use this routine to write a timing report to Google Cloud. We've run into this kind of thing as an issue a lot on the ML team. I think it's not much of an issue here, and you don't need to change it, mostly good to keep in mind when writing code in the fv3core package itself.

examples/standalone/runfile/dynamics.py
```diff
@@ -1,120 +1,152 @@
...
+    timer, experiment_name, time_step, backend, comm, root=0
+):
+    is_root = comm.Get_rank() == root
+    recvbuf = np.array(0.0)
+    experiment = {}
+    now = datetime.now()
+    sha = git.Repo(
+        pathlib.Path(__file__).parent.absolute(), search_parent_directories=True
+    ).head.object.hexsha
```
mcgibbon, line 29: Also OK with this as-is, but a more extensible/re-usable pattern would be to pass in a "label" for the report from the command-line. Having this low-level code here adds an extra responsibility to the function.

examples/standalone/runfile/dynamics.py
```diff
@@ -1,120 +1,152 @@
...
+    dt_string = now.strftime("%d/%m/%Y %H:%M:%S")
+    filename = now.strftime("%Y-%m-%d-%H-%M-%S")
+    experiment["setup"] = {}
+    experiment["setup"]["experiment time"] = dt_string
+    experiment["setup"]["data set"] = experiment_name
```
mcgibbon, line 34: We normally write "dataset" as one word in Python packages (e.g. netcdf4.Dataset). All the data is for the "experiment", so "experiment time" doesn't really clear things up - it could mean time the experiment took to run. I'd suggest "timestamp" which I think says what you want.

If using spaces I'd suggest underscores instead.

```suggestion
    experiment["setup"]["timestamp"] = dt_string
    experiment["setup"]["dataset"] = experiment_name
```

examples/standalone/runfile/dynamics.py
```diff
@@ -1,120 +1,152 @@
...
-    # namelist setup
-    spec.set_namelist(data_dir + "/input.nml")
-
-    nml2 = yaml.safe_load(
-        open(
-            data_dir + "/input.yml",
-            "r",
+    timer = Timer()
+    timer.start("total")
+    with timer.clock("init"):
```
mcgibbon, line 64: Please expand "init".

examples/standalone/runfile/dynamics.py
```diff
@@ -1,120 +1,152 @@
...
-    # Run the dynamics
-    timing.time("main_loop")
-    for i in range(time_step - 1):
+        parser.add_argument(
+            "backend",
+            type=str,
+            action="store",
+            help="path to the namelist",
+        )
+        args = parser.parse_args()
```
mcgibbon, line 87: Please refactor this and the above code into a routine so you can call `args = parse_args()`. Inside parse_args, you would define the ArgumentParser and `return parser.parse_args()`.

examples/standalone/runfile/dynamics.py
```diff
@@ -1,120 +1,152 @@
...
+        backend = args.backend
+        data_dir = args.data_dir
+        time_step = args.time_step
```
mcgibbon, line 90: Is it necessary to unpack these? If you use e.g. `args.backend` below, then it's clearer when reading e.g. line 97 that "backend" comes from the command-line arguments.

examples/standalone/runfile/dynamics.py
```diff
@@ -1,120 +1,152 @@
...
 import yaml
+from mpi4py import MPI
 
 import fv3core
 import fv3core._config as spec
 import fv3core.stencils.fv_dynamics as fv_dynamics
 import fv3core.testing
 import fv3gfs.util as util
-from fv3core.testing import GlobalTimer, write_to_json
+from fv3gfs.util import Timer
```
mcgibbon, line 17: Please use `util.Timer` instead, following the pattern for other objects from fv3gfs.util. I would also generally suggest `import fv3gfs.util as fv3util` instead of as `util`, but this is optional.

examples/standalone/runfile/dynamics.py
```diff
@@ -1,120 +1,152 @@
...
+        rank = comm.Get_rank()
+
+        # fv3core specific setup
+        fv3core.set_backend(backend)
+        fv3core.set_rebuild(False)
+
+        # namelist setup
+        spec.set_namelist(data_dir + "/input.nml")
+
+        nml2 = yaml.safe_load(
```
mcgibbon, line 103: Why is this "2"? I don't see a `nml0` or `nml1` or even `nml`.

fv3core/stencils/fv_dynamics.py
```diff
@@ -160,7 +160,7 @@ def compute_preamble(state, comm):
         )
 
 
-def do_dyn(state, comm):
+def do_dyn(state, comm, timer=None):
```
mcgibbon, line 163: if timer is None, please set it to a default value of `fv3gfs.util.NullTimer` (which we should make public), so you don't need repeated "if timer is not None" logic below. Then please use the `with` syntax for timings below.

fv3core/stencils/fv_dynamics.py
```diff
@@ -169,10 +169,19 @@ def do_dyn(state, comm):
         domain=grid.domain_shape_standard(),
     )
     print("DynCore", grid.rank)
+    if timer is not None:
+        # start dyn timer
```
mcgibbon, line 173: Please avoid comments saying what the code is doing, in favor of making the code clear enough that you can tell what it is doing. In this case, your code is already clear, and I don't think these comments are needed.

fv3core/stencils/fv_dynamics.py
```diff
@@ -301,10 +315,10 @@ def fv_dynamics(state, comm, consv_te, do_adiabatic_init, timestep, ptop, n_spli
             "ks": ks,
         }
     )
-    compute(state, comm)
+    compute(state, comm, timer)
 
 
-def compute(state, comm):
+def compute(state, comm, timer=None):
```
mcgibbon, line 321: Should also include a "default to NullTimer" block in this function.

fv3core/stencils/fv_dynamics.py
```diff
@@ -169,10 +169,19 @@ def do_dyn(state, comm):
...
     dyn_core.compute(state, comm)
+    if timer is not None:
+        # pause dyn timer
+        timer.stop("Dynamics")
     if not spec.namelist.inline_q and state.nq != 0:
         if spec.namelist.z_tracer:
             print("Tracer2D1L", grid.rank)
+            if timer is not None:
+                # start tracer timer
+                timer.start("Tracer advection")
```
mcgibbon, line 184: Can all these labels be changed to lowercase/pothole case, as we would if they were variable names?

fv3core/stencils/fv_dynamics.py
```diff
@@ -169,10 +169,19 @@ def do_dyn(state, comm):
         domain=grid.domain_shape_standard(),
     )
     print("DynCore", grid.rank)
+    if timer is not None:
+        # start dyn timer
+        timer.start("Dynamics")
```
mcgibbon, line 174: This label could probably be `dyn_core`? "dynamics" also includes tracer advection and remapping.

mcgibbon (changes requested):
I didn't realize NullTimer wasn't public in fv3gfs-util, sorry. We should fix that and use NullTimer for defaults. A couple other renames/refactors requested.

examples/standalone/runfile/dynamics.py
oelbert: I think I set this to distinguish it from the namelist coming from input.nml back in August or something

rheacangeo:
launch jenkins

examples/standalone/benchmarks/plotting.py
```diff
@@ -27,12 +27,12 @@
...
-            k["setup"]["experiment time"], "%d/%m/%Y %H:%M:%S"
-        )
+        key=lambda k: datetime.strptime(k["setup"]["timestamp"], "%d/%m/%Y %H:%M:%S")
     )
-    for plottype in ["mainLoop", "initTotal"]:
-        keyval = ["main_loop"] if plottype == "mainLoop" else ["init", "total"]
+    for plottype in ["mainLoop", "initializationTotal"]:
+        keyval = (
+            ["main_loop"] if plottype == "mainLoop" else ["initialization", "total"]
+        )
```
mcgibbon, line 35: Confused about what this does, can you add an intention-describing comment or use a variable name for `keyval` that reflects it?

e.g. something to do with subplots or line_names or whatever it actually is?

examples/standalone/runfile/dynamics.py
mcgibbon: Can you add a try: except block around this sha line which puts a default value if this is run not in a git repo?

examples/standalone/runfile/dynamics.py
```diff
@@ -37,84 +38,112 @@
...
+        ]:
+            comm.Reduce(np.array(value), recvbuf, op=op)
+            if is_root:
+                if label == "mean":
+                    recvbuf /= comm.Get_size()
+                print(f"    {label}: {recvbuf}")
+                experiment["times"][name][label] = float(recvbuf)
+
+
+def write_global_timings(experiment, comm, root=0):
```
mcgibbon, line 82: I would suggest refactoring this so one of its arguments is either a filename or (better) an open file object. Might mean writing another `get_filename` helper function.

fv3core/stencils/fv_dynamics.py
```diff
@@ -169,10 +169,13 @@ def do_dyn(state, comm):
         domain=grid.domain_shape_standard(),
     )
     print("DynCore", grid.rank)
+    timer.start("DynCore")
     dyn_core.compute(state, comm)
+    timer.stop("DynCore")
```
mcgibbon, line 174: Please use the context manager here and in the below blocks.
```suggestion
    with timer.clock("DynCore"):
        dyn_core.compute(state, comm)
```

mcgibbon (commented):
Good improvement! Only required change is to use the `clock` context manager instead of start and stop. The other suggestions are optional.

fv3core/stencils/fv_dynamics.py
mcgibbon: I'd also ask we use pothole case for these timing labels.

examples/standalone/benchmarks/plotting.py
oelbert: Gonna deputize @twicki on this one

mcgibbon (approved):
LGTM.

rheacangeo:
launch jenkins

twicki (approved):
lgtm.
there is an empty file `.jenkins/actions/nightly_bench.sh ` is this an artifact?
And the change to the utils subdirectory is intentional?

PR #205, opened 2021-01-29
Refactor VorticityVolumeMean

mcgibbon:
This PR refactors the VorticityVolumeMean stencil in the following ways:

- three sub-stencils are merged into one stencil, converting two storages into temporaries
- wk is renamed to "vorticity" within the stencil, though the function name elaborates that this is horizontal relative vorticity and the docstring elaborates that it's an area mean. This could be propagated into d_sw in a later PR (or renamed along with wk to another new name in a later PR).
- the stencil is moved to d_sw, the only place where it is used
- added docstring and type hinting
- updated grid.domain_shape_standard and grid.domain_shape_compute to accept "add" keyword argument
- stencil tests for VorticityVolumeMean no longer use or check ut and vt, but it is clear from the regression tests and reading the Fortran code that these are temporaries - their inputs are never read, and their outputs are never re-used down the line.


tests/translate/translate_vorticityvolumemean.py
```diff
@@ -1,18 +1,27 @@
...
     def __init__(self, grid):
         super().__init__(grid)
-        self.in_vars["data_vars"] = {"u": {}, "v": {}, "ut": {}, "vt": {}, "wk": {}}
+        self.in_vars["data_vars"] = {"u": {}, "v": {}, "wk": {}}
         self.out_vars = {
             "wk": {},
-            "ut": grid.x3d_domain_dict(),
-            "vt": grid.y3d_domain_dict(),
+            # "ut": grid.x3d_domain_dict(),
+            # "vt": grid.y3d_domain_dict(),
```
jdahm, line 12: Was there a reason these are commented out here, but removed above?

mcgibbon: Nope, can remove.

mcgibbon, line 12:
```suggestion
```

fv3core/stencils/d_sw.py
```diff
@@ -18,14 +18,15 @@
...
 import fv3core.stencils.heatdiss as heatdiss
-import fv3core.stencils.vorticity_volumemean as vort_mean
 import fv3core.stencils.xtp_u as xtp_u
 import fv3core.stencils.ytp_v as ytp_v
 import fv3core.utils.corners as corners
 import fv3core.utils.global_constants as constants
 import fv3core.utils.gt4py_utils as utils
 from fv3core.decorators import gtstencil
 
+from ..utils.typing import FloatField
```
jdahm, line 28: I'd rather see all these or none follow relative imports.

mcgibbon: I'm fine with switching it.

fv3core/utils/grid.py
```diff
@@ -305,14 +305,14 @@ def y3d_compute_domain_x_dict(self):
...
 
     def domain_shape_buffer_k(self):
         return (self.nid, self.njd, self.npz + 1)
 
-    def domain_shape_compute(self):
-        return (self.nic, self.njc, self.npz)
+    def domain_shape_compute(self, add: Tuple[int, int, int] = (0, 0, 0)):
+        return (self.nic + add[0], self.njc + add[1], self.npz + add[2])
 
     def domain_shape_compute_buffer_2d(self, add: Tuple[int, int, int] = (1, 1, 0)):
```
rheacangeo, line 317: These two functions are now identical except for the default add value. We should replace all instances of domain_shape_buffer_2d with domain_shape_compute with the correct add if you are going to make this change.

tests/translate/translate_vorticityvolumemean.py
```diff
@@ -1,18 +1,25 @@
...
         self.make_storage_data_input_vars(inputs)
-        vm.compute(**inputs)
+        vorticity_from_winds(
+            inputs["u"],
+            inputs["v"],
+            self.grid.dx,
+            self.grid.dy,
+            self.grid.rarea,
+            inputs["wk"],
+            origin=(0, 0, 0),
```
rheacangeo, line 22: self.grid.default_origin() is preferred for consistency

rheacangeo (commented):
I like the idea of removing ut and vt from validation, but they are actually used by later stencils in d_sw u_from_ke and v_from_ke within d_sw. So there is still a benefit to checking they are correct on the output of of this test. Also, d_sw will likely fail as ut and vt are modified and then used.

tests/translate/translate_vorticityvolumemean.py
```diff
@@ -1,18 +1,25 @@
-import fv3core.stencils.vorticity_volumemean as vm
+from fv3core.stencils.d_sw import vorticity_from_winds
```
rheacangeo, line 1: I think this is not what you named it

mcgibbon:
launch jenkins

mcgibbon:
> Also, d_sw will likely fail as ut and vt are modified and then used.
I thought I had run it and it hadn't failed, but it turns out it does! Good catch!

tests/translate/translate_vorticityvolumemean.py
mcgibbon: Good catch!

tests/translate/translate_vorticityvolumemean.py
mcgibbon: Updated.

We should maybe consider renaming this later, it was unclear to me from the name whether this meant start of halo data or start of compute domain. And looking at the code didn't really help, because I wasn't sure what all the single-letter-append after is/ie meant.

fv3core/utils/grid.py
mcgibbon: Done.

We should consider doing the same for the various other grid functions in a later PR, many of them can be written as a `domain_shape_compute` call (e.g. `domain_shape_compute_x` is `domain_shape_compute(add=(1, 0, 0))`.

mcgibbon:
launch jenkins

tests/translate/translate_vorticityvolumemean.py
rheacangeo: 'default' or 'standard' refer to the domain isd:ied, jsd:jed,0:npz-1 inclusive, while names with 'compute' in them refer to the is/ie, js/je compute region. By single letter append, do you mean the d in isd? That is just to follow the conventions of the fortran code naming of domain indices.  Renaming all of it is of course open to discussion to do at some point. As we merge stencils they will become more compute region centric, in which case using 'default' or 'standard' to name the full domain including halos will make less sense.

fv3core/stencils/d_sw.py
```diff
@@ -63,6 +63,36 @@ def flux_adjust(w: sd, delp: sd, gx: sd, gy: sd, rarea: sd):
...
+    vorticity: FloatField,
+):
+    """
+    Compute the area mean relative vorticity in the z-direction from the D-grid winds.
+
+    Args:
+        u (in): x-direction wind on D grid
+        v (in): y-direction wind on D grid
+        ut (out): u * dx
+        vt (out): v * dy
```
rheacangeo, line 84: we can see immediately below that ut and vt are vector windspeed * del (cell). it doesn't hurt, but without words to describe these, do we need this?

rheacangeo (approved):
Great! This indeed is a case where gt4py can do the work for us rather than specifying different bounds for each variable computation.

fv3core/stencils/d_sw.py
mcgibbon: This is where I'd look to figure out what things are, and writing this docstring I did have to take time to read every line of code to see the input data is entirely overwritten and only overwritten with that value. I think even with code below saying what these are, it's better having this than having nothing.

In the longer term when more stencils are joined, these are intermediate values which should probably exist within a single gtscript function, instead of being passed between two functions/stencils.

jdahm (approved):

PR #206, opened 2021-01-29
Refactor heat_damping in D_SW

mcgibbon:
This PR refactors the heat damping code in D_SW in the following ways:

- two stencils over 3 calls are combined into one stencil. While it would be nice to split that stencil into routines, I cannot determine a meaningful name to split them over, since it's not very clear what the code is doing
- Added a docstring with intents, and at least one variable definition
- Converted four storages into temporaries: fx, fy, gx, and gy. gx and gy are still defined as storages since they are used in another routine (probably also can be converted to temporaries later)
- Converted ub and vb from inout to in within stencil. Since tests pass, these are not actually used after heat is calculated

### DIFF
--- a/fv3core/stencils/d_sw.py
+++ b/fv3core/stencils/d_sw.py
@@ -146,7 +146,7 @@ def heat_damping_term(ub, vb, gx, gy, rsin2, cosa_s, u2, v2, du2, dv2):
 
 
 @gtstencil()
-def heat_damping(
+def heat_source_from_vorticity_damping(
     ub: FloatField,
     vb: FloatField,
     ut: FloatField,
@@ -228,7 +228,7 @@ def heat_from_damping(
     diss_est,
     kinetic_energy_fraction_to_damp,
 ):
-    heat_damping(
+    heat_source_from_vorticity_damping(
         ub,
         vb,
         ut,
--- a/fv3core/stencils/d_sw.py
+++ b/fv3core/stencils/d_sw.py
@@ -135,14 +135,6 @@ def adjust_w_and_qcon(w: sd, delp: sd, dw: sd, q_con: sd, damp_w: float):
         q_con = q_con / delp
 
 
-@gtstencil()
-def heatdamping_setup(ub: sd, vt: sd, fy: sd, u: sd, gy: sd, rdx: sd, sign: float):
-    with computation(PARALLEL), interval(...):
-        ub[0, 0, 0] = (ub + sign * vt) * rdx
-        fy[0, 0, 0] = u * rdx
-        gy[0, 0, 0] = fy * ub
-
-
 @gtscript.function
 def heat_damping_term(ub, vb, gx, gy, rsin2, cosa_s, u2, v2, du2, dv2):
     return rsin2 * (
@@ -153,9 +145,13 @@ def heat_damping_term(ub, vb, gx, gy, rsin2, cosa_s, u2, v2, du2, dv2):
 
 
 @gtstencil()
-def heatdamping(
+def heat_damping(
     ub: sd,
     vb: sd,
+    ut: sd,
+    vt: sd,
+    u: sd,
+    v: sd,
     delp: sd,
     fx: sd,
     fy: sd,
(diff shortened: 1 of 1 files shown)


mcgibbon:
launch jenkins

fv3core/stencils/d_sw.py
```diff
@@ -153,29 +146,67 @@ def heat_damping_term(ub, vb, gx, gy, rsin2, cosa_s, u2, v2, du2, dv2):
...
+        kinetic_energy_fraction_to_damp (in): according to its comment in fv_arrays,
+            the fraction of kinetic energy to explicitly damp and convert into heat.
+            TODO: confirm this description is accurate, why is it multiplied
+            by 0.25 below?
+        calculate_dissipation_estimate (in): If 1, calculate dissipation estimate.
+            Equivalent in Fortran model is do_skeb
+    """
     with computation(PARALLEL), interval(...):
+        ubt = (ub + vt) * rdx
+        fy = u * rdx
```
mcgibbon, line 194: @jdahm from an optimization perspective, is there any penalty to removing fx/fy entirely and replacing it with e.g. u * rdx? the latter is much more clear, I'm sure these temporaries were only made to manually optimize the code.

jdahm: If it’s not already read in the computation elsewhere, it will generate an extra read, but that is really minor compared to the other issues we are facing. Go for clarity!

mcgibbon:
launch jenkins

jdahm (approved):
Looks good!

fv3core/stencils/d_sw.py
```diff
@@ -153,29 +146,67 @@ def heat_damping_term(ub, vb, gx, gy, rsin2, cosa_s, u2, v2, du2, dv2):
...
-        du2 = ub + ub[0, 1, 0]
+        du2 = ubt + ubt[0, 1, 0]
         v2 = fx + fx[1, 0, 0]
-        dv2 = vb + vb[1, 0, 0]
-        dampterm = heat_damping_term(ub, vb, gx, gy, rsin2, cosa_s, u2, v2, du2, dv2)
-        heat_source[0, 0, 0] = delp * (heat_source - damp * dampterm)
-        diss_est[0, 0, 0] = diss_est - dampterm if do_skeb == 1 else diss_est
+        dv2 = vbt + vbt[1, 0, 0]
+        dampterm = heat_damping_term(ubt, vbt, gx, gy, rsin2, cosa_s, u2, v2, du2, dv2)
+        heat_source[0, 0, 0] = delp * (
```
oelbert, line 204: Not a need, but at some point we should try to standardize whether to use the [0,0,0] indices on assignments or not

mcgibbon: Yeah I was leaning towards liking [0, 0, 0] but the code just looks so nice when there are few/no square brackets on either side. And for temporaries like dv2, you cannot put square brackets on the left.

oelbert (approved):
Good work!

fv3core/stencils/d_sw.py
rheacangeo: Also the square brackets bring an implication that you could do something other than [0, 0, 0], which you can't, but we've already discussed that. It does help remind you if you are looking at an input field or a temporary. Always tradeoffs.

rheacangeo (approved):
Awesome. This was definitely low hanging fruit that fit the pattern of what gt4py can do for you, good find. Adding the docstrings helps it as well.

fv3core/stencils/d_sw.py
```diff
@@ -744,10 +768,20 @@ def d_sw(
...
+            ut,
+            vt,
+            u,
+            v,
+            delp,
+            fx,
+            fy,
+            heat_s,
+            diss_e,
+            kinetic_energy_fraction_to_damp,
```
rheacangeo, line 784: This name is a bit long and turns one line into 12, but it is indeed descriptive

mcgibbon:
> Awesome. This was definitely low hanging fruit that fit the pattern of what gt4py can do for you, good find. Adding the docstrings helps it as well.

Thanks!

mcgibbon:
launch jenkins

PR #210, opened 2021-01-30
Region refactor: xppm/yppm/xtp_u/ytp_v

jdahm:


fv3core/stencils/ytp_v.py
```diff
@@ -1,230 +1,165 @@
...
 ):
-    with computation(PARALLEL), interval(...):
-        b0 = get_b0(bl=bl, br=br)
-        smt5 = is_smt5_mord5(bl, br) if mord == 5 else is_smt5_most_mords(bl, br, b0)
-        tmp = smt5[0, -1, 0] + smt5 * (smt5[0, -1, 0] == 0)
-        cfl = c * rdy[0, -1, 0] if c > 0 else c * rdy
-        fx0 = fx1_fn(cfl, br, b0, bl)
-        # TODO: add [0, 0, 0] when gt4py bug is fixed
-        flux = final_flux(c, q, fx0, tmp)  # noqa
+    from __externals__ import mord
```
mcgibbon, line 30: For readability and documentation, I would suggest doing this and then using the longer name below. We can consider refactoring the namelist name itself much later, for now it's worth staying consistent with Fortran.
```suggestion
    from __externals__ import mord
    momentum_advection_scheme = mord
```

fv3core/stencils/ytp_v.py
```diff
@@ -1,230 +1,165 @@
...
+    b0 = bl + br
+    cfl = c * rdy[0, -1, 0] if c > 0 else c * rdy
+    fx1 = yppm.fx1_fn(cfl, br, b0, bl)
+    return q[0, -1, 0] + fx1 if c > 0.0 else q + fx1
 
 
-@gtstencil()
-def get_flux_v_stencil(
-    q: sd, c: sd, al: sd, rdy: sd, bl: sd, br: sd, flux: sd, mord: int
+def get_flux_v(
```
mcgibbon, line 27: Please add docstrings to gtscript functions, too!

fv3core/stencils/ytp_v.py
```diff
@@ -1,230 +1,165 @@
...
+
+def compute(c: FloatField, u: FloatField, v: FloatField, flux: FloatField):
+    """
+    Compute flux of kinetic energy in x-dir using the PPM method.
+
+    Notes:
+        u is passed in here, but is unused in the stencil.
+
+    Args:
+        c (in): Courant number
```
mcgibbon, line 136: Careful, I think this is the courant number "in flux form". Double-check units in how it's used below to confirm this. Also, I suggest renaming this everywhere (in this file) to "courant". No need to rename it where this function is called, we can work our way up over time (and besides, we might need to rename it further).

fv3core/stencils/ytp_v.py
```diff
@@ -1,230 +1,165 @@
...
+                bl = 0.0
+                br = 0.0
+            # }
+
+            flux = get_flux_v_ord8plus(v, c, rdy, bl, br)
+
+
+def compute(c: FloatField, u: FloatField, v: FloatField, flux: FloatField):
+    """
+    Compute flux of kinetic energy in x-dir using the PPM method.
```
mcgibbon, line 130: Are all of the mord options the PPM method? Some of them seemed to be other methods, according to the documentation.

fv3core/stencils/ytp_v.py
```diff
@@ -1,230 +1,165 @@
...
+    if jord not in (5, 6, 7, 8):
+        raise NotImplementedError(
+            "Currently ytp_v is only supported for hord_mt == 5,6,7,8"
         )
+
+    stencil = gtstencil(
+        definition=_compute_stencil,
+        externals={
+            "jord": jord,
+            "mord": jord,  # Note: no abs here
```
mcgibbon, line 152: What does this note mean? Please expand/clarify it or remove it.

fv3core/stencils/ytp_v.py
```diff
@@ -1,230 +1,165 @@
...
+    fx0 = yppm.fx1_fn(cfl, br, b0, bl)
+    return yppm.final_flux(c, q, fx0, tmp)
 
 
-@gtstencil()
-def br_bl_corner(br: sd, bl: sd):
-    with computation(PARALLEL), interval(...):
-        bl = 0
-        br = 0
+def _compute_stencil(
```
mcgibbon, line 49: Please add docstring.

fv3core/stencils/ytp_v.py
```diff
@@ -1,230 +1,165 @@
...
+    c: FloatField,
+    v: FloatField,
+    flux: FloatField,
+    dy: FloatField,
+    dya: FloatField,
+    rdy: FloatField,
+):
+    from __externals__ import i_end, i_start, j_end, j_start, jord, namelist
```
mcgibbon, line 57: Similarly to hord, I would suggest renaming jord after this line to be something more verbose. If not, a code comment would be great.

fv3core/stencils/yppm.py
```diff
@@ -229,306 +120,211 @@ def xt_dya_edge_1_base(q, dya):
...
-                    domain=x_edge_domain,
-                )
-        if jord < 0:
-            floor_cap(
-                al,
-                0.0,
-                origin=(ifirst, grid().js - 1, kstart),
-                domain=(ilast - ifirst + 1, grid().njc + 3, nk),
-            )
+def compute_al(q: FloatField, dya: FloatField):
```
mcgibbon, line 250: If you understand what al is, I'd really appreciate seeing it in a docstring.

mcgibbon (commented):
Only reviewed ytp_v, since I worked on it last Friday. Big improvement! Please add docstrings and some comments/renames for unclear variable names.

I suggest no change on this for now, but an outstanding readability thing i see is `bl` and `br`, I think they're used for two different meanings and could be isolated as temporaries in two different gtscript functions later. To do this we would need to actually understand what they mean.

fv3core/stencils/ytp_v.py
jdahm: Let's do this in a follow-up PR.

fv3core/stencils/ytp_v.py
jdahm: Let's do that in a follow-up PR.

fv3core/stencils/yppm.py
jdahm: Done!

fv3core/stencils/ytp_v.py
jdahm: This one I'd argue does not need one because the public method below simply calls this one immediately, but I could be persuaded I suppose.

mcgibbon: Docstrings are needed right now for our internal development purposes. We don't know what will or won't be public down the line. As the structure settles we can start deleting duplicate/unnecessary/outdated docstrings, but for now the more, the better.

This stencil could be called by code other than the public function below. If a code author reads down the call chain from somewhere else and gets here, it's hard for them to know they should look at some other public function which calls this stencil to see documentation for the inputs.

fv3core/stencils/ytp_v.py
mcgibbon: Yes, we've chatted about this and it seems like something which should be done throughout the affected code, and would be a larger PR to do on its own.

fv3core/stencils/ytp_v.py
```diff
@@ -1,230 +1,166 @@
...
+
+            with horizontal(region[:, j_end - 1]):
+                bl, br = yppm.north_edge_jord8plus_0(v, dm, al)
+                bl, br = yppm.pert_ppm_standard_constraint_fcn(v, bl, br)
+
+            with horizontal(region[:, j_end]):
+                bl, br = yppm.north_edge_jord8plus_1(v, dy, dm)
+
+            with horizontal(region[:, j_end + 1]):
+                bl, br = yppm.north_edge_jord8plus_2(v, dy, dm)
```
mcgibbon, line 113: In a later PR, maybe we can join these into a single function `blbr_edge_correction_jord8plus`, and maybe it could be called within `blbr_jord8`.

mcgibbon (approved):
Everything looks good from my end. Only optional comments and suggestions. May want to add a PR description to use for the merge commit, before merging.

fv3core/stencils/ytp_v.py
```diff
@@ -1,230 +1,166 @@
...
+
+def compute(c: FloatField, u: FloatField, v: FloatField, flux: FloatField):
+    """
+    Compute flux of kinetic energy in y-dir.
+
+    Notes:
+        u is passed in here, but is unused in the stencil.
+
+    Args:
+        c (in): Courant number in flux form
```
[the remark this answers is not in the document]
jdahm: 👍

jdahm:
launch jenkins

rheacangeo:
launch jenkins

rheacangeo:
launch jenkins

rheacangeo:
launch jenkins

tests/translate/translate_xppm.py
```diff
@@ -33,7 +33,7 @@ def process_inputs(self, inputs):
 
     def compute(self, inputs):
         self.process_inputs(inputs)
-        inputs["xflux"] = utils.make_storage_from_shape(inputs["q"].shape, xppm.origin)
+        inputs["xflux"] = utils.make_storage_from_shape(inputs["q"].shape, origin=(2, 0, 0))
```
rheacangeo, line 36: I prefer if we use the grid variable names rather than hard coded numbers, e.g. (grid.is_ - 1, grid.jsd, 0). We have started using harded coded 'add' tuples, which could be used as well (note there is an open PR changing the names of some of the grid shorthands)

tests/translate/translate_yppm.py
```diff
@@ -30,7 +30,7 @@ def ivars(self, inputs):
     def process_inputs(self, inputs):
         self.ivars(inputs)
         self.make_storage_data_input_vars(inputs)
-        inputs["flux"] = utils.make_storage_from_shape(inputs["q"].shape, yppm.origin)
+        inputs["flux"] = utils.make_storage_from_shape(inputs["q"].shape, origin=(0, 2, 0))
```
rheacangeo, line 33: same comment as for xppm

rheacangeo (commented):
This looks pretty amazing, I am super excited for this! I need to look over this once more but is probably almost ready to go!

fv3core/stencils/xppm.py
```diff
@@ -168,236 +106,247 @@ def xt_dxa_edge_1_base(q, dxa):
...
-                domain=(grid().nic + 3, jlast - jfirst + 1, nk),
-            )
-    return al
+@gtscript.function
+def compute_al(q: FloatField, dxa: FloatField):
+    """
+    Interpolate q at interface.
 
+    Inputs:
+        q: Tracer
```
rheacangeo, line 216: I think q is not always a tracer, should probably label it the same as you have for the outer function "Transported scalar"

fv3core/stencils/xtp_u.py
```diff
@@ -1,287 +1,171 @@
...
-def get_flux_u_ord8plus(q: sd, c: sd, rdx: sd, bl: sd, br: sd, flux: sd):
-    with computation(PARALLEL), interval(...):
-        b0 = get_b0(bl, br)
-        cfl = c * rdx[-1, 0, 0] if c > 0 else c * rdx
-        fx1 = fx1_fn(cfl, br, b0, bl)
-        flux = q[-1, 0, 0] + fx1 if c > 0.0 else q + fx1
+        if smt5[-1, 0, 0]:
+            tmp = smt5[-1, 0, 0]
+        else:
+            tmp = smt5[-1, 0, 0] + smt5[0, 0, 0]
```
rheacangeo, line 51: This is duplicated in xppm, can we use a function like we did before?

fv3core/stencils/xtp_u.py
```diff
@@ -1,287 +1,171 @@
...
+            br = al[1, 0, 0] - u[0, 0, 0]
 
+            # Zero corners
+            with horizontal(
+                region[i_start - 1 : i_start + 1, j_start],
+                region[i_start - 1 : i_start + 1, j_end + 1],
+                region[i_end : i_end + 2, j_start],
+                region[i_end : i_end + 2, j_end + 1],
+            ):
+                bl = 0.0
```
rheacangeo, line 86: This long if else block might be more readable if the cases were written in separate functions

fv3core/stencils/xppm.py
```diff
@@ -168,236 +106,247 @@ def xt_dxa_edge_1_base(q, dxa):
...
-        bl = s14 * dm[-1, 0, 0] + s11 * (q[-1, 0, 0] - q)
-        xt = xt_dxa_edge_0(q, dxa, xt_minmax)
-        br = xt - q
+@gtscript.function
+def west_edge_iord8plus_0(
+    q: FloatField,
+    dxa: FloatField,
+    dm: FloatField,
+):
+    bl = yppm.s14 * dm[-1, 0, 0] + yppm.s11 * (q[-1, 0, 0] - q)
```
rheacangeo, line 142: I don't know if yppm is the best place for these shared constants, I just ported yppm first and then reused parts of it. These might make more sense to come from fv_tp_2d, but probably also fine to leave as-is.

rheacangeo (commented):
Pretty fun. I made some comments about potential functions and duplicated code. I assume the comments made in the xppm/xtp_u also apply to the equivalent in the yppm/ytp_v stencils

fv3core/stencils/xppm.py
jdahm: yeah I had that thought too. Perhaps we should move most of this into a new `ppm.py` file instead.

fv3core/stencils/xtp_u.py
jdahm: Good point. This is mixing low-level code and function calls. I'll give this a refactor over the weekend.

jdahm:
launch jenkins

jdahm:
I'm going to leave the remainder of the points for a future pass through this code. For now I think it's good enough to merge.

PR #211, opened 2021-01-30
get_test_data auto-updates

rheacangeo:
Currently running 'make get_test_data' only checks whether you have data, not whether it is up to date with the FORTRAN_SERIALIZED_DATA_VERSION set in the Makefile. This does not work for those who maintain their git repos in place, as when the version updates upon a git pull, the data will not update. . 
This adds a simple check that md5sums of the local test data matches that in the cloud. If not, the local data is removed and the data version specified in the Makefile is downloaded.


Makefile
```diff
@@ -175,7 +175,8 @@ push_python_regressions:
 	gsutil -m cp -r $(TEST_DATA_HOST)/python_regressions $(REGRESSION_DATA_STORAGE_BUCKET)/$(FORTRAN_SERIALIZED_DATA_VERSION)/$(EXPERIMENT)/python_regressions
 
 get_test_data:
-	if [ ! -f "$(TEST_DATA_HOST)/input.nml" ]; then \
+	if [ ! -f "$(TEST_DATA_HOST)/input.nml" ] || [ "$$(gsutil cat $(REGRESSION_DATA_STORAGE_BUCKET)/$(FORTRAN_SERIALIZED_DATA_VERSION)/${EXPERIMENT}/md5sums.txt)" != "$$(cat $(TEST_DATA_HOST)/md5sums.txt)"  ]; then \
```
jdahm, line 178: is it possible to break this onto shorter lines and use backslashes?

rheacangeo: not sure -- Makefile syntax is uber picky, but it can probably be broken into multiple lines

ofuhrer:
Would a rsync instead of a copy work better?

rheacangeo:
Rsync would work better if the data was not tarred, but I don't think it necessarily helps with the small number of files. I could be wrong. rsync has been a little finnicky for me.

jdahm:
rsync only copies the diffs, which is nice. Either way...

rheacangeo:
yep that's true, though I'd rather get this in then go through another cycle of testing if that ok with you?

rheacangeo:
wait we do already use rsync. the cp you are seeing is for the python regressions, unrelated to the this PR, and not in use right now. In that case the amount of data was so tiny it didn't really matter whether cp or rsync were used. If we revive that approach, we should probably just store the regressions in github like the fortran repo does.

Makefile
```diff
@@ -171,13 +173,15 @@ run_tests_parallel:
...
+	gsutil -m cp -r $(TEST_DATA_HOST)/python_regressions $(DATA_BUCKET)python_regressions
 
 get_test_data:
-	if [ ! -f "$(TEST_DATA_HOST)/input.nml" ]; then \
+	if [ ! -f "$(TEST_DATA_HOST)/input.nml" ] || \
+	[ "$$(gsutil cat $(DATA_BUCKET)md5sums.txt)" != "$$(cat $(TEST_DATA_HOST)/md5sums.txt)"  ]; then \
+	rm -rf $(TEST_DATA_HOST) && \
 	$(MAKE) sync_test_data && \
 	$(MAKE) unpack_test_data ;\
 	fi
```
jdahm, line 187: Would be cool to indent blocks like this for readability next time we touch the Makefile.

jdahm (approved):
Looks good. This is going to save us a lot of hassle!

PR #213, opened 2021-02-02
Consolidate grid origin and domain calls using 'add' kwarg

mcgibbon:
## Purpose

There are many grid methods which provide variants of "give the compute domain origin or shape, plus/minus 1/number of halo points". It's not immediately clear from the existing names what these do. In later PRs, an `add` keyword argument was included for `grid.domain_shape_compute`, which more clearly communicates that it is providing the compute domain shape modulo (usually) 1.

This PR removes a large number of grid methods by replacing them with calls to `domain_shape_compute`, `domain_shape_full`, or `compute_origin`. It also renames `grid.default_origin` to `grid.full_origin` to describe it in contrast to `grid.compute_origin`.

Remove the sections below which do not apply.

## Code changes:

- `grid.default_origin` is renamed to `grid.full_origin`
- `grid.domain_shape_standard` is renamed to `grid.domain_shape_full`
- Removed methods `domain_shape_compute_buffer_2d`, `domain_shape_compute_buffer_k`, `domain_shape_compute_x`, `domain_shape_compute_y`, `domain_x_compute_y`, `domain_x_compute_ybuffer`, `domain_y_compute_x`, `domain_y_compute_xbuffer`, `domain_shape_buffer_cell`, `domain2d_ik_buffer_1cell`, `domain_shape_y`, `domain_shape_x`, `corner_domain`, and `domain_shape_buffer_2d` from Grid class. Replaced with calls to `domain_shape_full`, `domain_shape_compute`, or `compute_origin`.

## Checklist

- [x] Docstrings and type hints are added to new and updated routines, as appropriate (confirmed not mentioned in docs)


jdahm:
I understand this is not quite the scope of this PR, but why not specify a "grid" instead of origin and domain, both for stencils as well as storages? That way it is made clear that storage x is defined on the C grid, or that stencil is run over the C grid.

jdahm:
I'm totally fine with this being a step in that direction too. This already helps a ton with readability! Good work.

rheacangeo (approved):
Excellent improvement, reducing the number of 'special' names we need to be aware of and leading us away from the 'full' domain being the standard or default (as is in the Fortran code) as gt4py should in many cases compute to the right bounds with the compute domain being the central default focus.

mcgibbon:
> I understand this is not quite the scope of this PR, but why not specify a "grid" instead of origin and domain, both for stencils as well as storages? That way it is made clear that storage x is defined on the C grid, or that stencil is run over the C grid.

There isn't a 1-to-1 mapping between a grid and the origin/domain for stencil computation. For example, several stencils take variables which are on different grids, but the stencil is only run on one origin+domain. We've chatted before about possibly having the default origin+domain be set by the compute domain origin+extent of the first argument, if we use Quantity more extensively. We could consider that (or other similar options) in the future.

> Excellent improvement, reducing the number of 'special' names we need to be aware of and leading us away from the 'full' domain being the standard or default (as is in the Fortran code) as gt4py should in many cases compute to the right bounds with the compute domain being the central default focus.

Thanks!

jdahm:
@mcgibbon 

> There isn't a 1-to-1 mapping between a grid and the origin/domain for stencil computation. For example, several stencils take variables which are on different grids, but the stencil is only run on one origin+domain.

Agreed. Until we have much larger stencil, automatically determining the domain each time is not possible. I had imagined an intermediate step where we replace the `origin` and `domain` keyword arguments optionally with e.g. `domain_grid`. So, not quite automatically determining the domain from Quantities, but manually specifying it.

rheacangeo:
launch jenkins

rheacangeo:
launch jenkins

fv3core/stencils/dyn_core.py
```diff
@@ -149,19 +153,17 @@ def compute(state, comm):
...
             state.ak,
             state.bk,
             state.phis,
             state.dp_ref,
             state.zs,
             rgrav,
-            origin=grid.default_origin(),
-            domain=grid.domain_shape_standard(add=(0, 0, 1)),
+            origin=grid.full_origin(),
+            domain=grid.domain_shape_full(),
```
rheacangeo, line 166: I think you need an 'add=(0, 0, 1)' here - this may be why DynCore is failing the tests

rheacangeo (changes requested):
Parallel tests are failing -- I just found one potential correction needed, though there could be others.

fv3core/utils/grid.py
```diff
@@ -403,21 +361,23 @@ def overwrite_edges(self, var, edgevar, left_i_index, left_j_index):
     def compute_origin(self, add: Tuple[int, int, int] = (0, 0, 0)):
         return (self.is_ + add[0], self.js + add[1], add[2])
 
-    def default_origin(self):
+    def full_origin(self):
```
twicki, line 364: should we reflect the difference between compute full origin in a short docstring here?

twicki (approved):

fv3core/stencils/dyn_core.py
mcgibbon: That fixed it, thanks!

fv3core/utils/grid.py
mcgibbon: I added docstrings, let me know what you think.

twicki: looks good

twicki (approved):
lgtm

rheacangeo (approved):

rheacangeo:
launch jenkins

rheacangeo:
launch jenkins

mcgibbon:
The test are failing, it looks like I'll need to debug it.

mcgibbon:
Thank you @jdahm!

rheacangeo:
Bringing this PR back after reverting changes that broke gt backends

PR #214, opened 2021-02-03
remove equals and forward slashes from the JENKINS_TAG

rheacangeo:


rheacangeo:
launch jenkins

jdahm (approved):
From my experiments locally this _replaces_ instances of `=` and `/` with `-` throughout the string. As long as that is what was intended, it looks great! Thanks for the quick fix.

PR #218, opened 2021-02-04
remove --profile

rheacangeo:
## Purpose

Don't use the --profile with the pytest commands by default because this is making automated testing too slow.


jdahm (approved):
🙏

PR #219, opened 2021-02-04
Add automerge github action

mcgibbon:
## Purpose

Our tests take a little long to run, and there can be a delay between when a test finishes and when we remember to merge it.

This PR, once merged, allows you to label a github PR as "automerge". If you do, that PR will merge automatically when its tests pass if it has received github-required reviews. This will allow us to "set and forget" PRs which are ready for merge and have been reviewed, but have minor changes which re-trigger the tests.

## Infrastructure changes:

- Add the automerge github action, copied from the fv3net repository

## Checklist

It is very optional to use this, and I do not think we need to add an entry to the README about it.

### DIFF
--- a/.github/workflows/automerge.yaml
+++ b/.github/workflows/automerge.yaml
@@ -1,0 +1,29 @@
+name: automerge
+on:
+  pull_request:
+    types:
+      - labeled
+      - unlabeled
+      - synchronize
+      - opened
+      - edited
+      - ready_for_review
+      - reopened
+      - unlocked
+  pull_request_review:
+    types:
+      - submitted
+  check_suite:
+    types:
+      - completed
+  status: {}
+jobs:
+  automerge:
+    runs-on: ubuntu-latest
+    steps:
+      - name: automerge
+        uses: "pascalgn/automerge-action@f81beb99aef41bb55ad072857d43073fba833a98"
+        env:
+          GITHUB_TOKEN: "${{ secrets.GITHUB_TOKEN }}"
+          MERGE_COMMIT_MESSAGE: pull-request-title-and-description
+          MERGE_METHOD: squash


ofuhrer:
I like this a lot. I think it's a great idea to present it, since it essentially has two consequences from my point of view:
- Be serious about indicating that changes might be needed and wanting to re-review.
- Be serious about not adding additional features post-review but only addressing review comments.
From personal experience from my own PRs, I know that I have not _always_ been an example in this respect. Auto-merge will actually help improve this.

ofuhrer (approved):

rheacangeo:
If we do this, we should change the github review requirements if possible -- currently you can merge if 1 reviewer accepts, but if you have asked multiple reviewers for their opinion, you wouldn't want this to automerge to happen before all have had a chance to take a look. Also, I don't think it is too much work to follow up on your PRs and close them out manually. This will remind you to also update Jira.

rheacangeo:
Ah maybe I misunderstood the feature -- if making the PR automerge is a manual decision that you make only once you have received all reviews, that might be ok. I still think it's good practice to follow up on your tasks until they are really merged, but I won't oppose this feature if it is helpful for some cases.

rheacangeo:
The other concern would be whether pull-request-title-and-description is ok for the squash merge commit message

ofuhrer:
I fully understand Rhea's concerns. I think essentially we're saying the same thing: We would have to partially change our behavior (e.g. choose PR titles that are good squash merge messages). Generally, I feel like these changes in our behavior are going in the right direction, but presenting it to the team might give you a final answer in that respect.

mcgibbon:
> if making the PR automerge is a manual decision that you make only once you have received all reviews, that might be ok

Yes, that's how we do it in fv3net. It's worked pretty well. It only gets used for maybe 10-25% of PRs. For me, I only mark it in a situation where I wish I could merge the PR without running the tests, because I am 90% sure they're going to pass (e.g. when rebasing to the main branch at the end of the PR).

> The other concern would be whether pull-request-title-and-description is ok for the squash merge commit message

We could chat about this. On ML we've agreed to do this for every PR, not just automerge PRs. It works if you have a good PR template, which we do now. It also means reviewers can also review the commit message.

rheacangeo (approved):
Ok this seems like a good option to have. I suggest getting the team's reaction, and we also need to make sure this works even though repo does not require tests to pass to merge (e.g. that it won't automerge when tests fail because it can).

rheacangeo:
launch jenkins

rheacangeo:
launch jenkins

ofuhrer:
@mcgibbon @nbren12 Just saw that this is a third-party GitHub action. How concerned should we be about security? Pinning the action to a specific SHA goes some way, but I guess this action could have access to our secrets or could potentially write to our repos, correct?

nbren12:
I haven't thought about the security implications. That's a good point. The write access doesn't seem disturbing to me, but in principle it shouldn't need access to the keys. I don't know if github actions can prevent a workflow from using certain env variables for instance.

nbren12:
Re. the workflow of PRs with automerge. I think we mostly apply the "automerge" label to trivial changes or ones that have gone through a round of reviews. Basically, after the PR is "ready" and we're waiting for CI to finish/other PRs to merge.

mcgibbon:
@ofuhrer the external action (as you noted) is pinned by commit sha to [this commit](https://github.com/pascalgn/automerge-action/tree/f81beb99aef41bb55ad072857d43073fba833a98), so we could do a security audit of the files if we're concerned. We've been using it for months and have not run into issues. Since the repository is public and git allows reverting to previous commits (including restoring from our local copies), I think the blast radius is minimal in the worst case scenario. Github actions don't have access to CircleCI or Jenkins secrets.

My understanding is that all secrets need to be passed in by environment variables, and the only one we're giving it is an SSH token which grants access [only to the one repository](https://docs.github.com/en/actions/reference/authentication-in-a-workflow).

PR #221, opened 2021-02-04
update submodule daint_venv 

rheacangeo:
## Purpose

Update the daint_venv submodule so that we use the one that installs the latest stable gt4py:develop, not the latest one that has been built and might be broken

## Infrastructure changes:

- Update the daint_venv submodule to commit  5a465bcaa197721b6bd62b68baeb4423ad5c8e8c


twicki (approved):

ofuhrer:
We could abuse this PR to submodule venv and use dependabot to update it automagically. We've had pretty good experience with that strategy for fv3gfs-fortran and fv3gfs-util.

PR #224, opened 2021-02-05
remove commas from docker tags in jenkins

rheacangeo:
## Purpose

Commas also show up in multiconfiguration project JOB_NAMEs, so we should remove those from the tag as well to unblock the fv3core-docker-master-gtbackends jenkins plan.


rheacangeo:
launch jenkins

jdahm (approved):
Looks right to me!

PR #226, opened 2021-02-05
Change git hash to come in from the calling script

twicki:
## Purpose

Fix catch more kinds of errors when parsing the git hash.

## Code changes:

- Since daint can also throw os errors in parsing the git hash we now catch everything.


oelbert (approved):
Whoops

PR #209, opened 2021-01-30
Refactor methods for copying cell corners

jdahm:
* Merges `fill2_4corners` and `fill3_4corners` from d2a2c PR.
* Adds `mult`-based methods for `d2a2c_vect` and other modules.
* Affixes `_cells` to corner functions that are only meant to work on cell fields.

PR #228, opened 2021-02-05
Regions refactor: c_sw and d2a2c

jdahm:
## Purpose

Refactor d2a2c into single function, and c_sw into a single stencil.

To do:
- [x] Blocked by #209.

## Code changes:

- Code changes are self-contained inside c_sw and the functions it calls.


tests/translate/translate_d2a2c_vect.py
```diff
@@ -18,14 +69,48 @@ def __init__(self, grid):
...
-            "uc": grid.x3d_domain_dict(),
-            "vc": grid.y3d_domain_dict(),
+            # "uc": grid.x3d_domain_dict(),
+            # "vc": grid.y3d_domain_dict(),
             "ua": {},
             "va": {},
-            "utc": {},
-            "vtc": {},
+            # "utc": {},
+            # "vtc": {},
```
rheacangeo, line 77: aren't uc and vc modified here? shouldn't we still compare them? If for some reason not, we can remove the lines. same question for utc/vtc. This may have been from me when I was debugging this stencil

tests/translate/translate_d2a2c_vect.py
```diff
@@ -18,14 +69,48 @@ def __init__(self, grid):
...
         self.max_error = 2e-10
+
+    def compute(self, inputs):
+        if spec.namelist.npx != spec.namelist.npy:
+            raise NotImplementedError("D2A2C assumes a square grid")
+        if spec.namelist.npx <= 13 and spec.namelist.layout[0] > 1:
+            D2A2C_AVG_OFFSET = -1
+        else:
+            D2A2C_AVG_OFFSET = 3
+
```
rheacangeo, line 91: I'm so glad this works. Our tests currently don't cover everything here, so would you mind manually testing c12_54ranks and c48_54ranks one more time? I'd be happy to do this for you later if you'd like.

rheacangeo (approved):
I haven't looked super closely at this right now, but did the other week when was helping debug issues, so have strong confidence in it. As noted in comments, please correct the commented out outputs for d2a2c and try testing with c48_54ranks (or ask me to, I can tonight) before merging. This is so freaking exciting. Stunning job.

tests/translate/translate_d2a2c_vect.py
jdahm: Good find! I just tried it locally and these fields validate again. This may have been leftover as an intermediate step as we were having issues. I'm reverting them.

tests/translate/translate_d2a2c_vect.py
jdahm: Good idea. I'm running those now before merging the PR.

PR #230, opened 2021-02-05
Restrict JENKINS_TAG size to legal docker tag length

rheacangeo:
## Purpose

When the jenkins tag is too long, turn it into a 32 character hash string to prevent illegal docker tags (they can't be longer than 128 characters, and multiconfiguration projects can end up with really long job names)

## Infrastructure changes:

JENKINS_TAG in .jenkins/jenkins will include a md5sum of the JOB_NAME if it is too long to be a docker tag. This should re-enable multiconfiguration plans that don't take the VIRTUALENV or JENKINS_TAG as parameters. This for example, enables this: https://jenkins.ginko.ch/job/fv3core-regressions-cron/3/backend=numpy,experiment=c12_6ranks_standard,runaction=run_parallel_regression_tests,slave=gce-cpu/console

### DIFF
--- a/.jenkins/jenkins.sh
+++ b/.jenkins/jenkins.sh
@@ -143,7 +143,7 @@ export FV3_STENCIL_REBUILD_FLAG=False
 export TEST_DATA_HOST="${TEST_DATA_DIR}/${experiment}/"
 export EXPERIMENT=${experiment}
 if [ -z ${JENKINS_TAG} ]; then
-    if [ ${#JOB_NAME} -gt 75 ]; then
+    if [ ${#JOB_NAME} -gt 85 ]; then
 	NAME=`echo ${JOB_NAME} | md5sum | cut -f1 -d" "`
     else
 	NAME=${JOB_NAME}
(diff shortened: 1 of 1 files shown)


.jenkins/jenkins.sh
```diff
@@ -143,7 +143,12 @@ export FV3_STENCIL_REBUILD_FLAG=False
 export TEST_DATA_HOST="${TEST_DATA_DIR}/${experiment}/"
 export EXPERIMENT=${experiment}
 if [ -z ${JENKINS_TAG} ]; then
-    export JENKINS_TAG=${JOB_NAME//[,=\/]/-}-${BUILD_NUMBER}
+    if [ ${#JOB_NAME} > 122 ]; then
```
eddie-c-davis, line 146: Just for my understanding, the `md5sum` guarantees a job name length of 32, and the check is for a length >= 123 because our prefixes are ~5 characters, e.g., `daint`, or `gce`?

eddie-c-davis (approved):
One question to confirm my understanding, but looks good to me.

.jenkins/jenkins.sh
rheacangeo: The max docker tag is 128 characters. I picked 122 to leave enough characters for a really large unrealistic build number. We could probably have this be 124 or 125 and be fine, or could hash the whole thing with the build number. But i thought it would be nice if at least some part of the tag had useful information. With the multiconfiguration projects, we are getting tags like v0.3.0-gnu7-mpich314-nocuda-developfv3core-regressions-cron-backend-numpy-experiment-c12_6ranks_standard-runaction-run_regression_tests-slave-gce-cpu-2
Oh actually, I should lower that threshold more, to ~75!

eddie-c-davis: Good idea!

PR #235, opened 2021-02-11
Feature/regions a2b ord4

jdahm:
## Purpose

Rewrites a2b_ord4 as a single stencil.

## Code changes:

- `a2b_ord4` rewritten
- `float_type` -> `Float`, `int_type` -> `int`


jdahm:
Looks like we found a bug in the `numpy` backend. This works locally with `debug`. I'll take a look this afternoon or tomorrow.

jdahm:
launch jenkins

jdahm:
@rheacangeo  😆 
```
E               AssertionError: Failed count: 1/27018 (0.0%),
E                 first failed index [15, 3, 47], computed:-0.1783100709178987, reference: -0.17831007091790105, diff: 2.3592239273284576e-15
```

fv3core/utils/typing.py
```diff
@@ -21,9 +21,9 @@
 DTypes = Union[bool, np.bool, int, np.int32, np.int64, float, np.float32, np.float64]
 
 # Default float and int types
-float_type = np.float_
-int_type = np.int_
-bool_type = np.bool_
+Float = np.float_
+Int = np.int_
+Bool = np.bool_
```
rheacangeo, line 26: This makes sense given the FloatField name

fv3core/stencils/a2b_ord4.py
```diff
@@ -22,485 +36,435 @@
 sd = utils.sd
```
rheacangeo, line 36: We can delete this now :)

fv3core/stencils/a2b_ord4.py
```diff
@@ -22,485 +36,435 @@
...
-def grid():
-    return spec.grid
-
-
-@gtstencil()
-def ppm_volume_mean_x(qin: sd, qx: sd):
-    with computation(PARALLEL), interval(...):
-        qx[0, 0, 0] = b2 * (qin[-2, 0, 0] + qin[1, 0, 0]) + b1 * (qin[-1, 0, 0] + qin)
+@gtscript.function
+def ppm_volume_mean_x(qin: FloatField):
```
rheacangeo, line 40: Are we type-hinting the functions too? I realize I have reviewed and accepted other PRs you've opened with these haha. So no objection, just trying to remember when we changed our minds on that (previously we had suggested only type hinting stencils and some functions on a case-by-case basis). I haven't been adding type-hints to functions. We can probably do a style-consistency sweep at some point to pick up the stragglers.

fv3core/stencils/a2b_ord4.py
```diff
@@ -22,485 +36,435 @@
...
-    with computation(PARALLEL), interval(...):
-        qout[0, 0, 0] = 0.5 * (qxx + qyy)
 
+@gtscript.function
+def qout_x_edge(
+    qin: FloatField, dxa: FloatFieldIJ, edge_w: FloatFieldIJ, q2: FloatField
+):
+    """
+    Note:
+        q2 could be 2D, but it is a temporary currently which are all 3D.
```
rheacangeo, line 85: temporaries can be local scalars sometimes too I thought? Does type-hinting force it to be a field? Probably note, the reduction to a scalar probably happens later after inlining I would assume. So from a front-end perspective this note is correct.

fv3core/stencils/a2b_ord4.py
```diff
@@ -22,485 +36,435 @@
...
-        b = 1
-    else:
-        a = -1
-        b = -2
-    return (a, b)
-
-
-def ec2_offsets_dirs(corner, lower_direction, other_direction):
-    if lower_direction in corner or other_direction in corner:
-        a = -1
```
rheacangeo, line 232: Wow! Great job unraveling all of these python functions!

fv3core/stencils/a2b_ord4.py
```diff
@@ -22,485 +36,435 @@
...
+        domain=(grid.nic + 1, grid.njc + 1, nk),
+    )
+
+    # if replace:
+    #     copy_stencil(
+    #         qout,
+    #         qin,
+    #         origin=(grid.is_, grid.js, kstart),
+    #         domain=(grid.nic + 1, grid.njc + 1, nk),
+    #     )
```
rheacangeo, line 470: Is this intended to be commented out or was that for debugging?

fv3core/stencils/a2b_ord4.py
```diff
@@ -22,485 +36,435 @@
...
+        backend=global_config.get_backend(),
+        default_origin=grid.compute_origin()[:-1],
+        shape=grid_array[:, :, index].shape,
+        dtype=grid_array.dtype,
+        mask=(True, True, False),
     )
 
 
-def compute(qin, qout, kstart=0, nk=None, replace=False):
+def compute(
```
rheacangeo, line 424: Would be nice to start a docstring on this one (at least with just a TODOC) since it gets called in so many different ways.

fv3core/stencils/a2b_ord4.py
```diff
@@ -22,485 +36,435 @@
...
-                origin=(grid().is_, grid().js, kstart),
-                domain=(grid().ie - grid().is_ + 2, grid().je - grid().js + 2, nk),
-            )
-    else:
-        raise Exception("grid_type >= 3 is not implemented")
+        nk = grid.npz - kstart
+    agrid1 = _make_grid_storage_2d(grid.agrid1)
+    agrid2 = _make_grid_storage_2d(grid.agrid2)
+    bgrid1 = _make_grid_storage_2d(grid.bgrid1)
+    bgrid2 = _make_grid_storage_2d(grid.bgrid2)
```
rheacangeo, line 437: So these could all be temporary storages inside the stencil, but we can't have 2d storages, so I can see why you'd keep it outside the stencil call. But what is worse, allocating new 2d storages every timestep and passing them, or having the stencil allocate 3d storages on the fly but taking advantage of the 'wheel' (or will the 'wheel' be available in this scope too? my impression was no?).

rheacangeo (commented):
Awesome work! This one was a beast with all of those corners and then functions calling functions calling functions, excellent job. I have a few minor questions and comments, but otherwise approve.

fv3core/stencils/a2b_ord4.py
rheacangeo: Oh I see now, can be deleted, awesome

fv3core/stencils/a2b_ord4.py
jdahm: I've started adding them because once we have 2D arrays it will be somewhat necessary to keep track of things (although it should work with or without type hints). I thought it's confusing for the reader to see only 2 indices on some fields without the type hint telling the user that's really supposed to be a 2D field.

fv3core/stencils/a2b_ord4.py
jdahm: Yes good point, I'll make that change in the docstring. That was a note to myself and future refactors that because `q2` is accessed with an offset in this function, it can't be a scalar, so it's a 3D field right now, and technically a 2D field is sufficient.

Of course now that I look at it more closely, since we're in a parallel computation, it actually HAS to be 3D. So, I'll remove this comment altogether!

fv3core/stencils/a2b_ord4.py
jdahm: Thanks! It was a brain twister.

fv3core/stencils/a2b_ord4.py
jdahm: Ah, I forgot to delete it after adding back the external. Good catch!

fv3core/stencils/a2b_ord4.py
jdahm: These should be just views on the 3D data for the sake of making the stencils using `FloatFieldIJ` in the gtscript. This shouldn't have a large impact, and can be removed when the grid storage refactor is complete.

fv3core/stencils/a2b_ord4.py
jdahm: Good idea. I'll write one!

fv3core/stencils/a2b_ord4.py
rheacangeo: Ah gotchya, thanks for the clarification

rheacangeo (approved):

PR #236, opened 2021-02-11
run warmup on one node and multinode after

twicki:
## Purpose

- This PR changes from using the pre-installed vcm_venv on daint to installing it's own for performance runs.
- To get fair performance numbers, we do a compilation run first on one node and move to a multi-node execution with the already compiled code afterwards

## Code changes:

- only changes how run_on_daint is executing.

### DIFF
--- a/examples/standalone/benchmarks/run_on_daint.sh
+++ b/examples/standalone/benchmarks/run_on_daint.sh
@@ -112,10 +112,10 @@ echo "submitting script to do performance run"
 sed -i s/\<NAME\>/standalone/g run.daint.slurm
 sed -i s/\<NTASKS\>/$ranks/g run.daint.slurm
 sed -i s/\<NTASKSPERNODE\>/1/g run.daint.slurm
-sed -i s/\<CPUSPERTASK\>/12/g run.daint.slurm
+sed -i s/\<CPUSPERTASK\>/$nthreads/g run.daint.slurm
 sed -i s/--output=\<OUTFILE\>/--hint=nomultithread/g run.daint.slurm
 sed -i s/00:45:00/00:30:00/g run.daint.slurm
-sed -i s/cscsci/debug/g run.daint.slurm
+sed -i s/cscsci/normal/g run.daint.slurm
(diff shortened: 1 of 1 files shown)


examples/standalone/benchmarks/run_on_daint.sh
```diff
@@ -83,19 +88,39 @@ else
...
+wait
+echo "compilation step finished"
+rm *.json
+
+echo "submitting script to do performance run"
+# Adapt batch script to run the code:
+sed -i s/\<NAME\>/standalone/g run.daint.slurm
+sed -i s/\<NTASKS\>/$ranks/g run.daint.slurm
+sed -i s/\<NTASKSPERNODE\>/1/g run.daint.slurm
+sed -i s/\<CPUSPERTASK\>/12/g run.daint.slurm
```
eddie-c-davis, line 115: Not a big deal, but should this be using $nthreads?

examples/standalone/benchmarks/run_on_daint.sh
```diff
@@ -83,19 +88,39 @@ else
...
+sed -i s/--output=\<OUTFILE\>/--hint=nomultithread/g run.daint.slurm
+sed -i s/00:45:00/00:30:00/g run.daint.slurm
+sed -i s/cscsci/debug/g run.daint.slurm
```
eddie-c-davis, line 118: Does the `debug` queue work with multiple nodes?

eddie-c-davis (approved):
Looks good, but have a couple small questions...

examples/standalone/benchmarks/run_on_daint.sh
twicki: yeah - oversight. Changed

examples/standalone/benchmarks/run_on_daint.sh
twicki: it does work. But I just compared with our old data where a timestep is still close to 45' with the c128 data, i switched it to normal.

eddie-c-davis (approved):
Looks great!

PR #238, opened 2021-02-12
Addressed "TODOs" in Saturation stencil

gmao-ckung:
## Purpose

PR addresses the "TODOs" within the Saturation stencil, including removal of temporaries, incorporating function calls within conditionals when appropriate, using min/max functions when appropriate, and using function calls where previous versions of GT4Py would not allow.  Also, Field-type arguments for stencils are declared using "FloatField" from "fv3core.utils.typing" instead of using "utils.sd" from "fv3core.utils.gt4py_utils".

## Code changes:

Code changes are within /fv3core/stencils/saturation_adjustment.py

### DIFF
--- a/fv3core/stencils/saturation_adjustment.py
+++ b/fv3core/stencils/saturation_adjustment.py
@@ -60,6 +60,7 @@ def q_table_oneline(delta_heat_capacity, latent_heat_coefficient, tem):
         / constants.RVGAS
     )
 
+
 @gtscript.function
 def table_vapor_oneline(tem):
     return q_table_oneline(DC_VAP, LV0, tem)
@@ -210,6 +211,7 @@ def melt_cloud_ice(
 def minmax_tmp_h20(qa, qb):
     tmpmax = max(qb, 0.0)
     return min(-qa, tmpmax)
+
 
 @gtscript.function
 def fix_negative_snow(qs, qg):
--- a/fv3core/stencils/saturation_adjustment.py
+++ b/fv3core/stencils/saturation_adjustment.py
@@ -9,13 +9,13 @@ import fv3core.utils.gt4py_utils as utils
 from fv3core.decorators import gtstencil
 from fv3core.stencils.basic_operations import dim
 from fv3core.stencils.moist_cv import compute_pkz_func
+from fv3core.utils.typing import FloatField
 
 
 # TODO: This code could be reduced greatly with abstraction, but first gt4py
 # needs to support gtscript function calls of arbitrary depth embedded in
 # conditionals.
-sd = utils.sd
-si = utils.si
+
 DC_VAP = constants.CP_VAP - constants.C_LIQ  # - 2339.5, isobaric heating / cooling
 DC_ICE = constants.C_LIQ - constants.C_ICE  # 2213.5, isobaric heating / cooling
 LV0 = (
@@ -50,45 +50,30 @@ def tem_upper(i):
     return 253.16 + DELT * i
 
 
-# TODO this abstraction is not possible in gt4py as these are called in
-# conditionals and temporaries get created. but would be nice to have 1 version
-# of the equation.
-# {
-# @gtscript.function
-# def q_table_oneline(delta_heat_capacity, latent_heat_coefficient, tem):
-#     return E00 * exp((delta_heat_capacity * log(tem / TICE)
-#         + (tem - TICE) / (tem * TICE) *  latent_heat_coefficient) / constants.RVGAS)
-
-# @gtscript.function
-# def table_vapor_oneline(tem):
-#     return q_table_oneline(DC_VAP, LV0, tem)
-
-
-# @gtscript.function
-# def table_ice_oneline(tem):
-#     return q_table_oneline(D2ICE, LI2, tem)
-# }
-
+@gtscript.function
+def q_table_oneline(delta_heat_capacity, latent_heat_coefficient, tem):
+    return E00 * exp(
+        (
+            delta_heat_capacity * log(tem / TICE)
+            + (tem - TICE) / (tem * TICE) * latent_heat_coefficient
+        )
+        / constants.RVGAS
+    )
 
 @gtscript.function
 def table_vapor_oneline(tem):
-    return E00 * exp(
-        (DC_VAP * log(tem / TICE) + (tem - TICE) / (tem * TICE) * LV0) / constants.RVGAS
-    )
+    return q_table_oneline(DC_VAP, LV0, tem)
 
 
 @gtscript.function
 def table_ice_oneline(tem):
-    return E00 * exp(
-        (D2ICE * log(tem / TICE) + (tem - TICE) / (tem * TICE) * LI2) / constants.RVGAS
-    )
+    return q_table_oneline(D2ICE, LI2, tem)
 
 
 # TODO Math can be consolidated if we can call gtscript functions from
 # conditionals, fac0 and fac2 functions and others.
 @gtscript.function
 def qs_table_fn(i):
-    table = 0.0
     tem_l = tem_lower(i)
     tem_u = tem_upper(i - 1400)
     if i < 1600:
@@ -97,9 +82,6 @@ def qs_table_fn(i):
         table = 0.0
     if i >= 1600 and i < (1400 + 1221):
         table = table_vapor_oneline(tem_u)
-    wice = 0.0
-    wh2o = 0.0
-    esupc = 0.0
     if i >= 1400 and i < 1600:
         esupc = table_vapor_oneline(tem_u)
         wice = 0.05 * (TICE - tem_u)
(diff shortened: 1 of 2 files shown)


fv3core/stencils/saturation_adjustment.py
```diff
@@ -50,45 +50,31 @@ def tem_upper(i):
...
-        (D2ICE * log(tem / TICE) + (tem - TICE) / (tem * TICE) * LI2) / constants.RVGAS
-    )
+    return q_table_oneline(D2ICE, LI2, tem)
 
 
 # TODO Math can be consolidated if we can call gtscript functions from
 # conditionals, fac0 and fac2 functions and others.
 @gtscript.function
 def qs_table_fn(i):
-    table = 0.0
```
rheacangeo, line 91: good catch, we no longer need to 'declare' the temporary variables before conditionals to ensure they exist and have a value regardless of the branching conditions? Since temporaries are sometimes just local scalars, I think this works, pretty helpful. Is it ok to do this? The coder just needs to make sure if they use the temporary after the condition, it got assigned to something in one of the branches

fv3core/stencils/saturation_adjustment.py
```diff
@@ -113,49 +96,41 @@ def qs_table_fn(i):
...
         table2 = table_vapor_oneline(tem0)
-    table2_m1 = 0.0
-    table2_p1 = 0.0
     # TODO is there way to express the code below with something closer to this?:
     # if i == 1599:
     #    table2 = 0.25 * (table2(i-1) + 2.0 * qs_table_fn(i) + qs_table2_fn(i+1))
     # if i == 1600:
     #    table2 = 0.25 * (table2(i-1) + 2.0 * qs_table_fn(i) + qs_table2_fn(i+1))
-    table = 0.0
+    # table = 0.0
```
rheacangeo, line 110: can remove this line

fv3core/stencils/saturation_adjustment.py
```diff
@@ -113,49 +96,41 @@ def qs_table_fn(i):
...
     else:
         # compute es over water between 0 deg c and 102 deg c.
         table2 = table_vapor_oneline(tem0)
-    table2_m1 = 0.0
-    table2_p1 = 0.0
     # TODO is there way to express the code below with something closer to this?:
     # if i == 1599:
     #    table2 = 0.25 * (table2(i-1) + 2.0 * qs_table_fn(i) + qs_table2_fn(i+1))
     # if i == 1600:
     #    table2 = 0.25 * (table2(i-1) + 2.0 * qs_table_fn(i) + qs_table2_fn(i+1))
```
rheacangeo, line 109: Do these work? (might not be correct). Probably not, though what's below looks closer. We can probably delete this comment

fv3core/stencils/saturation_adjustment.py
```diff
@@ -113,49 +96,41 @@ def qs_table_fn(i):
...
+            0.05 * (tem0 - 253.16)
+        ) * table_vapor_oneline(tem0)
         # table2(i - 1)
-        tem0 = TMIN + DELT * 1598  # tem_lower(1598)
+        tem0 = tem_lower(1598)
         table2_m1 = table_ice_oneline(tem0)
         # table2(i + 1)
-        tem0 = TMIN + DELT * 1600  # tem_lower(1600)
+        tem0 = tem_lower(1600)
         table2_p1 = table_vapor_oneline(tem0)
```
rheacangeo, line 123: Can these be merged into 1 statement currently? Or do we need a temporary assigned to pass to the other function?

fv3core/stencils/saturation_adjustment.py
```diff
@@ -199,9 +167,8 @@ def des2_table(i):
 # - 2].
 @gtscript.function
 def desw_table(i):
-    t_p1 = qs_tablew_fn(i + 1)
     t = qs_tablew_fn(i)
-    diff = t_p1 - t
+    diff = qs_tablew_fn(i + 1) - t
```
rheacangeo, line 171: nice

fv3core/stencils/saturation_adjustment.py
```diff
@@ -246,47 +209,42 @@ def melt_cloud_ice(
...
-# Fix negative snow with graupel or graupel with available snow.
-# TODO Fix so only compute tmp when it is needed
 @gtscript.function
 def fix_negative_snow(qs, qg):
-    tmp = minmax_tmp_h20(qg, qs)
     if qs < 0.0:
         qg = qg + qs
         qs = 0.0
     elif qg < 0.0:
+        tmp = minmax_tmp_h20(qg, qs)
```
rheacangeo, line 222: it really is nice we don't have to define this for every branch when it's only used in one.

fv3core/stencils/saturation_adjustment.py
```diff
@@ -459,9 +390,7 @@ def sublimation(
             src = sink if sink < maxtmp else maxtmp
             src = src if src < tmp / tcp2 else tmp / tcp2
         else:
-            dimtmp = (
-                pt1 - t_sub if (pt1 - t_sub) > 0.0 else 0.0
-            )  # dim(pt1, t_sub) * 0.2 TODO WHY DOES THIS NOT WORK
+            dimtmp = dim(pt1, t_sub)
```
rheacangeo, line 393: should we double check where the * 0.2 comes from in the comment? could be a bug in the comment or in the code

fv3core/stencils/saturation_adjustment.py
```diff
@@ -484,10 +413,8 @@ def update_latent_heat_coefficient_i(pt1, cvm):
 @gtscript.function
 def update_latent_heat_coefficient(pt1, cvm, lv00, d0_vap):
     lhl = lv00 + d0_vap * pt1
-    lhi = LI00 + DC_ICE * pt1
     lcp2 = lhl / cvm
```
rheacangeo, line 416: These other two could be abstracted into a parallel function to grab lhl and lcp2

fv3core/stencils/saturation_adjustment.py
```diff
@@ -618,13 +544,22 @@ def wqs1_fn_2(it, ap1, ta, den):
 
 
 @gtstencil()
-def wqs2_stencil_w(ta: sd, den: sd, wqsat: sd, dqdt: sd):
+def wqs2_stencil_w(
+    ta: FloatField, den: FloatField, wqsat: FloatField, dqdt: FloatField
+):
     with computation(PARALLEL), interval(...):
         wqsat, dqdt = wqs2_fn_w(ta, den)
```
rheacangeo, line 551: I think we can delete this now? is anything using it anymore?

fv3core/stencils/saturation_adjustment.py
```diff
@@ -527,7 +453,7 @@ def wqsat_correct(src, pt1, lhl, qv, ql, q_liq, q_sol, mc_air, c_vap):
 
 
 @gtstencil()
-def ap1_stencil(ta: sd, ap1: sd):
+def ap1_stencil(ta: FloatField, ap1: FloatField):
     with computation(PARALLEL), interval(...):
         ap1 = ap1_for_wqs2(ta)
```
rheacangeo, line 458: I think we can delete this? Is there anything left that is using it?

rheacangeo (commented):
Awesome! It's so cool to see this one joining the new generation of style and features. I just have a few comments and questions. Also, due to the large number of branching conditionals here, and that we recently 'turned off' a number of the datasets being tested by CI, it would be great to manually test this on a few other datasets before merging so we have more confidence we have tested most of the branches -- , like c48_6ranks_standard, c48_6ranks_baroclinic, c12_54ranks_standard (we only run the parallel tests right now). I am willing to help with that so let me know. 
Great work!

fv3core/stencils/saturation_adjustment.py
gmao-ckung: I think this is fine since the value of `table` will be assigned at least in one of the first two conditions.  A better fix that gets rid of a conditional is to initially declare `table = 0.0` and get rid of the first `if i >= 1600...` conditional since it wouldn't be needed.

fv3core/stencils/saturation_adjustment.py
gmao-ckung: Done!

fv3core/stencils/saturation_adjustment.py
gmao-ckung: I tried these and they didn't work. The assignment to `table2` makes a recursive call to `qs_table2_fn`, and my thought was that GT4Py wasn't capable of executing recursive code since the error says `Unknown symbol "qs_table2_fn"`

fv3core/stencils/saturation_adjustment.py
gmao-ckung: The temporary is still needed since GT4Py doesn't support function calls as arguments to function calls.

` gt4py.frontend.gtscript_frontend.GTScriptSyntaxError: Function calls are not supported in arguments to function calls`

fv3core/stencils/saturation_adjustment.py
gmao-ckung: In my branch, `dimtmp` gets multiplied by 0.2 when computing `pidep`.  Perhaps one change I can do is to move the multiplication of 0.2 to `dimtmp` and then remove the multiplication of 0.2 from `pidep`.

fv3core/stencils/saturation_adjustment.py
gmao-ckung: I didn't find anything using it, so I can delete it.

fv3core/stencils/saturation_adjustment.py
gmao-ckung: Nothing's using it so I'll delete this as well.

fv3core/stencils/saturation_adjustment.py
gmao-ckung: Ok, I can create two different functions for `lhl` and `lcp2`.

fv3core/stencils/saturation_adjustment.py
rheacangeo: either way is good by me! I'm so happy gt4py allows you to choose now

fv3core/stencils/saturation_adjustment.py
rheacangeo: Wow, what a helpful error message! this is great.

fv3core/stencils/saturation_adjustment.py
rheacangeo: ahh -- ok it's just an outdated comment there, awesome, can leave as is

fv3core/stencils/saturation_adjustment.py
rheacangeo: Just an idea to make the function operation depth look flatter, but not a requirement, totally up to you.

rheacangeo: But yes, that's what I meant

rheacangeo (approved):
Awesome!

PR #241, opened 2021-02-14
fxadv is a single stencil

rheacangeo:
## Purpose

We are reducing the number of stencils by using the new gt4py 'regions' feature. Here the 'fxadv' set of calculations (which compute fluxes prior to calling fv_tp_2d, the first step in d_sw) have been merged into a single stencil.
## Code changes:
 * edge stencils have been converted to gtscript functions and combined 
 * The corner calculations previously reused a single stencil definition with varying externals, but this does not work within the context of a single stencil. So instead, we now explicitly compute all those corner values. The duplication might be hard to look at, but without being able to use a runtime integer as an offset to a field, we need it. A benefit to it is that it makes the validation better, as we no longer get shuffled order of operations using a single version of the equation. I prefer to preserve the single stencil version of the code at the bottom of the file for now. The only reason switching back to explicit corners was easy, was that I had the original version commented out in the file. 
 * Code that calls fxadv, e.g. d_sw and the test code, now only have access to the fxadv_stencil, and there is no python 'compute' function
 * The declaration of ra_x and ra_y has been moved to d_sw, as it no longer served a purpose to be declared inside of fxadv

## Checklist


fv3core/stencils/fxadv.py
```diff
@@ -1,314 +1,381 @@
...
+            xfx_adv = dy * prod * sin_sg3[-1, 0, 0] if prod > 0 else dy * prod * sin_sg1
+    with computation(PARALLEL), interval(...):
+        with horizontal(region[local_is : local_ie + 2, :]):
+            ra_x = ra_x_func(area, xfx_adv)
+    with computation(PARALLEL), interval(...):
+        with horizontal(region[:, local_js : local_je + 2]):
+            prod = dt * vt
+            cry_adv = prod * rdya[0, -1, 0] if prod > 0 else prod * rdya
+            yfx_adv = dx * prod * sin_sg4[0, -1, 0] if prod > 0 else dx * prod * sin_sg2
+    with computation(PARALLEL), interval(...):
```
rheacangeo, line 79: Although we strive to keep code at the same depth level, here that would mean creating gtscript functions for the horizontal regions specified directly here.  But, I found that hindered the readability, as it required passing through many variables that get listed one at a time, and it actually helps to see it here. I could unroll the functions that compute ut, it just then makes for a very long stencil definition, especially with the corners... I am open to discussion on what is most readable.

fv3core/stencils/fxadv.py
```diff
@@ -1,314 +1,381 @@
...
+    #    )
+    # original bounds with stencil calls
+    # j1 = grid().js + 2 if grid().south_edge else grid().js
+    # j2 = grid().je if grid().north_edge else grid().je + 2
+    # TODO: this is a hack, copying vt to vtmp to 'correct' the edges
+    # Can we *just* apply edge calculations in the correct regions without overcomputing
+    # rank 0, 1, 2: local_js + 2:local_je + 2
+    # rank 3, 4, 5: local_js:local_je + 2
+    # rank 6, 7, 8: local_js:local_je
+    vtmp = vt
```
eddie-c-davis, line 160: Same question as above...

eddie-c-davis (approved):
Fantastic work, a couple of small questions, but quite an improvement!

jdahm:
Great work! One convention I see here that other stencils are not using, is to define the stencil at the top of the file, before the definition of the gtscript functions that it calls. Should this be used more often, or should we move the stencil definition to the bottom of the file? I'm open to either.

fv3core/stencils/fxadv.py
```diff
@@ -1,314 +1,381 @@
...
-    # fill in for j/=npy-1 and j /= npy
-    if grid().north_edge:
-        ut[:, grid().je : grid().je + 2, :] = ut_in[:, grid().je : grid().je + 2, :]
-    return ut
 
+@gtscript.function
+def ut_corners(uc, vc, cosa_u, cosa_v, ut, vt):
+    from __externals__ import i_end, i_start, j_end, j_start
+
+    """
```
jdahm, line 189: I believe the convention is that the docstring is the first thing in the function.

fv3core/stencils/fxadv.py
```diff
@@ -1,314 +1,381 @@
...
+@gtscript.function
+def ra_y_func(area, yfx_adv):
+    return area + yfx_adv - yfx_adv[0, 1, 0]
 
 
-# -------------------- CORNERS-----------------
+# -------------------- DEPRECATED CORNERS-----------------
+"""
+# Using 1 function with different sets of externals
+# Unlikely to use as different externals in single stencil version
```
jdahm, line 340: Not sure how I feel about this, but this could be better as a comment instead of block string.

fv3core/stencils/fxadv.py
```diff
@@ -1,314 +1,381 @@
...
+        vt = vt_corners(uc, vc, cosa_u, cosa_v, ut, vt)
+    with computation(PARALLEL), interval(...):
+        with horizontal(region[local_is : local_ie + 2, :]):
+            prod = dt * ut
+            crx_adv = prod * rdxa[-1, 0, 0] if prod > 0 else prod * rdxa
+            xfx_adv = dy * prod * sin_sg3[-1, 0, 0] if prod > 0 else dy * prod * sin_sg1
+    with computation(PARALLEL), interval(...):
+        with horizontal(region[local_is : local_ie + 2, :]):
+            ra_x = ra_x_func(area, xfx_adv)
+    with computation(PARALLEL), interval(...):
```
jdahm, line 74: Seeing the multiple computation and intervals at first made me think there is vertical dependence in this stencil, but I can't find any. Could you put this all into the same computation interval? Shouldn't matter from a correctness perspective, since it's merged anyway in the compilation process.

rheacangeo: Good point, these different computations were previously in different stencils, and do not actually need to be different computations

fv3core/stencils/fxadv.py
rheacangeo: Thanks! Do you think I should move some (or all) of the ut regions into the stencil directly? They also don't have a reuse purpose, it was just a lot of code.

fv3core/stencils/fxadv.py
```diff
@@ -1,314 +1,381 @@
...
-
+@gtscript.function
+def main_vt(uc, vc, cosa_v, rsin_v, vt):
+    from __externals__ import j_end, j_start, local_je, local_js
 
-@gtstencil()
-def main_vt(uc: sd, vc: sd, cosa_v: sd, rsin_v: sd, vt: sd):
-    with computation(PARALLEL), interval(...):
-        vt[0, 0, 0] = (
+    vtmp = vt
```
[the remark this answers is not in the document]
rheacangeo: Yeah -- maybe there's a better way to do this? With regions it's easy to apply a corner calculation to an expanded domain, but I don't know how to say -- in this region don't compute when we are on an edge...

fv3core/stencils/fxadv.py
```diff
@@ -1,314 +1,381 @@
...
+    return ut
 
 
 @gtscript.function
-def ra_x_func(area, xfx_adv):
-    return area + xfx_adv - xfx_adv[1, 0, 0]
+def ut_x_edge(uc, cosa_u, vt, ut):
+    from __externals__ import i_end, i_start, j_end, j_start, local_ie, local_is
+
+    # TODO: parallel to what done for the vt_y_edge section
```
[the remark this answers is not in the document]
rheacangeo: still todo -- it's the same thing where I make a temporary variables to preserve the original value, apply a computation over the largest area possible, then 'undo' that on the edges...

fv3core/stencils/fxadv.py
rheacangeo: Yeah I left the todo because I don't know if there is a good way to not have that temporary field copied from the vt... Maybe it isn't that impactful?

fv3core/stencils/fxadv.py
rheacangeo: ok, will change.

fv3core/stencils/fxadv.py
rheacangeo: I mentioned in the PR definition that it was extremely helpful to have the original version commented here to switch already, it saved me a lot of time and effort, so am motivated to keep around this 'single stencil' version for a bit. What is the benefit to making every line start with #? I could see grepping for the code, and thinking it is live code, could be one pitfall of not making it a comment.

fv3core/stencils/fxadv.py
rheacangeo: Oh actually sorry, just remembered -- we do need a computation separating the ra variables from the xfx, as the ra uses xfx at horiontal offsets. But I can reduce this down to 2 computations rather than 4

fv3core/stencils/fxadv.py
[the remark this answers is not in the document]
rheacangeo: I am fine to delete the comment however if it is not helpful and we think this pattern is not going to be super expensive

fv3core/stencils/fxadv.py
eddie-c-davis: I think it is fine as is, it does also have the benefit of modularizing the code.

fv3core/stencils/fxadv.py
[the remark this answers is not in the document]
eddie-c-davis: The comment is fine, I just did not understand the intent. Thank you for clarifying!

fv3core/stencils/fxadv.py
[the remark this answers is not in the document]
eddie-c-davis: This certainly works, I suppose it is just unclear where regions overlap, especially when local and non-local ones are involved. But that discussion is outside the scope of this PR, so no need to change.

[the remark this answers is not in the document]
rheacangeo: You are totally right, that is confusing, for this particular one,  it's also confusing the in the fortran code:
do j=js-1,je+2
       if( j/=1 .and. j/=npy )
where js - 1 = 0 and npy = je + 1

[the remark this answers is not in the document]
rheacangeo: Like, what, why block out a single index not at the end of your range?

rheacangeo:
@jdahm I would be happy to move the main stencil to the bottom. You are right a lot of languages would be upset about referencing functions not 'seen' yet. At the top of the file is nice because you see right away what the file is about. I am also open to either way but we should probably pick a convention... Maybe I should move down for now and we can discuss as a team at some point if there is any preference to move them up.

fv3core/stencils/fxadv.py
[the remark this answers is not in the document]
eddie-c-davis: As long as we match Fortran in confusion :)

fv3core/stencils/fxadv.py
```diff
@@ -1,938 +1,677 @@
...
+        ut: temporary x-velocity transformed from C-grid to D-grid equivalent(?) (inout)
+        vt: temporary y-velocity transformed from C-grid to D-grid equivalent(?) (inout)
+        ra_x: Area increased in the x direction due to flux divergence (inout)
+        ra_y: Area increased in the y direction due to flux divergence (inout)
+        dt: timestep in seconds
+    Grid variable inputs:
+        cosa_u, cosa_v, rsin_u, rsin_v, sin_sg1,sin_sg2, sin_sg3, sin_sg4,
+        rdxa, rdya, area, dy, dx
+    Returns:
+       Updates to xfx_adv, yfx_adv, crx_adv, cry_adv, ut, vt, ra_x, ra_y
```
jdahm, line 307: I like this descriptive docstring! I really like that the grid variables and returns are documented. One nit and then I think we can merge this. It doesn't follow the docstring format https://github.com/VulcanClimateModeling/fv3core/blob/master/CONTRIBUTING.md#docstrings. Napoleon rules dictate that the only sections allowed are those listed here: https://sphinxcontrib-napoleon.readthedocs.io/en/latest/#docstring-sections. Can we reformat this following those rules?

rheacangeo: Ok sure, will do, I just imitated other docstrings I saw in the repo

jdahm (approved):

rheacangeo:
launch jenkins

PR #242, opened 2021-02-15
Replace `k_split_run` call in `UpdateDzD`

eddie-c-davis:
## Purpose

The purpose of this PR is to refactor the `UpdateDzD` stencil to move the `k_split_run` loop explicitly into the code.  It also merges the `edge_profile` stencil into a single stencil call using functions and horizontal regions.

## Code changes:

- Move regions code from `ra_stencil_update` stencil to `ra_func` function.
- Add `edge_profile_top` and `edge_profile_reverse` functions.
- Combine `edge_profile` calls into one `edge_profile_stencil` call using regions.


fv3core/stencils/updatedzd.py
```diff
@@ -17,132 +17,237 @@
...
+# def edge_python(q1, q2, qe1, qe2, dp0, gam, islice, jslice, qe1_2, gam_2):
+#     grid = spec.grid
+#     dcol = dp0[0, 0, :]
+#     km = grid.npz - 1
+#     g0 = dcol[1] / dcol[0]
+#     xt1 = 2.0 * g0 * (g0 + 1.0)
+#     bet = g0 * (g0 + 0.5)
+#     qe1[islice, jslice, 0] = (xt1 * q1[islice, jslice, 0] +
+#                                     q1[islice, jslice, 1]) / bet
+#     qe2[islice, jslice, 0] = (xt1 * q2[islice, jslice, 0] +
```
rheacangeo, line 210: Should we remove this code? in general I am not opposed to keeping it around for a bit, but is there any reason why it would be helpful to bring back?

fv3core/stencils/updatedzd.py
```diff
@@ -151,154 +256,144 @@ def out(zs: sd, zh: sd, ws: sd, dt: float):
...
+):
     grid = spec.grid
-
-    ndif[-1] = ndif[-2]
-    damp_vtd[-1] = damp_vtd[-2]
+    halo = grid.halo
 
     crx_adv = utils.make_storage_from_shape(
-        crx.shape, grid.compute_origin(add=(0, -grid.halo, 0))
+        crx.shape, grid.compute_origin(add=(0, -halo, 0))
```
rheacangeo, line 276: alternatively can use the grid.full_origin and add an offset instead of subtract, but I don't think we've decided on a convention there, so either way is fine

fv3core/stencils/updatedzd.py
```diff
@@ -151,154 +256,144 @@ def out(zs: sd, zh: sd, ws: sd, dt: float):
...
         grid.area,
         xfx_adv,
-        ra_x,
         yfx_adv,
+        ra_x,
         ra_y,
-        origin=grid.compute_origin(add=(-grid.halo, -grid.halo, 0)),
-        domain=(grid.nid, grid.njd, grid.npz + 1),
+        origin=grid.compute_origin(add=(-halo, -halo, 0)),
+        domain=grid.domain_shape_compute(add=(2 * halo, 2 * halo, 1)),
```
rheacangeo, line 314: this one probably should be grid.full_origin() and grid.domain_shape_full() though, i think, because they come up quite often. if you disagree I am open to discuss

fv3core/stencils/updatedzd.py
```diff
@@ -151,154 +256,144 @@ def out(zs: sd, zh: sd, ws: sd, dt: float):
...
     )
 
-    # TODO, when consoldiating k plitting, change this too
-    data = {}
-    for varname in ["zh", "crx_adv", "cry_adv", "xfx_adv", "yfx_adv", "ra_x", "ra_y"]:
-        data[varname] = locals()[varname]
```
rheacangeo, line 218: yay!

fv3core/stencils/updatedzd.py
```diff
@@ -151,154 +256,144 @@ def out(zs: sd, zh: sd, ws: sd, dt: float):
...
+    ndif[-1] = ndif[-2]
+    damp_vtd[-1] = damp_vtd[-2]
```
rheacangeo, line 318: so these come in as columns now instead of scalars? excellent

fv3core/stencils/updatedzd.py
```diff
@@ -151,154 +256,144 @@ def out(zs: sd, zh: sd, ws: sd, dt: float):
...
+    kstarts = utils.get_kstarts({"ndif": ndif, "damp": damp_vtd}, grid.npz + 1)
 
-    col = {"ndif": ndif, "damp": damp_vtd}
+    for ki, nk in kstarts:
```
rheacangeo, line 321: ah! This is a good step forward. For future refactors getting rid of k_split_run, I think we might just want to hard code the first 2-3 'intervals' inside the stencils, but this is a good first step that keeps it more general, and we should probably discuss before hard coding in those levels. Also probably can't do that right now as this is calling other stencil methods

fv3core/stencils/updatedzd.py
```diff
@@ -151,154 +256,144 @@ def out(zs: sd, zh: sd, ws: sd, dt: float):
...
+        spec.namelist.hord_tm,
+        xfx_adv,
+        yfx_adv,
+        ra_x,
+        ra_y,
+        fx,
+        fy,
+        kstart=kstart,
+        nk=nk,
+    )
```
rheacangeo, line 384: how does this work without restricting the k domain to each subset?

rheacangeo (commented):
Awesome!! I don't totally get yet how this works since this function calls other ones that don't have the k_split_run removed from them? How does this validate? Will look more

rheacangeo (approved):
Ok I think I remember now -- last year I had done a start to a transition of all k_split_run to use kstart and nk in all the calls rather than forcing them to make sliced copies, but I think I didn't finish that route throughout the d_sw code. There were a few things that didn't work still, so k_split_run was still slicing. But for the modules called from update_dz_d, they do have that working, so you are able to simply call the sub functions on different k ranges.

fv3core/stencils/updatedzd.py
eddie-c-davis: I would be happy to remove it.

fv3core/stencils/updatedzd.py
eddie-c-davis: Yes, and they should be stencilisable in the lower-grid-dim-vars PR.

eddie-c-davis:
launch jenkins

rheacangeo:
launch jenkins

PR #247, opened 2021-02-18
Refactor/updatedzc

rheacangeo:
## Purpose

Consolidate updatedzc to a single stencil to reduce the number of stencil calls

## Code changes:

- updatedzc no longer calls python methods and has a single stencil
- removed unnecessary data copying methods from utils/grid.py
- updatedzc test directly calls the stencil
- dyn_core directly calls this stencil
- The variable 'zs' is changed to 'surface' everywhere it shows up in the model

## Checklist


fv3core/stencils/updatedzc.py
```diff
@@ -220,44 +78,29 @@ def update_dz_c(
...
+            )
     with computation(PARALLEL), interval(-1, None):
-        rdt = 1.0 / dt
-        ws3 = (zs - gz) * rdt
+        rdt = 1.0 / dt2
+        # TODO: region for local validation only
+        with horizontal(
+            region[local_is - 1 : local_ie + 2, local_js - 1 : local_je + 2]
+        ):
+            ws = (z_surface - gz) * rdt
```
mcgibbon, line 99: `rdt` only gets used in this one place below, so you could condense this for readability (pending possibly updating the `dt2` name).

```suggestion
        # TODO: region for local validation only
        with horizontal(
            region[local_is - 1 : local_ie + 2, local_js - 1 : local_je + 2]
        ):
            ws = (z_surface - gz) / dt2
```

fv3core/stencils/updatedzc.py
```diff
@@ -220,44 +78,29 @@ def update_dz_c(
             xfx = p_weighted_average_bottom(ut, dp_ref)
             yfx = p_weighted_average_bottom(vt, dp_ref)
     with computation(PARALLEL), interval(...):
+        gz_x = gz
+        gz_x = corners.fill_corners_2cells_mult_x(gz_x, gz_x, 1.0, 1.0, 1.0, 1.0)
+        gz_y = gz_x
+        gz_y = corners.fill_corners_2cells_mult_y(gz_y, gz_y, 1.0, 1.0, 1.0, 1.0)
```
mcgibbon, line 84: Suggestion: `gz_x`, `gz_y`, `xfx`, `yfx` are all temporaries used only in the context of this function. Are there any more explicit names you can think of if you have context for what these mean? I'm sure you want to keep them short since the `gz` expression below is complex, but a few more characters might give a lot more meaning.

fv3core/stencils/updatedzc.py
```diff
@@ -1,214 +1,72 @@
...
+    gz: FloatField,
+    ws: FloatField,
+    dt2: float,
 ):
+    """Update the model heights from the C-grid wind flux
+
+    After the model runs c_sw and advances the c-grid variables half a timestep,
+    the grid deforms with the flow in the Lagrangian coordinate system. This
+    module updates the model heights based on the flux, and the rate of change of
+    the lowest model height compared to the fixed surface height.
```
mcgibbon, line 54: Do you know from what rule/assumptions it's able to update the heights? It's not hydrostatic, so I am having a hard time imagining what constrains the height change.

If you know, it would be great to add here. If not, I think Lucas would be happy to receive a question about this.

fv3core/stencils/updatedzc.py
```diff
@@ -1,214 +1,72 @@
...
+
+    Args:
+         dp_ref: vertical delta in column reference pressure(in)
+         z_surface: surface height (m) (in)
+         ut: x-velocity on the C-grid, contravariantof the D-grid winds(in)
+         vt: y-velocity on the C-grid, contravariantof the D-grid winds(in)
+         gz: height of the model grid cells (m) (inout)
+         ws: difference in height between lowest model layer and the surface per C-grid
+             timestep(inout)
+         dt2: half a model timestep (for C-grid update) in seconds (in)
```
mcgibbon, line 64: I notice `dt -> dt2`, I'm a little concerned with this name being ambiguous/providing more context than this function needs. Out of context, in "half a model timestep for C-grid update", "model timestep for C-grid update" means half a dynamics timestep, so the phrase means 1/4th of the model timestep (because the model gets progressed forward half a timestep on each grid type).

Maybe that's actually what it means (1/4th of the dynamics timestep), I actually don't know.

When I say "providing more context than the function means", I mean the function is saying where you're calling it from (somewhere that has a timestep twice as long as this function does), but you're allowed to call it from a different context. Once we've learned/clarified what this function is doing, it could be called by a different model in someone's analysis script with a timestep that has nothing to do with FV3.

Would it make sense to provide this as just `dt` and say it's the timestep over which the dz update is taking place? dt gets used strictly to calculate vertical velocity, so this would be the most precise. You could add a clarification sentence along the lines of "Note within the FV3 dynamical core this is half a dynamics time step".

fv3core/stencils/updatedzc.py
mcgibbon, line 99: What's with the units on this? is z_surface actually gz_surface?

fv3core/stencils/updatedzc.py
```diff
@@ -220,44 +78,29 @@ def update_dz_c(
...
-    )
-    grid.overwrite_edges(gz, gz_in, 2, 2)
-    grid.overwrite_edges(ws3, ws, 2, 2)
-    return gz, ws3
+        # TODO region for local validation only
+        with horizontal(
+            region[local_is - 1 : local_ie + 2, local_js - 1 : local_je + 2]
+        ):
+            gz_kp1 = gz[0, 0, 1] + DZ_MIN
+            gz = gz if gz > gz_kp1 else gz_kp1
```
mcgibbon, line 106: Another local you could rename (`gz_kp1`), if you have a bit more context to add. Is this `gz_min` for the given height?

fv3core/stencils/updatedzc.py
```diff
@@ -1,214 +1,72 @@
...
+    the lowest model height compared to the fixed surface height.
+
+    Args:
+         dp_ref: vertical delta in column reference pressure(in)
+         z_surface: surface height (m) (in)
+         ut: x-velocity on the C-grid, contravariantof the D-grid winds(in)
+         vt: y-velocity on the C-grid, contravariantof the D-grid winds(in)
+         gz: height of the model grid cells (m) (inout)
+         ws: difference in height between lowest model layer and the surface per C-grid
+             timestep(inout)
```
mcgibbon, line 63: Really curious what `ws` is conceptually and why it is calculated. `(gz - gz_surface) / timestep`? Why would that be used? I see on the next line in dyn_core, we throw away everything but the last value in the vertical. It gets passed back to dyn_core where it's called `ws3`, then `riem_solver_c` where it's called `ws` again, then `sim1_solver` where it's called `wsr`, until it's fiiiinally used in one line as `wsr_top`:

```
w = (
    dm * w1 + dt * (pp[0, 0, 1] - pp) - p1 * wsr_top - aa * w[0, 0, -1]
) / bet
```

Crazy thing is, within that function we have everything available to compute ws - we have gz and rdt, and `ws = sum_z(gz/dt)`. Maybe we could move it there in a later refactor? It's such a strange temporary with no real meaning to be moving between all these high-level routines.

tests/translate/translate_riem_solver3.py
```diff
@@ -8,7 +8,7 @@ def __init__(self, grid):
         self.compute_func = riem_solver3.compute
         self.in_vars["data_vars"] = {
             "cappa": {},
-            "zs": {},
+            "z_surface": {"serialname": "zs"},
```
mcgibbon, line 11: ty for going throughout and updating this!

fv3core/stencils/updatedzc.py
```diff
@@ -220,44 +78,29 @@ def update_dz_c(
...
+            region[local_is - 1 : local_ie + 2, local_js - 1 : local_je + 2]
+        ):
+            gz = (gz_y * area + fx - fx[1, 0, 0] + fy - fy[0, 1, 0]) / (
+                area + xfx - xfx[1, 0, 0] + yfx - yfx[0, 1, 0]
+            )
     with computation(PARALLEL), interval(-1, None):
-        rdt = 1.0 / dt
-        ws3 = (zs - gz) * rdt
+        rdt = 1.0 / dt2
+        # TODO: region for local validation only
```
mcgibbon, line 95: Hopefully we can get a solution soon for specifying smaller validation subdomains in our translate classes, so we don't need to overcompute data that isn't used and can clean up the regions like this which result.

mcgibbon (commented):
It's exciting for me to look at more of the dycore and get thinking about how all these parts fit together, thanks for having me review!

Positive changes all throughout, so thanks for the PR so far! The Fortran model may have lied to us - is z_surface actually gz_surface? Otherwise, how does `z_surface - gz` make sense, units-wise?

Leaving as a comment review not because there are necessarily many changes needed, but because I posted a few open-ended suggestions and would like to re-review if you decide to go with some of them.

fv3core/stencils/updatedzc.py
rheacangeo: good idea

fv3core/stencils/updatedzc.py
rheacangeo: Good point, these gz_variables were temporaries to carry the corners filled in the x and y directions, but we actually don't need both, we just need 1 gz_tmp because of the functional nature of gtscript functions. xfx and yfx are kind of odd names for cell averaged averages of u and v. I've renamed and reorganized.

fv3core/stencils/updatedzc.py
mcgibbon: Yes, that's definitely true/accurate, but it's not the full story. I couldn't from that information write a formula (or have any idea how to start) that would take the winds and determine how much vertical expansion/contraction there is.

If this were an incompressible model, I could do it by using conservation of volume - based on how much horizontal divergence (convergence) there is, I could say how much contraction (expansion) of the column there is in the vertical. With a hydrostatic assumption, I could make a similar statement.

For this model, I have no idea what the assumption is that gives closure. As far as I know, all the horizontal divergence/convergence could go to increasing delp/density while keeping dz entirely constant.

Do you have any idea what the closure assumption is here?

fv3core/stencils/updatedzc.py
```diff
@@ -1,263 +1,98 @@
...
         with interval(1, -1):
-            xfx = p_weighted_average_domain(ut, dp_ref)
-            yfx = p_weighted_average_domain(vt, dp_ref)
+            u_average = p_weighted_average_domain(ut, dp_ref)
+            v_average = p_weighted_average_domain(vt, dp_ref)
         with interval(-1, None):
-            xfx = p_weighted_average_bottom(ut, dp_ref)
-            yfx = p_weighted_average_bottom(vt, dp_ref)
+            u_average = p_weighted_average_bottom(ut, dp_ref)
+            v_average = p_weighted_average_bottom(vt, dp_ref)
```
mcgibbon, line 72: Woooah I actually thought these were pressure for some reason, thanks for updating the names!

fv3core/stencils/updatedzc.py
mcgibbon: I'm good with whichever way we go on this, so long as we can eventually remove these computations. It's definitely true that once we have more stencils merged, the code should be figuring out automatically what is and isn't needed to be computed, and we should be able to remove these more confidently.

fv3core/stencils/updatedzc.py
mcgibbon: if possible I'd suggest `surface_delta_gz = gz[lowest level] - gz_surface` with a docstring/comment (I got confused and thought it was `gz[highest level] - gz_surface`, maybe because of the name `wsr_top`). `sim1_solver` already has `rdt` available, so you'd replace `wsr_top` with `surface_delta_gz * rdt`.

`surface_delta_gz_over_dt` is also fine, just a bit of a mouthful, and more complex/confusing than `surface_delta_gz` already is.

As you say we should visit this later and leave the structure for now, but some more thoughts:

> conceptually sort of makes sense it is happening here with the update to gz

I agree it makes a bit of sense, but it's also a little dangerous. `surface_delta_gz` is a diagnostic value, determined by gz (prognostic) and by surface_gz (a constant). If you were to call updatedzc, and then someone added another modification to gz before calling riem_solver, it would introduce a bug because surface_delta_gz would be inaccurate (it would only account for the update from updatedzc and not from the new modification).

Huh, I wonder if the purpose of this is to determine the vertical velocity you need in order to move the lowest level gz to the location of the surface gz in a single timestep, and gets used as the lowest level velocity (in geopotential units) later. That would maybe make sense. If we figure this out we could later name it something like `lowest_level_geopotential_velocity`. For now when we're unsure, something technical and explicit like `surface_delta_gz` is more practical.

rheacangeo: Given it is called 'ws' I tend to agree, it is likely supposed to be some sort of vertical velocity at the surface, but also agree going with surface_delta_gz for now is good. I think if we move this calculation to the riemann solver, we should do it later, as Eddie has a big lower dimensional storage refactor, so I'd rather make that change after the dust from that settles. Also note there is a parallel with a 'wsd' variables computed in updatedzd and using in riem_solver3 :).

rheacangeo:
Currently works with the gtcuda and numpy backend, but does not validate fv_dynamics with the gtx86 backend

mcgibbon:
launch jenkins

rheacangeo:
Closing because out of date and regions working better with gt backends unlikely to change anytime soon

PR #248, opened 2021-02-19
Refactor/pgradc

rheacangeo:
## Purpose

Update the stencil for p_grad_c to our current conventions, and reduce it from 2 stencils to 1. 

## Code changes:

- made a single p_grad_c_stencil stencil
- deleted the file fv3core/stencils/pgradc.py and instead moved the stencil the the dyn_core
- updated dyn_core stencils to use FloatField rather than sd
- the PGradC test calls the stencil directly

## Checklist


jdahm:
If pgradc.py is moved into dyn_core.py, should the translate class also be moved into the translate class for that file?

rheacangeo:
Sure! Good idea.

rheacangeo:
launch jenkins

rheacangeo:
launch jenkins

fv3core/stencils/dyn_core.py
```diff
@@ -56,12 +71,67 @@ def set_pem(delp: sd, pem: sd, ptop: float):
...
+         gz:  height of the model grid cells (m)(in)
+         dt2: half a model timestep (for C-grid update) in seconds (in)
+    Grid variable inputs:
+        rdxc, rdyc
+    """
+    from __externals__ import local_ie, local_is, local_je, local_js, namelist
+
+    with computation(PARALLEL), interval(...):
+        if __INLINED(namelist.hydrostatic):
+            wk = pkc[0, 0, 1] - pkc
```
oelbert, line 118: Using a [0,0,1] offset inside a full interval here works because the domain is constricted, right? Does this impact the stencil fusion process?

I guess I'm asking if it's preferable to restrict the domain and use `interval(...)` or to use the full domain and have `interval(0,-1)`? The former seems better from a data handling perspective, but the latter seems easier for the backend to fuse. Am I just overthinking this?

oelbert (approved):
I don't think my question actually blocks this, but I am curious. Also, good job reducing the amount of code!

fv3core/stencils/dyn_core.py
rheacangeo: Good question! I think we don't know yet whether this choice have an impact, though definitely seems worth considering. Our arrays have an extra buffer point in every directions because sometimes they are used, but most of the time 'full_domain' does not include that top buffer point. I think if we have the answer to your question we should do a refactor pass through the dycore as both patterns are abundant.  @eddie-c-davis what do you think?

gmao-ckung (approved):
Everything looks good from my perspective!

fv3core/stencils/dyn_core.py
eddie-c-davis: I would say it depends on the surrounding intervals right now. The current fusion heuristic requires the intervals to be the same in order to fuse, and improved fusion probably wins over storage reduction in that case. Would be an interesting experiment though!

PR #254, opened 2021-02-23
Feature/add sequential computation

twicki:
## Purpose

This PR ensures that compilation (not execution) happens sequentially. This change is motivated by nodes running out of memory when compiling large problems in parallel as the memory consumption is both from compilation as well as from runtime. 

## Code changes:

- Added utility to get size and rank (even if MPI is not present).
- Change to the decorator to send / receive messages to ensure sequential compilation

## Checklist

### DIFF
--- a/fv3core/decorators.py
+++ b/fv3core/decorators.py
@@ -250,8 +250,8 @@ class FV3StencilObject:
 
             # we ensure sequential compilation to reduce memory pressure
             if mpi.is_parallel_context() and (
-                global_config.get_backend() is "gtx86"
-                or global_config.get_backend() is "gtcuda"
+                global_config.get_backend() == "gtx86"
+                or global_config.get_backend() == "gtcuda"
             ):
                 rank = mpi.get_rank()
                 root = 0
@@ -277,8 +277,8 @@ class FV3StencilObject:
                 )
 
             if mpi.is_parallel_context() and (
-                global_config.get_backend() is "gtx86"
-                or global_config.get_backend() is "gtcuda"
+                global_config.get_backend() == "gtx86"
+                or global_config.get_backend() == "gtcuda"
             ):
                 if rank == root:
                     mpi.send(rank + 1)
--- a/fv3core/decorators.py
+++ b/fv3core/decorators.py
@@ -254,7 +254,8 @@ class FV3StencilObject:
                 if rank > root:
                     utils.recv(rank - 1)
 
-            # gtscript.stencil always returns a new class instance even if it used the cached module.
+            # gtscript.stencil always returns a new class instance even
+            # if it used the cached module.
             self.stencil_object = gtscript.stencil(
                 definition=self.func, **stencil_kwargs
             )
--- a/fv3core/decorators.py
+++ b/fv3core/decorators.py
@@ -248,7 +248,6 @@ class FV3StencilObject:
             }
 
             # Claim this stencil for our rank...
-            func_id = self.func.__module__ + "." + self.func.__qualname__
             if utils.get_size() > 1:
                 rank = utils.get_rank()
                 root = 0
--- a/fv3core/decorators.py
+++ b/fv3core/decorators.py
@@ -249,6 +249,15 @@ class FV3StencilObject:
 
             # gtscript.stencil always returns a new class instance even if it
             # used the cached module.
+
+            # Claim this stencil for our rank...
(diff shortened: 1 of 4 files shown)


fv3core/utils/gt4py_utils.py
```diff
@@ -52,8 +52,61 @@
...
 logger = logging.getLogger("fv3ser")
 
+
+def get_size() -> int:
+    """Get the size of the MPI COMM_WORLD communicator if it is defined.
+
+    Returns:
+        int: size
+    """
+    return MPI.COMM_WORLD.Get_size() if MPI else 0
```
jdahm, line 62: I understand why this was put here (so that it can be used in `decorators.py`), but this seems like the wrong place for MPI utilities...

twicki: what is your suggestion?

jdahm: If they're used within decorators, can they just be defined there? Also, probably want them to begin with a leading `_`.

twicki:
launch jenkins

fv3core/decorators.py
```diff
@@ -247,8 +248,18 @@ def __call__(self, *args, origin: Index3D, domain: Index3D, **kwargs) -> None:
...
-            # used the cached module.
+            # we ensure sequential compilation to reduce memory pressure
+            if mpi.is_parallel_context() and (
+                global_config.get_backend() == "gtx86"
+                or global_config.get_backend() == "gtcuda"
+            ):
+                rank = mpi.get_rank()
+                root = 0
+                if rank > root:
+                    mpi.recv(rank - 1)
```
rheacangeo, line 259: do we need a 'wait' at all? maybe not necessary

fv3core/utils/mpi.py
```diff
@@ -2,3 +2,66 @@
...
+
+
+def get_rank() -> int:
+    """Get the current rank in the MPI COMM_WORLD communicator if it is defined.
+
+    Returns:
+        int: rank
+    """
+    return MPI.COMM_WORLD.Get_rank() if is_parallel_context() else -1
+
```
rheacangeo, line 34: We could consider using some  of the functionality of the submodule fvgfs-util, and accessing through the 'comm' object, but it may not be available in the scope we need it currently (I think that was some of the idea Johann was proposing). So that might be a subsequent step, but I am ok with putting it here for now.

rheacangeo (approved):

twicki:
seems to no longer be required

fv3core/decorators.py
```diff
@@ -247,8 +248,18 @@ def __call__(self, *args, origin: Index3D, domain: Index3D, **kwargs) -> None:
                 **self.backend_kwargs,
             }
 
-            # gtscript.stencil always returns a new class instance even if it
-            # used the cached module.
+            # we ensure sequential compilation to reduce memory pressure
```
ofuhrer, line 251: Just a side note. This is a rather unintuitive way to implement ordering using MPI send and recv. The "standard" way of doing this is by doing...
```
for i = 1, number_of_ranks
    if (my_ranks == i) {
      # do work
    }
    MPI_Barrier()
```

fv3core/utils/mpi.py
```diff
@@ -2,3 +2,66 @@
     from mpi4py import MPI
 except ImportError:
     MPI = None
+from typing import Any
+
+
+def is_parallel_context() -> bool:
+    """Checks if we are in a parallel context, meaing the size of the
```
ofuhrer, line 9: meaning

fv3core/utils/mpi.py
```diff
@@ -2,3 +2,66 @@
     from mpi4py import MPI
 except ImportError:
     MPI = None
+from typing import Any
```
ofuhrer, line 5: Awesome idea to build a wrapper for MPI.

fv3core/utils/mpi.py
```diff
@@ -2,3 +2,66 @@
...
+def bcast(data) -> None:
+    """A wrapper for a broadcast in the MPI COMM_WORLD communicator from rank 0.
+
+    Args:
+        data: The data to broadcast
+    """
+    return MPI.COMM_WORLD.bcast(data, root=0)
+
+
+def send(dest: int, tag: int = 0) -> None:
```
ofuhrer, line 45: I would actually hope we never really use send/recv.

PR #256, opened 2021-02-23
Merging stencils in d_sw.py

gmao-ckung:
## Purpose

Several stencils within `fv3core/stencils/d_sw.py` are consolidated into a single stencil.

## Code changes:

The following change was made in `fv3core/stencils/corners.py`

- Code from `fix_corner_ke` was removed from `fv3core/stencils/corners.py` and moved to `fv3core/stencils/d_sw.py`

The following changes were made in `fv3core/stencils/d_sw.py`

- `initial_heat_source` stencil was removed
- Single stencil created to merge `ke_from_b_wind`, `fix_corner_ke`, `horizontal_relative_vorticity_from_winds`, and `adjust_w_and_qcon`
- Single stencil created to merge `ub_from_vort`, `vb_from_vort`, `zrat_vorticity`, and `basic_addition`
- Single stencil created to merge `u_from_ke` and `v_from_ke`
- Single stencil created to merge `heat_from_damping`, `basic_add_term_stencil`, and `basic_subtract_term_stencil`
- `FloatField` replaces `utils.sd`
- `VorticityVolumeMean` test case was removed


fv3core/stencils/d_sw.py
```diff
@@ -93,8 +94,92 @@ def horizontal_relative_vorticity_from_winds(
...
+                ke = corners.corner_ke(ke, u, v, ut, vt, dt, 0, 0, -1, 1)
+        with horizontal(region[i_end + 1, j_start]):
+            if not nested:
+                ke = corners.corner_ke(ke, u, v, ut, vt, dt, -1, 0, 0, -1)
+        with horizontal(region[i_end + 1, j_end + 1]):
+            if not nested:
+                ke = corners.corner_ke(ke, u, v, ut, vt, dt, -1, -1, 0, 1)
+        with horizontal(region[i_start, j_end + 1]):
+            if not nested:
+                ke = corners.corner_ke(ke, u, v, ut, vt, dt, 0, -1, -1, -1)
```
rheacangeo, line 163: Awesome! So you probably weren't aware -- but 'with horizontal' can happen inside of functions, which is really handy. So you could put all of these regions into a single gtscript function named something like corners.all_corners_ke (and still reuse your corner_ke function from within it!).   You will need to restate 'from__externals__ ...'

Also, it's great you kept the nested logic, but I think we can not include it as long as we document it somewhere. c_sw there is a comment " # additional assumption: not __INLINED(spec.grid.nested)" that we probably want to add to the ke corners function. Then you don't have to pass in nested at all and don't need as many if statements.

fv3core/stencils/d_sw.py
```diff
@@ -93,8 +94,92 @@ def horizontal_relative_vorticity_from_winds(
...
+
+        vt = u * dx
+        ut = v * dy
+        vorticity[0, 0, 0] = rarea * (vt - vt[0, 1, 0] - ut + ut[1, 0, 0])
```
rheacangeo, line 167: This is from the stencil horizontal_relative_vorticity_from_winds. You should probably delete that stencil as it is no longer used, except that it is used by the VorticityVolumeMean test. SInce it has a docstring it would be nice to preserve that work. You could make that a gtscript function, include the docstring, and call it from here. For the test, you can either adapt it to create a stencil in the test class (such as for d2a2c here https://github.com/VulcanClimateModeling/fv3core/commit/95456e9d15190519356adaecbbc559bdbbbf4d78#diff-8e0363bcb510536bac792406fdf19fb833aa3782aaa0d026afac68f87e456ec9), or possibly just delete the test since this is really a tiny amount of code.

fv3core/stencils/d_sw.py
```diff
@@ -93,8 +94,92 @@ def horizontal_relative_vorticity_from_winds(
...
+
+        with horizontal(region[local_is : local_ie + 1, local_js : local_je + 1]):
```
rheacangeo, line 169: 'the with horizontal can be put inside of the function (you will need to restate 'from__externals__ ...')

fv3core/stencils/d_sw.py
```diff
@@ -93,8 +94,92 @@ def horizontal_relative_vorticity_from_winds(
...
+            w, q_con = adjust_w_and_qcon(w, delp, dw, damp_w, q_con)
+
+
+@gtstencil()
+def not_inlineq_pressure(
```
rheacangeo, line 174: can this on could potentially be combined with vbke, but the regions would be awkward, and it wasn't in the list. curiious what you think? I am fine keeping it separate

fv3core/stencils/d_sw.py
```diff
@@ -105,64 +190,83 @@ def not_inlineq_pressure(gx: sd, gy: sd, rarea: sd, fx: sd, fy: sd, pt: sd, delp
...
+    u: FloatField,
+    v: FloatField,
+):
+    from __externals__ import local_ie, local_is, local_je, local_js
+
     with computation(PARALLEL), interval(...):
-        v[0, 0, 0] = ut + ke - ke[0, 1, 0] - fx
+        with horizontal(region[local_is : local_ie + 1, local_js : local_je + 2]):
+            u = u_from_ke(ke, vt, fy)
+        with horizontal(region[local_is : local_ie + 2, local_js : local_je + 1]):
```
rheacangeo, line 254: I think these regions would not be necessary if combined with the heat_from_damping stencil, which will only be possible when we can call delnflux from a stencil. If you agree, it would be cool to add a note saying as much, something like #TODO: may be able to remove local regions once heat from_damping is in the same stencil

fv3core/stencils/d_sw.py
```diff
@@ -190,92 +333,41 @@ def heat_source_from_vorticity_damping(
...
-        heat_source (out): heat source from vorticity damping
-            implied by energy conservation
-        diss_est (out): dissipation estimate, only calculated if
-            calculate_dissipation_estimate is 1
-        kinetic_energy_fraction_to_damp (in): according to its comment in fv_arrays,
-            the fraction of kinetic energy to explicitly damp and convert into heat.
-            TODO: confirm this description is accurate, why is it multiplied
-            by 0.25 below?
-        calculate_dissipation_estimate (in): If 1, calculate dissipation estimate.
-            Equivalent in Fortran model is do_skeb
```
rheacangeo, line 219: would be nice to not lose the docstring for variables still relevant in heat_source_from_vorticity_damping_fxn

rheacangeo (commented):
Looking great! I have some suggestions of comments to add, dead code to remove, and docstrings to preserve, but this looks very close to being mergeable!

fv3core/stencils/d_sw.py
gmao-ckung: Done!  I created a function `all_corners_ke` that encompasses all the calls to `corners.corner_ke` using `with_horizontal`...

fv3core/stencils/d_sw.py
gmao-ckung: I'm not sure how this is the case, but the VorticityVolumeMean test seems to be run every time I run the d_sw test case, and the test will complain that it can't find the `horizontal_relative_vorticity_from_winds` stencil if its removed.

fv3core/stencils/d_sw.py
gmao-ckung: Done!

fv3core/stencils/d_sw.py
gmao-ckung: I think it can be combined without too many issues with vbke since the origin is the same and the domain is almost the same.  I created a stencil `not_inlineq_pressure_and_vbke` that combines the two together.

fv3core/stencils/d_sw.py
gmao-ckung: Yep I agree with what you said!

fv3core/stencils/d_sw.py
gmao-ckung: Added docstring to `heat_source_from_vorticity _damping_fxn`

fv3core/stencils/d_sw.py
rheacangeo: Are you specifying --which_modules=D_SW when running the tests? Either way the way to delete a test is to remove the translate class file and remove reference to it in tests/translate/__init__.py. Or you can make a stencil inside of the VorticityVolumeMean test class that alls your function

fv3core/stencils/d_sw.py
```diff
@@ -93,76 +94,225 @@ def horizontal_relative_vorticity_from_winds(
...
+        local_ie,
+        local_is,
+        local_je,
+        local_js,
+    )
+
     with computation(PARALLEL), interval(...):
-        vb[0, 0, 0] = vort - vort[0, 1, 0]
+        if __INLINED(spec.namelist.inline_q == 0):
+            with horizontal(region[local_is : local_ie + 1, local_js : local_je + 1]):
```
rheacangeo, line 220: Nice work! Would you mind adding a "TODO: only needed for d_sw validation"
 -- it looks like not having this, DynCore would still work, and it will help us in a subsequent pass to know if some regions can be removed simply by restricting the validation region.

fv3core/stencils/d_sw.py
```diff
@@ -218,64 +367,78 @@ def heat_source_from_vorticity_damping(
...
+                rdy,
+                heat_source,
+                dissipation_estimate,
+                kinetic_energy_fraction_to_damp,
+                calculate_dissipation_estimate,
+            )
+
+        with horizontal(region[local_is : local_ie + 1, local_js : local_je + 2]):
+            if damp_vt > 1e-5:
+                # basic.add_term_stencil()
```
rheacangeo, line 436: can probably remove this comment

fv3core/stencils/d_sw.py
```diff
@@ -218,64 +367,78 @@ def heat_source_from_vorticity_damping(
...
+                u = u + vt
+        with horizontal(region[local_is : local_ie + 2, local_js : local_je + 1]):
+            if damp_vt > 1e-5:
+                # basic.subtract_term_stencil()
```
rheacangeo, line 440: can probably also remove this one

fv3core/stencils/d_sw.py
```diff
@@ -475,7 +638,16 @@ def damp_vertical_wind(w, heat_s, diss_e, dt, column_namelist):
 
 
 @gtstencil()
-def ubke(uc: sd, vc: sd, cosa: sd, rsina: sd, ut: sd, ub: sd, dt4: float, dt5: float):
+def ubke(
```
rheacangeo, line 641: One last possible combination -- ubke looks like it could be combined with the step before it, computing ke from ub and vb, they are even in the same domain.

fv3core/stencils/d_sw.py
```diff
@@ -93,76 +94,225 @@ def horizontal_relative_vorticity_from_winds(
...
-    with computation(PARALLEL), interval(...):
-        pt = flux_integral(
-            pt, delp, gx, gy, rarea
-        )  # TODO: Put [0, 0, 0] on left when gt4py bug is fixed
-        delp = delp + flux_component(
-            fx, fy, rarea
-        )  # TODO: Put [0, 0, 0] on left when gt4py bug is fixed
-        pt[0, 0, 0] = pt / delp
+@gtscript.function
+def horizontal_relative_vorticity_from_winds_f(u, v, ut, vt, dx, dy, rarea, vorticity):
```
rheacangeo, line 98: Another option is to just have the stencil of the same name call this function rather than having duplicated code. But, like mentioned, this could be pushed into the VorticityVolumeMean test, or that tiny test can be removed.

rheacangeo (commented):
Looking very close! Just a couple of tiny comments. Also would be great to merge in master if you haven't already so that the CI tests can run on your next push.

fv3core/stencils/d_sw.py
gmao-ckung: I was specifying `--which_moudles=D_SW` during my tests, but the VoticityVolumeMean test was still running.  I ended up deleted the particular test and removing the reference in `tests/translate/init.py`.

fv3core/stencils/d_sw.py
gmao-ckung: Done!

fv3core/stencils/d_sw.py
gmao-ckung: Done!

fv3core/stencils/d_sw.py
gmao-ckung: Done!

fv3core/stencils/d_sw.py
gmao-ckung: With the VorticityVolumeMean test removed, I got rid of the `_f` in the name of this function.

fv3core/stencils/d_sw.py
gmao-ckung: Done!

rheacangeo:
Great! Looks like tests are failing because some of the translate classes and that test individual components of d_sw need to be updated to use the changed signatures. For stencils that have been combined with other stencils, you can still test the smaller units by separating the ideas that are merged into gtscript functions, then creating a stencil in the test code that calls that function. While it was fine to delete VorticityVolumeMean because it was so tiny, it is worthwhile to still test some sub-units of d_sw, because that one has relaxed thresholds and can let real errors slip through. Happy to chat about it if this doesn't make sense.

rheacangeo:
This is not validating on the gtcuda backend and appears to have gotten wedged/stuck doing the gtx86 backend. It may be this is on hold until the regions feature is update. I'll launch it again just to be sure.

rheacangeo:
launch jenkins

rheacangeo:
launch jenkins

rheacangeo:
launch jenkins

rheacangeo:
launch jenkins

rheacangeo:
launch jenkins

PR #257, opened 2021-02-23
enable gtcuda backend

twicki:
## Purpose

This PR changes the structure in which the pre-compiled gt_caches are stored to also distinguish the backends in a more obvious way to copy less data around and therefore enables runs with the cuda backend

## Infrastructure changes:

-  The plan fv3core-regression_PR now has an additional label in the backend column being gtcuda.
-  The plan zz_fv3core_cacheSetup now has an additional label in the backend column being gtcuda.


twicki:
launch jenkins

rheacangeo (approved):
this works! https://jenkins.ginko.ch/job/fv3core-regression_PR/307/

PR #240, opened 2021-02-12
Regions/neg adj3

oelbert:
## Purpose

Refactoring neg_adj3 to improve performance by removing Python code between stencil calls and by adding 2D fields 

## Code changes:

- `fillq` now uses 2d fields to accumulate sums
- Renamed internal stencil variables for clarity and consistency with the rest of the code
- Pulled Python code into stencils to allow full fusion
- Added a docstring to neg_adj3.compute()

PR #260, opened 2021-02-23
Regions/neg adj3 (#240)

oelbert:
## Purpose

Re-implementing the refactor for neg_adj3:

Refactoring neg_adj3 to improve performance by removing Python code between stencil calls and by adding 2D fields

## Code changes:
- fillq now uses 2d fields to accumulate sums
- Renamed internal stencil variables for clarity and consistency with the rest of the code
- Pulled Python code into stencils to allow full fusion
- Added a docstring to neg_adj3.compute()

### DIFF
--- a/fv3core/stencils/neg_adj3.py
+++ b/fv3core/stencils/neg_adj3.py
@@ -1,184 +1,157 @@
 import gt4py.gtscript as gtscript
-import numpy as np
 from gt4py.gtscript import BACKWARD, FORWARD, PARALLEL, computation, interval
 
 import fv3core._config as spec
 import fv3core.utils.global_constants as constants
 import fv3core.utils.gt4py_utils as utils
 from fv3core.decorators import gtstencil
+from fv3core.utils.typing import FloatField, FloatFieldIJ
 
 
-sd = utils.sd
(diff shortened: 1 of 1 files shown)


fv3core/stencils/neg_adj3.py
```diff
@@ -1,184 +1,157 @@
...
-def fillq(q:sd, dp:sd):
+def fillq(q: FloatField, dp: FloatField, sum1: FloatFieldIJ, sum2: FloatFieldIJ):
+    with computation(FORWARD), interval(...):
+        # reset accumulating fields
+        sum1 = 0.0
+        sum2 = 0.0
     with computation(FORWARD), interval(...):
         if q > 0:
-            sum1 = sum1[0, 0, -1] + q * dp
+            sum1 = sum1 + q * dp
```
jdahm, line 103: This looks a bit suspect. Can you double check that it was intended?

jdahm (approved):
Looks good otherwise!

fv3core/stencils/neg_adj3.py
oelbert: Sum1 and Sum2 are 2D fields now, so yes this was intended.

PR #265, opened 2021-02-25
Attempt to improve performance history plot

ofuhrer:
## Purpose

The Fortran timings were not visible on the performance plot. Since the numpy times will probably not go down significantly, this is an attempt to make everything visible. This includes abbreviating the legend to have more plot area.

## Code changes:

Just a change in `plotting.py`

## Plot changes:

image

image

### DIFF
--- a/examples/standalone/benchmarks/plotting.py
+++ b/examples/standalone/benchmarks/plotting.py
@@ -27,7 +27,7 @@ if __name__ == "__main__":
     plot_variance = True
     plots = {
         "per_timestep": {
-            "title": "Performance history of components",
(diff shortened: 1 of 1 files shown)


examples/standalone/benchmarks/plotting.py
```diff
@@ -76,6 +76,11 @@
             specific = [x for x in alldata if x["setup"]["version"] == backend]
             if specific:
                 for timer in plot_config["timers"]:
+                    label = None
```
twicki, line 79: I think stripping the plot down to only the most relevant components makes a lot of sense

twicki (approved):

PR #266, opened 2021-02-25
Fix/global config

rheacangeo:
## Purpose

Fix a typo that cause a mix of halo updates being on and off, resulting in a crash in unrecognized requests when running the standalone dynamics.py in compilation mode with halo updates 'off.'


rheacangeo:
We don't actually need the change
`import fv3core.utils.global_config as global_config`
instead of the relative path import. I don't have a strong preference either way, though most imports aren't relative so this matches other imports better. But if you want to undo that change I don't mind.

twicki (approved):
lgtm

PR #267, opened 2021-02-25
Move to the new name of generated caches

twicki:
## Purpose

By renaming the jenkins plan, we need to change what we search and replace in the py files to get the cache to work properly


rheacangeo (approved):

twicki:
launch jenkins

PR #269, opened 2021-02-26
Lower dimensional grid and input variables

eddie-c-davis:
## Purpose

This PR refactors all stencils to be able to use lower dimensional input data and grid variables.

## Code changes:

- Serialized input fields and grid variables are no longer automatically expanded to 3D.
- Inputs or grid variables that do need to be 3D are expanded in the `make_storage` methods.
- The `read_only` input is added to the `make_storage` methods to control field expansion.
- The `full_shape` member is added to storage info dictionaries in the translate methods to prevent reduction of field dimensions from I,J, or K to 1.
- The `gtstencil` decorator passes an origin dictionary to the gt4py stencil decorator to allow fields to have different origins.
- The `FV3StencilObject._get_field_origins` method is added to generate the origin dictionary.
- Type hints are added to all modified stencils where they did not already exist.
- Type hints are modified to reflect the dimensionality of the input fields.
- Field accesses within stencils are modified when fields are reduced, e.g., `[0,1,0]` to `[0,1]`.
- Fields allocated between stencil calls are removed if no longer required.

<b>NOTE</b>: CI tests for the GT backends will not work at this time....

## Checklist


fv3core/stencils/d2a2c_vect.py
```diff
@@ -298,84 +296,83 @@ def compute(dord4, uc, vc, u, v, ua, va, utc, vtc):
...
+    domain_edge_x = (1, grid.njc + 2, grid.npz)
+    jslice = slice(grid.js - 1, grid.je + 2)
+    if grid.west_edge and not grid.nested:
+        vol_conserv_cubic_interp_x(
+            utmp, uc, origin=(i1, j1, 0), domain=domain_edge_x
+        )
+        islice = slice(grid.is_ - 2, grid.is_ + 2)
+        utc[grid.is_, jslice, -1] = edge_interpolate4_x(
+            ua[islice, jslice, -1], grid.dxa[islice, jslice]
+        )
```
rheacangeo, line 308: oh man I'm so sorry, there is so much python code you have to deal with here that just gets removed by the open PRs! 
I think this may be one of the reasons d2a2c is not validating, the edge interpolate 4 should be called over the whole column, previously was ':'. I think python won't like that though (combining arrays of different shapes), so we might just want to keep it simple and add a loop over k, that'll get blown away anyways when we have the d2a2c as a single stencil

fv3core/stencils/d2a2c_vect.py
```diff
@@ -298,84 +296,83 @@ def compute(dord4, uc, vc, u, v, ua, va, utc, vtc):
...
+
+    if grid.east_edge and not grid.nested:
+        vol_conserv_cubic_interp_x(
+            utmp, uc, origin=(nx - 1, j1, 0), domain=domain_edge_x
+        )
+        islice = slice(nx - 2, nx + 2)
+        utc[nx, jslice, -1] = edge_interpolate4_x(
+            ua[islice, jslice, -1], grid.dxa[islice, jslice]
+        )
```
rheacangeo, line 346: same as a above

fv3core/stencils/d2a2c_vect.py
eddie-c-davis: Thank you!

fv3core/utils/gt4py_utils.py
```diff
@@ -486,3 +538,42 @@ def squeeze(array, axis: Union[int, Tuple[int]] = None):
...
+                arr_2d = repeat(array[:, np.newaxis], new_shape[1], axis=1)
+                return repeat(arr_2d[:, :, np.newaxis], new_shape[2], axis=2)
+        else:
+            return array.reshape(new_shape)
+    return array
+
+
+def unique(
+    array,
+    return_index: bool = False,
```
rheacangeo, line 561: oh bummer that you needed to reimplement all these numpy operations? hopefully we can eliminate all/most at some point (or relegate them to initialization usage only)

fv3core/utils/gt4py_utils.py
```diff
@@ -349,6 +380,27 @@ def make_storage_dict(
     return data_dict
 
 
+# def k_slice_operation(key, value, ki, dictionary):
```
rheacangeo, line 383: will we need this later?

rheacangeo (approved):
Really amazing work! Great job getting master back in and all the mess. 'write_vars' vs 'out_vars' is a little confusing, but I think I get why it is needed. No objections to getting this merged. If I come across anything later that seems like we should discuss and change, we can always open a new PR.

fv3core/utils/gt4py_utils.py
eddie-c-davis: Agreed, I also hope that we can remove these soon in upcoming PRs.

fv3core/utils/gt4py_utils.py
eddie-c-davis: Happy to remove the comment, is that what you mean?

PR #270, opened 2021-02-26
bumping data version to 7.2.5

rheacangeo:
## Purpose

We are updating the data version, because this one is the latest one generated that is running and succeeding in jenkins. 

## Code changes:

- Makefile indicates data version is 7.2.5
- updating the CS_Profile_2d test code to work with a slight change to the serialization statements in fv3gfs-fortran

### DIFF
--- a/tests/translate/translate_remap_profile_2d.py
+++ b/tests/translate/translate_remap_profile_2d.py
@@ -1,6 +1,7 @@
 import numpy as np
 
 import fv3core.stencils.remap_profile as Profile
+import fv3core.utils.gt4py_utils as utils
 from fv3core.testing import TranslateFortranData2Py
 
 
@@ -9,7 +10,6 @@ class TranslateCS_Profile_2d(TranslateFortranData2Py):
         super().__init__(grid)
         self.compute_func = Profile.compute
         self.in_vars["data_vars"] = {
-            "qs": {"serialname": "qs_column"},
             "a4_1": {"serialname": "q4_1"},
             "a4_2": {"serialname": "q4_2"},
             "a4_3": {"serialname": "q4_3"},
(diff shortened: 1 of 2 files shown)


tests/translate/translate_remap_profile_2d.py
```diff
@@ -63,7 +63,8 @@ def compute(self, inputs):
         inputs["i1"] = self.grid.global_to_local_x(inputs["i1"] - 1)
         inputs["i2"] = self.grid.global_to_local_x(inputs["i2"] - 1)
         inputs["jslice"] = slice(0, 1)
-
+        if "qs" not in inputs:
+            inputs["qs"] = utils.make_storage_from_shape(self.maxshape)
```
ofuhrer, line 67: Makes perfect sense, since `remap_profile.compute()` expects `qs` as one of the inputs.

What I'm not sure how we'll deal with in the long-run, is that we actually don't want to call `make_storage_from_shape` in any of the compute methods when running the dycore, e.g.
```
    qs = utils.make_storage_from_shape(pe1.shape, origin=(0, 0, 0))
```
in `mapn_tracer.compute()`, but this is not related to this PR.

ofuhrer (approved):
Thanks. LGTM!

rheacangeo:
launch jenkins

tests/translate/translate_remap_profile_2d.py
rheacangeo: agreed! we need to get rid of those as much as possible. We are reducing some of them by putting more code into stencils, but there will still be some we may need to push up a few levels and pass through, or something.

rheacangeo:
launch jenkins

rheacangeo:
launch jenkins

rheacangeo:
launch jenkins

rheacangeo:
launch jenkins

PR #271, opened 2021-02-26
divergence damping, and bgrid/dgrid corners use regions

rheacangeo:
## Purpose

In our ongoing quest to reduce the volume of straight Python code outside of stencils, this PR makes a step towards reducing the number of stencil calls in divergence_damping. Filling B-grid and vector D-grid corners is done expressly hard coded for 3 halo lines for now, as their pattern do not match gt4py features. Divergence damping itself has a reduced number of stencils, but there are still a few left. It looks clear how to unroll this and get rid of column_split_compute, but a2b_ord4 needs to do it first. The loop over higher order nord used the information that nord is only ever 0, 1, or 2. 

## Code changes:

- divergence_damping has less stencils
- fv3core/util/corners has updates to fill_corners for bgrid winds and dgrid winds
- test code for divergence damping and these corners are updated

## Checklist


fv3core/stencils/basic_operations.py
```diff
@@ -39,10 +39,15 @@ def copy(q_in, origin=(0, 0, 0), domain=None):
...
+@gtscript.function
+def adjustmentfactor(adjustment: sd, q_out: sd):
+    return q_out * adjustment
+
+
 @gtstencil()
 def adjustmentfactor_stencil(adjustment: sd, q_out: sd):
     with computation(PARALLEL), interval(...):
-        q_out[0, 0, 0] = q_out * adjustment
+        q_out[0, 0, 0] = adjustmentfactor(adjustment, q_out)
```
oelbert, line 50: Is this a pattern we're moving to now? Otherwise I'm not sure why to put the calculation in a function vs. leaving it in a stencil.

rheacangeo: it's mostly because I made a function, so I might as well call it, and I was thinking we'd probably eventually delete the stencil version of it. But, I can undo this, it does a layer of obscurity to it.

fv3core/stencils/divergence_damping.py
```diff
@@ -143,16 +146,114 @@ def vorticity_calc(wk, vort, delpc, dt, nord, kstart, nk):
...
+):
+
+    with computation(PARALLEL), interval(...):
+        delpc = divg_d
+        # TODO, can we call the same function 3 times, let gt4py do the extent analysis?
+        # currently does not work because corner calculations need entire array,
+        # and vc/uc need offsets
+        # Another TODO: once there is an idea of 'repetition', we need to loop
+        # namelist.nord times, not always 3 times, that is just what we have
+        # it set to. nord = 0, 1, or 2 are possible options
```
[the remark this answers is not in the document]
rheacangeo: I could, though wouldn't be able to test it without making more data, unlike delnflux which actually does get called with different 'nords'. This approach though would make it more complete and not require a new gt4py feature

fv3core/stencils/divergence_damping.py
```diff
@@ -1,58 +1,84 @@
...
-def corner_south_remove_extra_term(vort: sd, delpc: sd):
-    with computation(PARALLEL), interval(...):
-        delpc[0, 0, 0] = delpc - vort[0, -1, 0]
+@gtscript.function
+def remove_extra_term_south_corner(extra: FloatField, field: FloatField):
+    # from __externals__ import i_start, i_end,  j_start
+    # TODO: why does this not work?
+    # with horizontal(region[i_start, j_start], region[i_end + 1, j_start]):
+    #    field = field - extra[0, -1, 0]
+    # return field
```
oelbert, line 70: I don't know, but I'd also like an answer. I guess we just have to do regions outside of function calls?

fv3core/stencils/divergence_damping.py
```diff
@@ -162,98 +263,81 @@ def compute(
     kstart=0,
     nk=None,
 ):
+    """Applies divergence damping to the momentum equations
+
+    The divergence damping term of the momentum equation is computed using
+    the double laplacian of the horizontal divergence oon a Lagrangian surface.
+    The vorticity, kinetic energy, divergence and C-grid winds are updated
+    accordingly.
+    Assumes this is not a nested grid.
```
oelbert, line 272: Very nice

fv3core/stencils/divergence_damping.py
```diff
@@ -143,16 +146,114 @@ def vorticity_calc(wk, vort, delpc, dt, nord, kstart, nk):
...
+        vc = vc_from_divg(divg_d, divg_u)
+    divg_d = corners.fill_corners_bgrid_y(divg_d)
+    with horizontal(region[local_is - 2 : local_ie + 4, local_js - 3 : local_je + 4]):
+        uc = (divg_d[0, 1, 0] - divg_d) * divg_v #uc_from_divg(divg_d, divg_v)
+    vc, uc = corners.fill_corners_dgrid(vc, uc, -1.0)
+    divg_d = update_divg_d(rarea_c, divg_d, uc, vc)
+    return divg_d, uc, vc
+
+@gtscript.function
+def divergence_subsequent1(
```
oelbert, line 187: Might suggest a slightly different name here

fv3core/stencils/divergence_damping.py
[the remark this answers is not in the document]
oelbert: I was mostly kidding about the unrolled implementation since I did that in delnflux. Once we have repetition both delnflux and this should probably use it.

In the meantime, though, if this implementation is only for nord=2, would it make sense to say that in the names or docstrings somewhere? Just so that it's clear that we may have to add other nord modes when we refactor?

fv3core/stencils/basic_operations.py
oelbert: I was mostly curious, I don't think you need to revert this

rheacangeo:
Closing because way out of date and regions working well in GT backends is not coming anytime soon.

PR #272, opened 2021-02-26
Test/barrier

twicki:
## Purpose

Adds a barrier to the end of the call to avoid `slurmstepd` to cancel jobs if they are slightly out of sync


eddie-c-davis (approved):
Looks good to me.

PR #273, opened 2021-02-26
use caches for performance tests

twicki:
## Purpose

We also use the generated caches for the performance runs and for future cache builds

## Code changes:

- the `run_on_daint` script now also consumes the caches generated nightly


rheacangeo (approved):
Happy to approve, but I thought we ran into issues with this before? Probably had to do with failed jobs not deleting their data, which we have resolved.

PR #274, opened 2021-02-26
only copy caches if they already exist

rheacangeo:
## Purpose

We only need to load in existing caches for the gt backends, it is fine to not copy in numpy caches. This change stops the jenkins plan fv3core-performance-test from failing on the numpy backend


eddie-c-davis (approved):
Looks great!

PR #275, opened 2021-03-01
fix zh/gz copy order in dyn_core

rheacangeo:
## Purpose

In the first timestep, zh is copied from gz, for all subsequent timesteps, gz is copied from zh. Somehow this did not happen properly in the first version and our test code doesn't catch it. 

## Code changes:

- minor change to dyn_core


eddie-c-davis (approved):
Nice catch!

PR #277, opened 2021-03-01
Regions/delnflux

oelbert:
## Purpose

This PR uses regions to merge stencils in delnflux.py. I've also added some implementations of compute_no_sg for comparison. One maintains the previous structure, one merges the stencils inside the loop over `nord` (which requires defining the stencil in each iteration), and the current implementation unrolls the `nord` loop explicitly to fuse the stencils. I've also added some docstrings and started changing the names of a few stencils to increase readability.

## Code changes:

`compute_delnflux_no_sg`: merged `diffusive_damp()` and `add_diffusive_component()` x and y components to be single stencils. 

`compute_no_sg`: merged `copy_corners` and flux calculations into one `fxy_stencil` call. Also implemented 3 versions of the loop over nord. `compute_no_sg_multi_loop` keeps the loop structure we had previously, `compute_no_sg_merge_loop` uses regions and externals to merge the stencil computation inside the loop (though the decorator is also invoked in the nord loop), and `compute_no_sg_unroll` uses one stencil that unrolls the loop over nord to merge the interior stencils together (and also with `fxy_stencil`). Once it's implemented we'll likely want to use the new `with repetition` functionality to refactor `compute_no_sg_multi_loop`, pulling the nord loop inside the stencil so it's only defined once and keeps a compact form, as this will likely be the best combination of readability and performance.

## Checklist


oelbert:
After opening the PR I realized that I pulled Rhea's copy-corners improvements in here as well, so we'll have to wait for that to get resolved anyway

fv3core/stencils/delnflux.py
```diff
@@ -1,94 +1,404 @@
...
+    fx: FloatField, fx2: FloatField, fy: FloatField, fy2: FloatField
+):
+    from __externals__ import local_ie, local_je
+
     with computation(PARALLEL), interval(...):
-        fx[0, 0, 0] = fx + fx2
+        with horizontal(region[:, : local_je + 1]):
+            fx[0, 0, 0] = fx + fx2
+        with horizontal(region[: local_ie + 1, :]):
+            fy[0, 0, 0] = fy + fy2
```
rheacangeo, line 109: I assume these local regions are necessary to validate? Curious, if you don't include them, I am guessing delnflux and d_sw will fail, but will dyncore? Would be helpful to mark whether these regions are just needed for local validation, or if they are important all the way up the chain. Usually would look for the things using fx and fy to see if gt4py could extend the domain for us here, but looks like it's pretty far away from that, so we probably need these for now.

fv3core/stencils/delnflux.py
```diff
@@ -1,94 +1,404 @@
...
+):
+    from __externals__ import local_ie, local_je
+
     with computation(PARALLEL), interval(...):
-        fx[0, 0, 0] = fx + 0.5 * damp * (mass[-1, 0, 0] + mass) * fx2
-        fy[0, 0, 0] = fy + 0.5 * damp * (mass[0, -1, 0] + mass) * fy2
+        with horizontal(region[:, : local_je + 1]):
+            fx[0, 0, 0] = fx + 0.5 * damp * (mass[-1, 0, 0] + mass) * fx2
+        with horizontal(region[: local_ie + 1, :]):
+            fy[0, 0, 0] = fy + 0.5 * damp * (mass[0, -1, 0] + mass) * fy2
```
rheacangeo, line 127: The same question applies here, probably the same answer give it is just an if else mass is None that triggers it

fv3core/stencils/delnflux.py
```diff
@@ -100,93 +410,256 @@ def compute_delnflux_no_sg(
...
+            higher_order_compute_unroll3(
+                fx,
+                fy,
+                grid.rarea,
+                d2,
+                grid.del6_u,
+                grid.del6_v,
+                origin=(grid.isd, grid.jsd, kstart),
+                domain=(grid.nid, grid.njd, nk),
+            )
```
rheacangeo, line 640: ah this is more complicated than divergence damping, because it looks like delnflux might in fact get called with different nords, while divergence damping always has nord = namelist value or 0.

fv3core/stencils/delnflux.py
```diff
@@ -1,94 +1,404 @@
...
+            ]
+        ):
+            fx = fx_calculation(d2, del6_v, 3)
+        d2 = corners.copy_corners_y(d2)
+        with horizontal(
+            region[
+                (local_is - 0) : (local_ie + 0 + 1), (local_js - 0) : (local_je + 0 + 2)
+            ]
+        ):
+            fy = fy_calculation(d2, del6_u, 3)
```
rheacangeo, line 375: It would be helpful for me to see the unrolled versions with functions, so it is easier to see what is repeated and altered.

rheacangeo (commented):
Oh wow! It will be so awesome when we can get this one to validate

fv3core/stencils/delnflux.py
rheacangeo: You may also then get to reuse some of the functions in the other unroll function

fv3core/stencils/delnflux.py
```diff
@@ -1,94 +1,404 @@
...
 
 
 @gtstencil()
-def fx2_order(q: sd, del6_v: sd, fx2: sd, order: int):
+def fx_calc_stencil(q: FloatField, del6_v: FloatField, fx: FloatField, order: int):
     with computation(PARALLEL), interval(...):
-        fx2[0, 0, 0] = del6_v * (q[-1, 0, 0] - q)
-        fx2[0, 0, 0] = -1.0 * fx2 if order > 1 else fx2
+        fx[0, 0, 0] = del6_v * (q[-1, 0, 0] - q)
+        fx[0, 0, 0] = -1.0 * fx if order > 1 else fx
```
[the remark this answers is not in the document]
oelbert: That makes sense, though I was also thinking of just deleting these stencils since they're not called anymore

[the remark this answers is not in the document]
rheacangeo: Oh yeah if you plan to delete them that's fine, it'd only be helpful if we plan to keep both versions around for awhile

fv3core/stencils/delnflux.py
eddie-c-davis: Are the `nord` values known at compile time? Maybe something we could unroll in the backend later?

eddie-c-davis (approved):
Couple of small questions, but this is impressive work! Hopefully we have verification soon.

fv3core/stencils/delnflux.py
oelbert: It does seem to work, at least in the local tests?

fv3core/stencils/delnflux.py
oelbert: We can abstract most of the unrolled loop calculations, but the first one explicitly depends on nord, so it would be a separate function for each unrolled loop. I've tried to clean it up in the latest commit

oelbert:
@eddie-c-davis Precisely. Because nord is known at compile-time we can unroll the loops in the backend once there's an iteration concept in gy4py stencils. For now manually unrolling the loops is the most performant thing to do, but ideally we'll be able to use something like `compute_no_sg_merge_loop` but with the loop over nord inside the stencil. Either way we can probably get rid of `compute_no_sg_multi_loop` now...

oelbert:
launch jenkins

PR #278, opened 2021-03-02
use existing data rather than always untarring

rheacangeo:
## Purpose

We keep running out of space on scratch. This is partly due to the multiple copies of 2-3 GB of data every performance test. The gt caches bring another 3 gb of data and it pretty quickly fills up. 
Here,  instead of copying the data, and untarring it multiple times for every performance run, keep the untarred data with all the other data and use it. This PR currently does not handle all the steps to make sure the data is there nd input.yml is there, so more could be done to make it easier to use a new dataset. 
This may not work for very high resolution data if we increase this. But currently the tarred c128 data is 2 GB and untarred is 3gb, so seems like it could be worthwhile. 

I don't mind closing this if it is preferable to not do this. It may be enough to delete the tar file after unzipping it as propose n another PR> 
## Infrastructure changes:

- Update the jenkins plan code that relied on the run_on_daint data to be in a local directory named test_data

## Checklist

### DIFF
--- a/.jenkins/actions/full_cache_build.sh
+++ b/.jenkins/actions/full_cache_build.sh
@@ -3,19 +3,19 @@ set -e -x
 SCRIPT=`realpath $0`
 SCRIPTPATH=`dirname $SCRIPT`
 ROOT_DIR="$(dirname "$(dirname "$SCRIPTPATH")")"
-TESTDATA_PATH="/project/s1053/fv3core_serialized_test_data/"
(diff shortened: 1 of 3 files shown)


twicki (commented):
lgtm

twicki (approved):

ofuhrer (commented):
This has the risk of failing, since there is a cleaning policy on /scratch and I sometimes clean away stuff there.

rheacangeo:
Yes totally, this doesn't solve all the data issues, mostly trying to stop filling up scratch, we should do a follow-up task 
We could use the 'make get_test_data' from the Makefile and/or just keep the data in /project.

PR #279, opened 2021-03-02
pk3_halo.py update

gmao-ckung:
## Purpose

This PR reduces the number of stencil call by utilizing horizontal regions within the stencil call to define the halo computation domain.

## Code changes:

`fv3core/stencils/pk3_halo.py`
- In `compute`, the 4 `edge_pe` stencil calls are replaced with a single `edge_pe_update` stencil call that incorporates horizontal regions.  Also, the two three-dimensional fields, `pei` and `pej`, allocated as temporaries are replaced by a two-dimensional field temporary `pe_tmp`.  
- In the `edge_pe_update` stencil, horizontal regions are defined to address the halo computation domain.  The `pk3` computation is rolled into the `FORWARD` computation.  Also, `pk3 = exp(akap * log(pe))` is replaced with `pk3 = pe ** akap` since a potential horizontal region bug was discovered.  Using the `pk3` computation with `log` and `exp` results in an inconsistent data type error when multiple `region` declarations are used within a `with horizontal` declaration.
- Type references to `utils.sd` are replaced using `FloatField` and `FloatFieldIJ` from `fv3core.utils.typing`

### DIFF
--- a/fv3core/stencils/pk3_halo.py
+++ b/fv3core/stencils/pk3_halo.py
@@ -1,63 +1,49 @@
-from gt4py.gtscript import FORWARD, PARALLEL, computation, exp, interval, log
+from gt4py.gtscript import FORWARD, computation, horizontal, interval, region
 
 import fv3core._config as spec
 import fv3core.utils.gt4py_utils as utils
 from fv3core.decorators import gtstencil
+from fv3core.utils.typing import FloatField, FloatFieldIJ
 
 
-sd = utils.sd
 # TODO merge with pe_halo? reuse partials?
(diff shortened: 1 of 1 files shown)


rheacangeo:
Very cool! This seems to fail DynCore with 54 ranks, FYI

oelbert (approved):
Good stuff. If you want to add a docstring to the compute function that would be a nice bonus, but otherwise this looks good.

gmao-ckung:
Using the "local" horizontal region variables seemed to fix the issues with the 54 rank tests!

PR #281, opened 2021-03-04
Specify gt4py version using GT4PY_TAG, remove gt4py image

rheacangeo:
## Purpose

We need to be able to continue to develop gt4py without potentially impacting fv3core master's regression and performance testing. Here we pin the repo to a specific gt4py tag, and will have to manually update it when we think a new version of gt4py:develop is ready to use. The plan that triggers from new build of gt4py:develop will need to pass its commit hash as the variable GT4PY_VERSION to the plan that runs gt4py tests in the fv3core image, and also trigger fv3core-regressions_PR, which will use this version of gt4py o test fv3core using all the datasets we require for a fv3core commit to master. 

## Infrastructure changes:

- add GT4PY_VERSION to Makefile structure, as a build-arg to the Dockerfile, and to the jenkins environment
- Remove processes around updating and using a gtp4y_develop docker image
- Use a daint_venv submodule that also uses this tag


rheacangeo:
launch jenkins

rheacangeo:
launch jenkins

rheacangeo:
launch jenkins

jdahm:
This looks really good. I think this enables what we want to do with the CI too!

twicki (approved):
looks really good.

PR #282, opened 2021-03-04
adding an action to get test data on daint

rheacangeo:
## Purpose

The regression tests no longer have 'touchstone' configurations, so when the data on daint needs to be updated, the multiconfiguration project messes this up and usually fails. Adding a simple action to make sure daint has the test data for the datasets we are using, to be added to the fv3core-daint-setup plan that runs before the fv3core-regressions_PR runs. 

## Infrastructure changes:

- added a jenkins actions get_test_data


rheacangeo:
launch jenkins

rheacangeo:
launch jenkins

oelbert (approved):
awesome to have, thanks!

PR #283, opened 2021-03-05
get_test_data action needs EXPERIMENT to be specified

rheacangeo:
## Purpose

 #282 set the host location for the data correctly, but neglected to set the EXPERIMENT variable, which is needed to find the cloud bucket to pull from. This fixes that.


oelbert (approved):
oop

Sorry for not catching this last time...

PR #286, opened 2021-03-09
Update profiling

ofuhrer:
## Purpose

Current profiling also profiles init as well as warm-up pass. Furthermore, it is not MPI-aware and the *.prof file which is written probably comes from a random rank. It also had the issue of actually writing performance results to the `/project` directory when being run (see many recent points on the performance plot).

## Purpose

This PR addresses the aforementioned shortcomings, by integrating profiling directly into the `dynamics.py` script with a new `--profiling` option (instead of preloading the `cProfile` module).

## Code changes:

`dynamics.py`: Add new argument `--profile` to activate profiling with `cProfile`. Profiling is MPI-aware and writes a `*.prof` file from each MPI rank. Also, profiling is only enabled around the actual hotloop (ignoring any initialization or the warm-up timestep).

`run_on_daint.sh`: Added a new option `run_args` which allows to pass runtime arguments to the `dynamics.py` script being executeed. This script has been modified, so that it's possible to run it multiple times manually on Piz Daint without errors (i.e. cleans away pre-existing clones, installations or virtual environments). The time for the actual run job has been increased to 40 minutes (from 30 mins, which was right at the limit of the actual time taken). Also, the resulting `*.json` files are only rsync'ed to a destination in case a `target_dir` is actually set.

`run_profile.sh`: Remove the `py_args` and replace with a `run_args` which enables profiling. Install Python virtual environment in top directory and name it `venv` instead of `test_ve`. Adapt to new naming of `*.prof` files and process the `*.prof` file from MPI rank 0 (which should always exist). Print the top 200 entries of the profile in terms of cumulative sum and in terms of number of calls. 

`full_cache_build`: Use new option to pass an empty `target_dir` to the `run_on_daint.sh` script.

## Checklist

- [x] Docstrings and type hints are added to new and updated routines, as appropriate.
- [x] All relevant documentation has been updated or added (e.g. README, CONTRIBUTING docs).
- [x] Unit tests are added or updated for non-stencil code changes.

### DIFF
--- a/examples/standalone/benchmarks/run_on_daint.sh
+++ b/examples/standalone/benchmarks/run_on_daint.sh
@@ -25,6 +25,7 @@ ROOT_DIR="$(dirname "$(dirname "$(dirname "$SCRIPTPATH")")")"
 # check sanity of environment
 test -n "$1" || exitError 1001 ${LINENO} "must pass a number of timesteps"
 timesteps="$1"
+
 test -n "$2" || exitError 1002 ${LINENO} "must pass a number of ranks"
 ranks="$2"
 
@@ -44,9 +45,8 @@ if [ -z "$5" ] ; then
 fi
 
 py_args="$6"
-if [ -z "$6" ] ; then
-    py_args=""
-fi
+
+run_args="$7"
 
 # set up the virtual environment
 cd $ROOT_DIR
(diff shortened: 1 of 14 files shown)


examples/standalone/benchmarks/run_on_daint.sh
```diff
@@ -121,14 +125,16 @@ sed -i s/\<NTASKS\>/$ranks/g run.daint.slurm
 sed -i s/\<NTASKSPERNODE\>/1/g run.daint.slurm
 sed -i s/\<CPUSPERTASK\>/$nthreads/g run.daint.slurm
 sed -i s/--output=\<OUTFILE\>/--hint=nomultithread/g run.daint.slurm
-sed -i s/00:45:00/00:30:00/g run.daint.slurm
+sed -i s/00:45:00/00:45:00/g run.daint.slurm
```
eddie-c-davis, line 128: Is this line now redundant?

examples/standalone/benchmarks/run_on_daint.sh
```diff
@@ -44,48 +45,51 @@ if [ -z "$5" ] ; then
...
-pip install external/fv3gfs-util/
+source ./venv/bin/activate
+echo "install requirements..."
+pip install ./external/fv3gfs-util/
 pip install .
-
 pip list
 
 # set the environment
+if [ -d ./buildenv ] ; then rm -rf buildenv ; fi
```
eddie-c-davis, line 68: Nice catch, this one bit me the other day...

examples/standalone/runfile/dynamics.py
```diff
@@ -182,7 +197,16 @@ def write_global_timings(experiment, filename, comm, root=0):
                 timer,
             )
 
+    if profiler is not None:
+        profiler.disable()
+
     timer.stop("total")
+
+    # output profiling data
+    comm.Barrier()
```
eddie-c-davis, line 206: Can we get by with one barrier? No comm in between I guess...

eddie-c-davis (approved):
A few small comments, but looks great!

examples/standalone/benchmarks/run_on_daint.sh
ofuhrer: Yes, I retained it for possible future changes to be easier.

examples/standalone/runfile/dynamics.py
ofuhrer: Will remove.

twicki (approved):
lgtm

PR #287, opened 2021-03-09
Combine pe_halo.py stencils by using horizontal regions

gmao-ckung:
## Purpose

The four stencil calls within the `compute` routine in `pe_halo.py` are combined into one stencil using horizontal regions.

## Code changes:

All code changes are in `/fv3core/stencils/pe_halo.py`.

- Use the `FloatField` type instead of `utils.sd` type for Field variables
- Rewrite `edge_pe` to use horizontal regions that combines the 4 `edge_pe` calls in `compute` into one call
- Define the `origin` as `grid.full_origin()` and `domain` as `grid.domain_shape_full(add(0,0,1))` when calling `edge_pe`


oelbert (approved):
Looks good, thanks for doing this!

PR #288, opened 2021-03-09
jenkins.sh should only set the GT4PY_VERSION if needed

rheacangeo:
## Purpose
jenkins.sh was setting GT4PY_VERSION so that the virtualenv could pick it up (and any other action). But we shoud only set it to what is in the Makefile if it is not set in the environment, which it is when testing gt4py.


twicki (approved):
lgtm!

PR #289, opened 2021-03-10
Revert performance plot back to linear y-axis

ofuhrer:
## Purpose

This PR reverts the y-axis of the performance plot back to linear and removes the numpy backend from the plot.


twicki (commented):
going back to linear scale we're probably no longer able to see any meaningful change in the fortran code. Should we add a separate plot just for that?

ofuhrer:
history_absolute_time
history_per_timestep_only_fortran
history_per_timestep

PR #290, opened 2021-03-10
Memoize outputs of make_storage_from_shape

mcgibbon:
## Purpose

Currently we re-allocate storages in many places within loops. This PR adds memoization to `make_storage_from_shape`, so that the same storage can be provided each time the same storage is requested. There are some caveats and gotchas to this implementation, and in the long run we should replace it with a safer implementation.

The key used for memoization is the arguments used combined with the calling scope file and line number, as well as the file and line number which called in to that scope. This handles cases where a utility function (such as `copy`) calls our `make_storage_from_shape`, since `copy` will be called from different places each time. This does *not* handle any more deeply nested duplicate calls, such as if another utility function were to call `copy`, and does not handle allocations which take place within for loops, such as tracer allocations. In those cases, memoization will provide the same storage to two conceptually different objects, causing a bug. For this reason, and because of the significant overhead cost of `inspect`, we should move away from this implementation in the longer term.

In the short term, this reduces our time spent allocating storages on each timestep in our profiling job from 130s to 5s, plus an inspect overhead of about 35s. When we move to another implementation, this inspect overhead can also be removed.

To avoid having to zero out the storages each time a memoized storage is returned, the default behavior for `make_storage_from_shape` has been changed to return `empty` arrays. In places where it is critical to initialize zeroed arrays, the call has been changed to explicitly set `init=True`.

## Code changes:

- Memoization has been added to `get_storage_from_shape`, keyed on the arguments as well as the filename and line number of the two frames above the current frame (the calling filename and line number, and also the filename and line number which called into that scope).
- Default value for `init` in `make_storage_from_shape` has been changed to `False`, and code which requires zero values has been modified to explicitly set `init=True`

## Checklist

- [x] You have followed the coding standards guidelines established at [Code Review Checklist](https://paper.dropbox.com/doc/Code-Review-Checklist--BD7zigBMAhMZAPkeNENeuU2UAg-IlsYffZgTwyKEylty7NhY) (to an extent - the memoization is a "side effect" done this way to minimize code changes, and the plan is to change its implementation in the future).
- [x] All relevant documentation has been updated or added (e.g. README, CONTRIBUTING docs) - this detail is documented only in the docstring for `make_storage_from_shape`.


rheacangeo:
launch jenkins

tests/test_memoize_storages.py
```diff
@@ -0,0 +1,59 @@
...
+    make_storage_from_shape_uncached,
+)
+
+
+# shape: Tuple[int, int, int],
+# origin: Tuple[int, int, int] = origin,
+# *,
+# dtype: DTypes = np.float64,
+# init: bool = False,
+# mask: Optional[Tuple[bool, bool, bool]] = None,
```
rheacangeo, line 14: are these left here on purpose?

fv3core/utils/gt4py_utils.py
```diff
@@ -259,6 +260,71 @@ def make_storage_from_shape(
     return storage
 
 
+storage_shape_outputs = {}
+
+
+@wraps(make_storage_from_shape_uncached)
```
rheacangeo, line 266: why do we need/want this decorator here?

fv3core/utils/gt4py_utils.py
```diff
@@ -259,6 +260,71 @@ def make_storage_from_shape(
...
+               qin.shape, origin=(grid().is_, grid().jsd, kstart)
+           )
+        3) q_out = utils.make_storage_from_shape(q_in.shape, origin, init=True)
+    """
+    # The caching used here is dangerous, in that e.g. if you call this in a
+    # loop with the same arguments you will get the same storage.
+    # This was implemented this way for fast results with minimal code
+    # changes.
+    # We should shift to an explicit caching or array re-use system down
+    # the line.
```
rheacangeo, line 310: We probably don't need to explain this both in the docstring and in a comment

fv3core/stencils/basic_operations.py
```diff
@@ -20,21 +20,30 @@ def copy_stencil(q_in: sd, q_out: sd):
         q_out = q_in
 
 
-def copy(q_in, origin=(0, 0, 0), domain=None):
+def copy(q_in, origin=(0, 0, 0), domain=None, use_cache=True):
```
rheacangeo, line 23: hopefully at some point soon we don't need this function at all, and all copies are done in another stencil

fv3core/stencils/fxadv.py
```diff
@@ -235,7 +235,7 @@ def update_ut_x_edge(uc, cosa_u, vt, ut):
 
 def compute_vt(uc_in, vc_in, cosa_v, rsin_v, sin_sg2, sin_sg4, vt_in):
     vt_origin = (grid().isd, grid().js - 1, 0)
-    vt = utils.make_storage_from_shape(vt_in.shape, vt_origin)
+    vt = utils.make_storage_from_shape(vt_in.shape, vt_origin, init=True)
```
rheacangeo, line 238: soon this stencil will be refactored and this won't exist. but it's small enough the merge conflict should be fine.

rheacangeo (approved):
Seems like a good solution for the moment. We should be removing storage allocations rather than making more, and definitely not making more complex ones. So if this works for the tests, it's probably good to go. Once the tests pass that is. Well documented and actually added tests. I have a couple of comments and questions, but nothing blocking.

rheacangeo:
launch jenkins

fv3core/stencils/basic_operations.py
mcgibbon: Yes, since it's a two-line function we could replace it already pretty easily, wherever it is called. I think it would be fine to do that whenever, to move all the calls to `make_storage_from_shape` to the same place (the compute functions).

fv3core/utils/gt4py_utils.py
mcgibbon: To turn the *args **kwargs signature into the real call signature.

fv3core/utils/gt4py_utils.py
mcgibbon: I agree with you in general - I noticed this, but then thought "well, it's dangerous enough that it's probably worth saying twice". For now, I want to avoid the situation where someone editing the function glosses over the docstring, and where someone reading docstrings doesn't notice a comment inside the code.

Let's not keep this duplication in our end product, but this caching logic will be removed by then.

tests/test_memoize_storages.py
mcgibbon: No, just used as a reference when writing. Removed, thanks!

PR #291, opened 2021-03-10
jenkins tag should be fixed whether or not it exists already

rheacangeo:
## Purpose

We had a workflow where the JENKINS_TAG got assigned to a valid tag if it didn't exist yet, but a new plan sent through bad values from a multiconfiguration process, and this didn't catch it. Now it should. 
Also fix a bug where by run-command failures were not causing the jenkins script to fail when run through docker (vs a slurm job)
## Infrastructure changes:

- .jenkins/jenkins.sh redefines JENKINS_TAG to something valid whether it exists or not
- .jenkins/jenkins.sh checks for slurm job failures directly after running
- Update daint_venv submodule to point back at master now that the GT4PY_VERSION pr on that repo is merged

### DIFF
--- a/.jenkins/jenkins.sh
+++ b/.jenkins/jenkins.sh
@@ -142,12 +142,11 @@ export FV3_STENCIL_REBUILD_FLAG=False
 export TEST_DATA_HOST="${TEST_DATA_DIR}/${experiment}/"
 export EXPERIMENT=${experiment}
 if [ -z ${JENKINS_TAG} ]; then
-    if [ ${#JOB_NAME} -gt 85 ]; then
-	NAME=`echo ${JOB_NAME} | md5sum | cut -f1 -d" "`
-    else
-	NAME=${JOB_NAME}
-    fi
-    export JENKINS_TAG=${NAME//[,=\/]/-}-${BUILD_NUMBER}
+    export JENKINS_TAG=${JOB_NAME//[,=\/]/-}-${BUILD_NUMBER}
+fi
+if [ ${#JENKINS_TAG} -gt 85 ]; then
+	NAME=`echo ${JENKINS_TAG} | md5sum | cut -f1 -d" "`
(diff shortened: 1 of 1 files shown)


.jenkins/jenkins.sh
```diff
@@ -142,12 +142,11 @@ export FV3_STENCIL_REBUILD_FLAG=False
 export TEST_DATA_HOST="${TEST_DATA_DIR}/${experiment}/"
 export EXPERIMENT=${experiment}
 if [ -z ${JENKINS_TAG} ]; then
-    if [ ${#JOB_NAME} -gt 85 ]; then
-	NAME=`echo ${JOB_NAME} | md5sum | cut -f1 -d" "`
-    else
-	NAME=${JOB_NAME}
-    fi
-    export JENKINS_TAG=${NAME//[,=\/]/-}-${BUILD_NUMBER}
+    export JENKINS_TAG=${JOB_NAME//[,=\/]/-}-${BUILD_NUMBER}
```
gmao-ckung, line 145: Since I'm only marginal at scripting, can you quickly explain what the `//[,=\/]/-` does?

gmao-ckung (approved):
Everything looks good!  I just had that one question for my benefit...

.jenkins/jenkins.sh
rheacangeo: We want the jenkins tag to be a unique identifier for the test case being run in case another job is running on the same machine, that they don't accidentally use the wrong image. docker tags can only have a limited character set ... but we want a tag that is based on the name of the project, JOB_NAME is the closest thing to what we want. Except, when using a multiconfiguration project (e.g. run the same plan on different machines and/or with different arguments), the JOB_NAME gets characters ',', '=' and '/', that are illegal as a tag.

rheacangeo: Oh you mean for scripting -- that does a find replace of those characters with a '-'

jdahm: should be replacing all `=` and `/` by `-`, I think.

PR #294, opened 2021-03-10
Select directories for gt4py tests

jdahm:
## Purpose

Selects directories for gt4py tests. Fixes gt4py tests.

## Code changes:

- `gt4py_tests_gpu` will now only run tests in certain directories.


eddie-c-davis (approved):
LGTM!

PR #296, opened 2021-03-11
Add requirements for GTC and Eve

jdahm:
## Purpose

Adds requirements for GTC and Eve to requirements.txt so that the gt4py tests can run.

## Code changes:

- Changes `requirements.txt`

## Requirements changes:

- Requires `factory-boy` and a few pytest addons.


rheacangeo (approved):
I think that pip-compile put all the comments below is harder to read, but eh. Otherwise looks good.

PR #300, opened 2021-03-12
reduce the filecount on daint scratch 

rheacangeo:
## Purpose

Remove the files that take up a lot of space and are not that helpful for debugging when run_on_daint fails, and only submodule update the ones we actually use right now

## Infrastructure changes:

- run_on_daint.sh will remove gt caches and the virtualenv when it fails
- Makefile by default only submodule updates the ones we use


twicki (approved):
looks good to me!

PR #303, opened 2021-03-12
Cleanup Jenkins actions for standalone

ofuhrer:
## Purpose

We have three scripts in the `.jenkins/actions/` directory to run the standalone. These are mostly redundant. We also have a script `run_on_daint.sh` which serves to manually run the standalone, but also potentially copies some artifacts to locations that should not be touched upon manual execution. This PR unifies the 3 scripts into a single script, adds a bit more safety, and centralizes the archival of artifacts from the standalone from Jenkins in a single place.

This PR explicitly does not do the following:
- Unify the way actions are being called from Jenkins by launching this script as `.jenkins/jenkins.sh run_standalone`, since this would potentially require a larger refactor.
- Do any significant changes in the responsibilities between `run_on_daint.sh` and `run_standalone.sh` except for artifact archival.
- Architectural changes to the CI system we currently have in place.

Finally, I have tested this PR manually on daint by running `.jenkins/actions/run_standalone` with/without `profile` or `build_cache` arguments. If this PR is merged, it will require corresponding changes on the relevant Jenkins plans.

## Code changes:

- `full_cache_build.sh`, `run_performance.sh`, and `run_profile.sh` have been removed.
- `run_standalone.sh` unifies the above 3 scripts, adds more safety and verbosity, and centralizes the archival of artifacts from the standalone.
- `process_profiling.sh` a new script that can be used to post-process profiling data (moved out of `run_profile.sh`, since this is also useful for manual execution).
- `run_on_daint.sh` minor cleanup, removal of `target_dir` and associated copying of timing information in order to avoid writing artifacts when being run manually
- `dynamics.py` prints a SUCCESS message if completed successfully for easier parsing of stdout for success.

## Infrastructure changes:

We need to change the Jenkins plans in the following way:
- `run_profile.sh` --> `run_standalone.sh profile`
- `full_cache_build.sh` --> `run_standalone.sh build_cache`
- `run_performance.sh` --> `run_standalone.sh`

## Checklist


.jenkins/actions/run_standalone.sh
```diff
@@ -0,0 +1,112 @@
...
+# only save timings if this is neither a cache build nor a profiling run
+if [ "${SAVE_CACHE}" != "true" -a "${DO_PROFILE}" != "true" ] ; then
+    SAVE_TIMINGS="true"
+fi
+
+# configuration
+SCRIPT=`realpath $0`
+SCRIPTPATH=`dirname $SCRIPT`
+ROOT_DIR="$(dirname "$(dirname "$SCRIPTPATH")")"
+FORTRAN_SERIALIZED_DATA_VERSION=7.2.5
```
rheacangeo, line 37: A lot of words for a very minor question:
Now that this is consolidated to one spot (yay!) for performance runs, do we want to keep the data version assigned specifically here, or grab it from the Makefile like .jenkins/jenkins.sh does here
export FORTRAN_VERSION=`grep "FORTRAN_SERIALIZED_DATA_VERSION=" Makefile  | cut -d '=' -f 2`
The risk of leaving it here is that we might forget to update it when we change the Makefile version. The risk of using the Makefile version is that we might forget to set up the c128 data in the project directory location and this will fail. The latter makes us stay up to date, the former allows us consistency with past runs, unless the data changes so much there are no longer the expected variables available for running the standalone. I don't mind either way, just wanted to bring it up.

.jenkins/actions/run_standalone.sh
```diff
@@ -0,0 +1,112 @@
...
+# run analysis and store profiling artifacts
+if [ "${DO_PROFILE}" == "true" ] ; then
+    echo "Analyzing profiling results"
+    ${BENCHMARK_DIR}/process_profiling.sh
+    echo "Copying profiling information to ${ARTIFACT_DIR}"
+    cp $ROOT_DIR/*.prof ${ARTIFACT_DIR}/
+fi
+
+# remove venv (too many files!)
+rm -rf $ROOT_DIR/venv
```
rheacangeo, line 110: I am sure it doesn't matter at the end here, but might do a 'deactivate' first?

examples/standalone/benchmarks/process_profiling.sh
```diff
@@ -0,0 +1,39 @@
+#!/bin/bash
+
+SCRIPT=`realpath $0`
+SCRIPT_DIR=`dirname $SCRIPT`
+ROOT_DIR="$(dirname "$(dirname "$(dirname "$SCRIPT_DIR")")")"
+
+# activate Python environment
+venv="$ROOT_DIR/venv"
```
rheacangeo, line 8: could pass this as an argument? though probably fine to leave as-is

examples/standalone/benchmarks/run_on_daint.sh
```diff
@@ -18,43 +23,31 @@ exitError()
...
-fi
-
-target_dir="$4"
-if [ -z "$4" ] ; then
-    target_dir="$ROOT_DIR"
-fi
-
-data_path="$5"
-if [ -z "$5" ] ; then
-    data_path="/scratch/snx3000/olifu/jenkins/scratch/fv3core_fortran_data/7.2.5/c12_6ranks_standard/"
```
rheacangeo, line 44: I could see this going either way. Having the defaults can be pretty convenient to getting started with a script with a lot of arguments, but can also see how one should probably figure them out pretty soon

examples/standalone/benchmarks/run_on_daint.sh
```diff
@@ -113,24 +104,31 @@ echo "submitting script to do compilation"
 sed -i s/\<NAME\>/standalone/g compile.daint.slurm
 sed -i s/\<NTASKS\>/$ranks/g compile.daint.slurm
 sed -i s/\<NTASKSPERNODE\>/1/g compile.daint.slurm
-sed -i s/\<CPUSPERTASK\>/$nthreads/g compile.daint.slurm
+sed -i s/\<CPUSPERTASK\>/$NTHREADS/g compile.daint.slurm
```
rheacangeo, line 107: I don't mind the change, but what is the motivation for the change? nthreads could be set the buildenv if we ever use that here (and we do use it to get the slurm script templates)

examples/standalone/benchmarks/run_on_daint.sh
```diff
@@ -140,8 +138,10 @@ sed -i "s#<CMD>#export PYTHONPATH=/project/s1053/install/serialbox2_master/gnu/p
...
-    rsync *.json $target_dir
+if grep -q SUCCESS slurm-*.out; then
+    echo "performance run sucessful"
+else
+    echo "ERROR: performance run not sucessful"
+    exit 1
 fi
-
-echo "performance run sucessful"
+mv -f slurm-*.out compile.daint.out
```
rheacangeo, line 147: You've be overwriting the output from the compilation step, did you mean for this to be run.daint.out?

rheacangeo (commented):
This consolidation of logic seems useful. There are benefits to how it was before, in that a person could tinker just with the pathway they were focused on without worrying about impacting the other pathways. But, it was also very easy for the pathways to get out of date with each other if we made a change in one of them and didn't duplicate it across them. run_standalone at first is several things to take in, but once you grasp it, seems like a convenient way to capture the different variations we have of running run_on_daint. I think the first approach was useful as we learned how to use the standalone and what we want from it, but I think it does make sense to consolidate now. None of us have had the bandwidth to approach it, so thank you for picking it up.

.jenkins/actions/run_standalone.sh
ofuhrer: Great suggestion. Done.

.jenkins/actions/run_standalone.sh
ofuhrer: There is no `source .../activate` in this script, so not needed.

examples/standalone/benchmarks/process_profiling.sh
ofuhrer: Since this script is intended to be able to run manually, I tried using the same logic as in `run_on_daint.sh` and avoided baking in too many arguments.

examples/standalone/benchmarks/run_on_daint.sh
ofuhrer: Yes. The integration with `jenkins.sh` and buildenv was left for a future PR. I just moved this to the configuration section to make it more visible.

examples/standalone/benchmarks/run_on_daint.sh
ofuhrer: Good catch.

examples/standalone/benchmarks/run_on_daint.sh
ofuhrer: Agreed. No strong feelings from my side. It felt like this was a left-over from when Tobias was testing things. Note also that this bakes in a fixed serialized data version number, which felt dangerous. Due to these two reasons, I think I would prefer to remove this default.

.jenkins/actions/run_standalone.sh
rheacangeo: looks like nothing that activates the venv does a deactivate as a final step, so it probably still is activated at this point. I guess it doesn't matter.

examples/standalone/benchmarks/process_profiling.sh
rheacangeo: Sure, it just means we need to assume the virutalenv to be used is always rootdir/venv, while there could be use cases to use others. but no need to change, someone can if they actually need that use case

.jenkins/actions/run_standalone.sh
ofuhrer: It is built and activated inside `run_on_daint.sh`. But once you exit a shell script venv's are deactivated. This is really just to make sure it is removed.

examples/standalone/benchmarks/run_on_daint.sh
rheacangeo: I more mean the renaming of lower case nthreads to allcaps NTHREADS? I saw the jenkins integration is for  future PR, but why use a different name? not a huge thing, just curious

rheacangeo (approved):

examples/standalone/benchmarks/run_on_daint.sh
rheacangeo: oh, 'to be more visible', you already said, got it :)

.jenkins/actions/run_standalone.sh
rheacangeo: ah didn't realize that, makes sense

rheacangeo:
Oh, you need to 'make lint' before it'll go through the tests

.jenkins/actions/run_standalone.sh
```diff
@@ -0,0 +1,112 @@
...
+# utility function for error handling
+exitError()
+{
+    echo "ERROR $1: $3" 1>&2
+    echo "ERROR     LOCATION=$0" 1>&2
+    echo "ERROR     LINE=$2" 1>&2
+    exit $1
+}
+
+# check arguments
```
twicki, line 18: do we have a convention for describing the arguments in a starting comment block?

It usually helps me personally (how often did I have to check if ranks or time steps was the first argument...)

twicki, line 18: do we have a convention for describing the arguments in a starting comment block?

It usually helps me personally (how often did I have to check if ranks or time steps was the first argument...)

.jenkins/actions/run_standalone.sh
```diff
@@ -0,0 +1,112 @@
...
+echo "Save GT4Py caches:    ${SAVE_CACHE}"
+echo "Save timings:         ${SAVE_TIMINGS}"
+echo "Root directory:       ${ROOT_DIR}"
+echo "Experiment:           ${experiment}"
+echo "Backend:              ${backend}"
+echo "Data version:         ${DATA_VERSION}"
+echo "Timesteps:            ${TIMESTEPS}"
+echo "Ranks:                ${RANKS}"
+echo "Benchmark directory:  ${BENCHMARK_DIR}"
+echo "Data directory:       ${DATA_DIR}"
```
twicki, line 71: it is definitely not bad like this but data_dir is derived from experiment and data version. Do we want all three (because the data dir can't be overridden)?

examples/standalone/benchmarks/run_on_daint.sh
```diff
@@ -139,42 +131,57 @@ echo "submitting script to do compilation"
...
+wait
+echo "DONE WAITING ${status1} ${status2}"
+if [ $status1 -ne 0 -o $status2 -ne 0 ] ; then
+    cleanupFailedJob "${res}"
+    echo "ERROR: compilation step failed"
+    exit 1
+else
+    echo "compilation step finished"
+fi
+mv -f slurm-*.out compile.daint.out
```
twicki, line 157: should we just sed the output to this already?

examples/standalone/benchmarks/run_on_daint.sh
```diff
@@ -139,42 +131,57 @@ echo "submitting script to do compilation"
...
+if [ $status1 -ne 0 -o $status2 -ne 0 ] ; then
+    cleanupFailedJob "${res}"
+    echo "ERROR: performance run not sucessful"
+    exit 1
+else
+    echo "performance run sucessful"
 fi
-
-echo "performance run sucessful"
+mv -f slurm-*.out run.daint.out
```
twicki, line 187: newline required for linting

twicki (commented):
some minor questions

examples/standalone/benchmarks/run_on_daint.sh
ofuhrer: Thx.

examples/standalone/benchmarks/run_on_daint.sh
ofuhrer: Agreed. Done.

examples/standalone/benchmarks/process_profiling.sh
ofuhrer: Changed.

examples/standalone/benchmarks/run_on_daint.sh
```diff
@@ -128,60 +129,56 @@ fi
...
-sed -i s/\<CPUSPERTASK\>/$NTHREADS/g run.daint.slurm
-sed -i s/--output=\<OUTFILE\>/--hint=nomultithread/g run.daint.slurm
-sed -i s/00:45:00/00:40:00/g run.daint.slurm
-sed -i s/cscsci/normal/g run.daint.slurm
-sed -i s/\<G2G\>//g run.daint.slurm
+sed -i "s/<NAME>/standalone/g" run.daint.slurm
+sed -i "s/<NTASKS>/$ranks/g" run.daint.slurm
+sed -i "s/<NTASKSPERNODE>/1/g" run.daint.slurm
+sed -i "s/<CPUSPERTASK>/$NTHREADS/g" run.daint.slurm
+sed -i "s/<OUTFILE>/run.daint.out\n--hint=nomultithread/g" run.daint.slurm
```
twicki, line 164: this needs an `#SBATCH` the  `nomultithread` now, right?

twicki (changes requested):

twicki (approved):

examples/standalone/benchmarks/run_on_daint.sh
ofuhrer: Thanks for the good catch.

PR #305, opened 2021-03-12
Refactor d2a2c into single stencil

jdahm:
## Purpose

Refactor d2a2c into a single stencil using regions.

## Code changes:

- `d2a2c_vect` is reduced into a single stencil.


oelbert:
launch jenkins

PR #311, opened 2021-03-16
shorten the compile time jenkins job

rheacangeo:
so it has a better chance of being let on a node


rheacangeo:
this was opened out of desperation as piz daint was shutting down nodes for maintenance and I didn't realize it, thought we were competing for compute

PR #315, opened 2021-03-17
Fvtp2d update

gmao-ckung:
## Purpose

This PR combines the `transport_flux` calls for `x` and `y` into a single call using horizontal regions.

## Code changes:

`fvtp2d.py` : Replace the back-to-back calls to `transport_flux` in `compute_no_sg` into a single call called `transport_flux_xy`.  `transport_flux` is changed to a function call.


rheacangeo (approved):
Looks great! Ah very cool tactic using '-1' in the region rather than the local indices names, makes a lot of sense in this case!

PR #316, opened 2021-03-17
Neg_adj3 update

gmao-ckung:
## Purpose

This PR simply combines the 5 stencil calls within `compute` into a single stencil call.

## Code changes:

`neg_adj3.py` : Combines `fix_neg_water`, `fillq`, `fix_water_vapor_down`, and `fix_neg_cloud` stencils into a single stencil called `fix_neg_values`.


oelbert (commented):
This looks like a great start, but it would be awesome if you could split out the different processes into functions, like a fillq function and a fix_neg_cloud function, etc.

rheacangeo:
neg_adj has now been refactored to be an object. this proposal is still relevant, but would need to be updated if we were to revive it.

gmao-ckung:
Closing due to OOing affecting the code from this PR.

PR #321, opened 2021-03-19
updates the gt4py directory when the tag does not match

rheacangeo:
## Purpose

 The gt4py directory gets cloned to the local directory, and the GT4PY_VERSION defined in Makefile.image_names is checked out. For development, when this tag changes, we want it to also trigger updating the gt4py directory.  The main problem with this is that sometimes there are multiple gt4py tags on the same commit, in which case the version will always not match and every time the code is built it will be slow as it reclones the directory and rebuilds the image using that directory... 

## Infrastructure changes:

- add a check to the get_gt4py endpoint in docker/Makefile


rheacangeo:
Closing as work duplicated in #335

PR #330, opened 2021-03-24
Ignore caches if GT4Py Version that generated the cache does not match

twicki:
## Purpose

If only the gt4py version changes, our current jenkins infrastructure copies in all the previously generated files and then runs with `rebuild=False`. This has the effect that changes to gt4py are never picked up. This PR changes it such that the gt4py version is written to the cache and the cache is only copied in if the versions match.

## Infrastructure changes:

- All the jenkins plans started from `gt4py-develop` are rebuilding everything
- If the `GT4PY_VERSION` changes, fv3core tests will rebuild everything

### DIFF
--- a/.jenkins/actions/run_cached_fv_dynamics.sh
+++ b/.jenkins/actions/run_cached_fv_dynamics.sh
@@ -8,7 +8,10 @@ export TEST_ARGS="-v -s -rsx --backend=${BACKEND} --which_modules=FVDynamics"
 make get_test_data
 
 if [ ! -d $(pwd)/.gt_cache_000000 ]; then
-    cp -r /scratch/snx3000/olifu/jenkins/scratch/store_gt_caches/$EXPNAME/$BACKEND/.gt_cache_0000* .
-    find . -name m_\*.py -exec sed -i "s|\/scratch\/snx3000\/olifu\/jenkins_submit\/workspace\/fv3core-cache-setup\/backend\/$BACKEND\/experiment\/$EXPNAME\/slave\/daint_submit|$(pwd)|g" {} +
+    version=`cat /scratch/snx3000/olifu/jenkins/scratch/store_gt_caches/$EXPNAME/$BACKEND/version.txt`
+    if [ "$version" = "$GT4PY_VERSION" ]; then
+        cp -r /scratch/snx3000/olifu/jenkins/scratch/store_gt_caches/$EXPNAME/$BACKEND/.gt_cache_0000* .
+        find . -name m_\*.py -exec sed -i "s|\/scratch\/snx3000\/olifu\/jenkins_submit\/workspace\/fv3core-cache-setup\/backend\/$BACKEND\/experiment\/$EXPNAME\/slave\/daint_submit|$(pwd)|g" {} +
+    fi    
 fi
 make tests_venv_mpi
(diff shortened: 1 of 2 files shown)


rheacangeo:
you need to 'make lint'

.jenkins/actions/run_cached_fv_dynamics.sh
```diff
@@ -8,7 +8,10 @@ export TEST_ARGS="-v -s -rsx --backend=${BACKEND} --which_modules=FVDynamics"
 make get_test_data
 
 if [ ! -d $(pwd)/.gt_cache_000000 ]; then
-    cp -r /scratch/snx3000/olifu/jenkins/scratch/store_gt_caches/$EXPNAME/$BACKEND/.gt_cache_0000* .
-    find . -name m_\*.py -exec sed -i "s|\/scratch\/snx3000\/olifu\/jenkins_submit\/workspace\/fv3core-cache-setup\/backend\/$BACKEND\/experiment\/$EXPNAME\/slave\/daint_submit|$(pwd)|g" {} +
+    version=`cat /scratch/snx3000/olifu/jenkins/scratch/store_gt_caches/$EXPNAME/$BACKEND/version.txt`
+    if [ "$version" = "$GT4PY_VERSION" ]; then
```
ofuhrer, line 12: From where is `$GT4PY_VERSION` set here?

ofuhrer, line 12: Just learned that `=` is equivalent to `==` in bash. This is fine, but I think we mostly use `==` in our scripts.

ofuhrer (approved):
This looks good to me. Two questions:
- How is "All the jenkins plans started from gt4py-develop are rebuilding everything" implemented?
- Would it make sense to manually add a `version.txt` to the caches right now?

.jenkins/actions/run_cached_fv_dynamics.sh
twicki: from here:
https://github.com/VulcanClimateModeling/fv3core/blob/415cfb9356a6c9cea697c6cea70989865eb2ca3b/.jenkins/jenkins.sh#L131

twicki: bash doing bashy thing. I'm fine following any style recommendations.

twicki:
Some replies to @ofuhrer questions:
> how is "All the jenkins plans started from gt4py-develop are rebuilding everything" implemented?

The plan that launches fv3core regression from a new version of  gt4py sets the variable to the hash (which never matches the version)
Jenkins config

> Would it make sense to manually add a `version.txt` to the caches right now?

Yes it would! I just assumed we are fiddling with the cache now so it might get deleted again until we merge this (which we can only do once we're back to green) - this is why i put it off for now

.jenkins/actions/run_cached_fv_dynamics.sh
ofuhrer: Maybe comment it?

ofuhrer:
Thanks. All good from me.

rheacangeo:
if we do #335 , we should use that here. otherwise, should change the check for GT4PY_VERSION to be like the other ones in master now

rheacangeo:
These changes are merged into #335, so we can close this one

twicki:
launch jenkins

twicki:
closed in favor of #335

PR #334, opened 2021-03-25
fix the determination of the GT4PY_VERSION for run_on_daint

rheacangeo:
## Purpose
We made a mistake, as mistakes happen when you duplicate code and change it only in one spot. This should stop the inconsistency we are seeing between run_on_daint and running the tests

### DIFF
--- a/examples/standalone/benchmarks/run_on_daint.sh
+++ b/examples/standalone/benchmarks/run_on_daint.sh
@@ -75,7 +75,7 @@ git submodule update --init external/fv3gfs-util external/daint_venv
 
 # set GT4PY version
 cd $ROOT_DIR
-export GT4PY_VERSION=`grep "GT4PY_VERSION ?=" docker/Makefile.image_names | cut -d '=' -f 2`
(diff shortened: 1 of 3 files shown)


eddie-c-davis (approved):
Great catch!

examples/standalone/benchmarks/run_on_daint.sh
```diff
@@ -75,7 +75,7 @@ git submodule update --init external/fv3gfs-util external/daint_venv
 
 # set GT4PY version
 cd $ROOT_DIR
-export GT4PY_VERSION=`grep "GT4PY_VERSION=" docker/Makefile.image_names | cut -d '=' -f 2`
+export GT4PY_VERSION=`grep "GT4PY_VERSION ?=" docker/Makefile.image_names | cut -d '=' -f 2`
```
ofuhrer, line 78: `export GT4PY_VERSION=`grep "GT4PY_VERSION" docker/Makefile.image_names  | tr -d ' ' | cut -d '=' -f 2`

ofuhrer (approved):

examples/standalone/benchmarks/run_on_daint.sh
rheacangeo: thanks, used that with a -m 1

PR #335, opened 2021-03-25
move the specification of the GT4PY_VERSION 

rheacangeo:
## Purpose

Move the specification of the GT4PY_VERSION  to a file called GT4PY_VERSION.txt rather than keeping it in the docker/Makefile.image_names, which is an awkward place to look for it.
Merged with #330 to not copy caches if the cache was built using a different gt4py version than being run

If only the gt4py version changes, our current jenkins infrastructure copies in all the previously generated files and then runs with rebuild=False. This has the effect that changes to gt4py are never picked up. This PR changes it such that the gt4py version is written to the cache and the cache is only copied in if the versions match.

## Infrastructure changes:

- add GT4PY_VERSION.txt to the top level that contains the tag we pin this fv3core commit to
- All the jenkins plans started from gt4py-develop are rebuilding everything
 - If the GT4PY_VERSION changes, fv3core tests will rebuild everything


ofuhrer (approved):
This looks good to me. Thanks!

@eddie-c-davis How do we get this into `daint_venv`?
@twicki This probably has impact on your PR https://github.com/VulcanClimateModeling/fv3core/pull/330

rheacangeo:
launch jenkins

rheacangeo:
@ofuhrer Tobias and I worked together to merge #330 here so we could do these small changes alltogether, fyi. The jenkins.sh and run_on_daint (and now standalone) scripts will still set the environment variable that daint_venv uses, but if preferred we can do something different.

PR #331, opened 2021-03-24
Revert "Lower dimensional grid and input variables (#269)"

rheacangeo:
## Purpose
This reverts commit 415cfb9356a6c9cea697c6cea70989865eb2ca3b that introduced using lower dimensional storages throughout the dycore. (#269)

We really don't want to do this, but are blocked on merging other code after this one went in as gt backend tests no longer pass (though they did for this PR), and haven't been able to get gt backend timing updates on the impact of that branch. Opening a PR  to see if reverting works to unblock other code in flight.

PR #336, opened 2021-03-25
Revert "Revert "Lower dimensional grid  (#269)" (#331)"

rheacangeo:
## Purpose

Bring back the lower dimensional storages commit, which was reverted when we did not know why it didn't work.


eddie-c-davis (approved):
Looks good!

PR #340, opened 2021-03-26
YTP_V and XTP_U are classes

rheacangeo:
## Purpose

In our aim to separate compilation from runtime via OOP, YTP_V and XTP_U are now classes and d_sw initializes (memoized) and calls these objects. When D_SW is an object, the initialization will no longer happen in runtime. 
Jira card https://vulcan.atlassian.net/browse/DSL-481

## Code changes:

- xtp_u and ytp_v are both changed to be classes in our OOP framework
- d_sw calls these the new way
- xtp_u and ytp_v test code is updated

## Checklist

### DIFF
--- a/fv3core/stencils/d_sw.py
+++ b/fv3core/stencils/d_sw.py
@@ -828,8 +828,8 @@ def d_sw(
         origin=grid().compute_origin(),
         domain=grid().domain_shape_compute(add=(1, 1, 0)),
     )
-
-    ytp_v.compute(vb, v, ub)
+    ytp_v_obj = ytp_v.YTP_V(spec.namelist)
+    ytp_v_obj(vb, v, ub)
 
     mult_ubke(
         vb,
@@ -845,8 +845,8 @@ def d_sw(
         origin=grid().compute_origin(),
         domain=grid().domain_shape_compute(add=(1, 1, 0)),
     )
-
-    xtp_u.compute(ub, u, vb)
+    xtp_u_obj = xtp_u.XTP_U(spec.namelist)
+    xtp_u_obj(ub, u, vb)
 
     ke_horizontal_vorticity(
         ke,
--- a/fv3core/stencils/d_sw.py
+++ b/fv3core/stencils/d_sw.py
@@ -17,12 +17,12 @@ import fv3core.stencils.divergence_damping as divdamp
 import fv3core.stencils.flux_capacitor as fluxcap
 import fv3core.stencils.fvtp2d as fvtp2d
 import fv3core.stencils.fxadv as fxadv
-import fv3core.stencils.xtp_u as xtp_u
-import fv3core.stencils.ytp_v as ytp_v
 import fv3core.utils.corners as corners
 import fv3core.utils.global_constants as constants
 import fv3core.utils.gt4py_utils as utils
 from fv3core.decorators import gtstencil
+from fv3core.stencils.xtp_u import XTP_U
+from fv3core.stencils.ytp_v import YTP_V
 from fv3core.utils.typing import FloatField, FloatFieldIJ
 
 
@@ -828,7 +828,7 @@ def d_sw(
         origin=grid().compute_origin(),
         domain=grid().domain_shape_compute(add=(1, 1, 0)),
     )
-    ytp_v_obj = ytp_v.YTP_V(spec.namelist)
+    ytp_v_obj = utils.cached_stencil_class(YTP_V)(spec.namelist, cache_key="ytp_v")
     ytp_v_obj(vb, v, ub)
 
     mult_ubke(
@@ -845,7 +845,7 @@ def d_sw(
         origin=grid().compute_origin(),
         domain=grid().domain_shape_compute(add=(1, 1, 0)),
     )
-    xtp_u_obj = xtp_u.XTP_U(spec.namelist)
+    xtp_u_obj = utils.cached_stencil_class(XTP_U)(spec.namelist, cache_key="xtp_u")
     xtp_u_obj(ub, u, vb)
 
     ke_horizontal_vorticity(
--- a/fv3core/stencils/xtp_u.py
+++ b/fv3core/stencils/xtp_u.py
@@ -129,14 +129,11 @@ class XTP_U:
             raise NotImplementedError(
                 "Currently xtp_v is only supported for hord_mt == 5,6,7,8"
             )
-        grid = spec.grid
-        self.origin = grid.compute_origin()
-        self.domain = grid.domain_shape_compute(add=(1, 1, 0))
-        ax_offsets = axis_offsets(spec.grid, self.origin, self.domain)
+        self.grid = spec.grid
+        self.origin = self.grid.compute_origin()
+        self.domain = self.grid.domain_shape_compute(add=(1, 1, 0))
+        ax_offsets = axis_offsets(self.grid, self.origin, self.domain)
         assert namelist.grid_type < 3
-        self.dxa = grid.dxa
-        self.dx = grid.dx
-        self.rdx = grid.rdx
         self.stencil = gtscript.stencil(
             definition=_compute_stencil,
             externals={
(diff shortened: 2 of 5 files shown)


fv3core/stencils/d_sw.py
```diff
@@ -828,8 +828,8 @@ def d_sw(
         origin=grid().compute_origin(),
         domain=grid().domain_shape_compute(add=(1, 1, 0)),
     )
-
-    ytp_v.compute(vb, v, ub)
+    ytp_v_obj = ytp_v.YTP_V(spec.namelist)
```
twicki, line 831: my understanding here is that `d_sw()` is called in it's `compute` function which is part of the main loop.
Why do we initialize the object in here? This leads to significant overhead as compared to initializing it once with the namelist and then continuously "only" calling it?

twicki, line 831: I think this naming is somewhat non-intuitive:
- Why do we import `import fv3core.stencils.ytp_u as ytp_u`and then only use its single call?  Would it not be more readable to `from  fv3core.stencils.ytp_u import YTP_U`?
- the fact that this is an object is an implementation detail, right? does this really have to be reflected in the name of `ytp_v_obj`?

fv3core/stencils/xtp_u.py
```diff
@@ -123,37 +122,49 @@ def _compute_stencil(
...
+        if iord not in (5, 6, 7, 8):
+            raise NotImplementedError(
+                "Currently xtp_v is only supported for hord_mt == 5,6,7,8"
+            )
+        grid = spec.grid
+        self.origin = grid.compute_origin()
+        self.domain = grid.domain_shape_compute(add=(1, 1, 0))
+        ax_offsets = axis_offsets(spec.grid, self.origin, self.domain)
+        assert namelist.grid_type < 3
+        self.dxa = grid.dxa
```
twicki, line 137: are we ever in a state where we have multiple of these objects with multiple namelists floating around? I am curious behind the reason to chose instance variables over class variables here?

fv3core/stencils/ytp_v.py
```diff
@@ -56,7 +57,7 @@ def _compute_stencil(
     dya: FloatFieldIJ,
     rdy: FloatFieldIJ,
 ):
-    from __externals__ import i_end, i_start, j_end, j_start, jord, namelist
+    from __externals__ import i_end, i_start, j_end, j_start, jord
```
twicki, line 60: i am still very bugged about having end before start. Why do we keep leaving this in?

tests/translate/translate_xtp_u.py
```diff
@@ -1,15 +1,17 @@
...
 
 class TranslateXTP_U(TranslateYTP_V):
     def __init__(self, grid):
         super().__init__(grid)
+        self.in_vars["data_vars"]["u"] = {}
         self.in_vars["data_vars"]["c"]["serialname"] = "ub"
         self.in_vars["data_vars"]["flux"]["serialname"] = "vb"
 
-    def compute(self, inputs):
-        self.make_storage_data_input_vars(inputs)
```
twicki, line 13: am i correct that this is wrapped by compute form storage now?

twicki (changes requested):

fv3core/stencils/d_sw.py
rheacangeo: We do this as an intermediate step, and when D_SW is an object, you are right, this part will happen in its __init__ routine. We've added the ability to memoize class instantiation -- I meant to add it here thanks for the reminder! (first version you just had to add the decorator to the class for it to work, now we decide to do it directly)

rheacangeo: sure I can do a different import to make it less repetitive. We've been adding 'obj' in the other classes as a convention because we have a lot of 'self.<>' variables. It is still open though for discussion and we can definitely change this convention. I'd rather do that for all the classes we added so far in another PR, if that's ok?

fv3core/stencils/xtp_u.py
rheacangeo: We currently don't have a use case where we need multiple namelists in the same thread, I think having global state for things that are essentially global for a run is intuitive and makes a lot of sense. But it does reduce the future potential flexibility to use the model and pieces of the model in isolated ways or for experiments where multiple namelists are used. I think when Jeremy presents the objects to us again we can definitely discuss more.

tests/translate/translate_xtp_u.py
rheacangeo: yeah we just have another parent method we can use to reduce duplication

fv3core/stencils/ytp_v.py
rheacangeo: it keeps sneaking back in with refactors and merges, good eye, I can update

rheacangeo: the linter won't let me! :)

fv3core/stencils/d_sw.py
twicki: and where would the DSW object be instantiated?

twicki: sure, just wanted to point that out

fv3core/stencils/xtp_u.py
```diff
@@ -123,37 +122,46 @@ def _compute_stencil(
...
-        domain=grid.domain_shape_compute(add=(1, 1, 0)),
-    )
+    def __call__(self, c: FloatField, u: FloatField, flux: FloatField):
+        """
+        Compute flux of kinetic energy in x-dir.
+
+        Args:
+            c (in): Courant number in flux form
+            u (in): x-dir wind on D-grid
+           flux (out): Flux of kinetic energy
```
mcgibbon, line 156: So clean!

Misaligned tab:
```suggestion
        Args:
            c (in): Courant number in flux form
            u (in): x-dir wind on D-grid
            flux (out): Flux of kinetic energy
```

fv3core/stencils/xtp_u.py
```diff
@@ -123,37 +122,46 @@ def _compute_stencil(
...
-        raise NotImplementedError(
-            "Currently xtp_v is only supported for hord_mt == 5,6,7,8"
+class XTP_U:
+    def __init__(self, namelist):
+        iord = spec.namelist.hord_mt
+        if iord not in (5, 6, 7, 8):
+            raise NotImplementedError(
+                "Currently xtp_v is only supported for hord_mt == 5,6,7,8"
+            )
+        self.grid = spec.grid
```
mcgibbon, line 132: Not necessary here or for the other first-round object-oriented PRs, but in the long term when we work to pass only the required namelist flags, I'd also like to pass only the required grid variables. This form makes it way easier to do that, since we can grep for `self.grid`!

fv3core/stencils/ytp_v.py
```diff
@@ -121,37 +120,47 @@ def _compute_stencil(
         flux = _get_flux(v, courant, rdy, bl, br)
 
 
-def compute(c: FloatField, v: FloatField, flux: FloatField):
-    """
-    Compute flux of kinetic energy in y-dir.
+class YTP_V:
```
mcgibbon, line 123: I'm good with capitalizing the existing names in this first iteration to keep the naming straightforward. When we make these class names readable, we should use CamelCase convention without underscores to keep in line with pep 8.

mcgibbon (approved):
Nice!

twicki (approved):
lgtm

fv3core/stencils/d_sw.py
rheacangeo: in the DynCore object when it is made, until it is made could be memoized init in the dyn_core compute method

rheacangeo: We are going to remove 'obj' from our naming conventions, just a heads up!

fv3core/stencils/xtp_u.py
rheacangeo: Yeah we just discussed for awhile, I still think it's nice on the user to have the available objects be categorized very clearly where they are used, but you and Oliver pointed out we should need an entire grid object for every module, and fortran actually does also explicitly call out which grid variables are used in each modules. So can move that direction.

PR #263, opened 2021-02-24
Reapply pr241 FxAdv single stencil

rheacangeo:
## Purpose

Reapply PR 241

## Code changes:

Please see:
- #241

PR #344, opened 2021-03-28
Fxadv does not include runtime decorator calls

rheacangeo:
## Purpose

PR #263 makes Fxadv a single stencil, but does not work yet with the gt backends. gt4py-develop: v20 *does* work for that one, but D_SW calling it crashes. Maybe that will be resolved very quickly, but this is an easy intermediate step. Here we introduce code that allows us to get rid of the runtime gtstencil decoration in the model that is obscuring the profile, so we can make decision on the most impactful area of the code to focus on. I am not updating documentation in this one, because we will come back to it very soon. 

## Code changes:

- Built from #263, FxAdv split into enough stencils to pass gt backend tests, with corners in a single stencil. runtime gtstencil decorator calls removed. 

## Checklist


fv3core/stencils/fxadv.py
```diff
@@ -1,1026 +1,861 @@
...
-
+        ut = main_ut(uc, vc, cosa_u, rsin_u, ut)
+        ut = ut_y_edge(uc, sin_sg1, sin_sg3, ut, dt)
+        vt = main_vt(uc, vc, cosa_v, rsin_v, vt)
+        vt = vt_y_edge(vc, cosa_v, ut, vt)
+        vt = vt_x_edge(vc, sin_sg2, sin_sg4, vt, dt)
+        ut = ut_x_edge(uc, cosa_u, vt, ut)
+        ut = ut_corners(uc, vc, cosa_u, cosa_v, ut, vt)
+        vt = vt_corners(uc, vc, cosa_u, cosa_v, ut, vt)
+"""
```
oelbert, line 356: Why is this code block-commented?

oelbert (approved):
👍 👍 👍

fv3core/stencils/fxadv.py
rheacangeo: This is what the single stencil version is, all the things that are stencils now can be functions with the numpy backend, so when the gtbackends support the regions craziness we can use this

fv3core/stencils/fxadv.py
```diff
@@ -1,1026 +1,855 @@
...
+    with computation(PARALLEL), interval(...):
+        utmp = ut
+        with horizontal(region[local_is - 1 : local_ie + 3, :]):
+            ut = (
+                uc - 0.25 * cosa_u * (vc[-1, 0, 0] + vc + vc[-1, 1, 0] + vc[0, 1, 0])
+            ) * rsin_u
+        with horizontal(
+            region[:, j_start - 1 : j_start + 1], region[:, j_end : j_end + 2]
+        ):
+            ut = utmp
```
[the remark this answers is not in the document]
rheacangeo: added, thanks for the suggestion! we should revisit this

PR #351, opened 2021-03-31
remove more k loops from d_sw

rheacangeo:
## Purpose

Another step to simplifying the loops over k. Turn the column namelist into column storages instead of lists to make it possible to use them inside of stencils. This adds a lot of temporary storage column variables creation that will be good to have in compile-time when D_SW is a class. But caching should help. Left the k_bounds loops that were directly outside of other code that needs to be refactored and is in flight, e.g. delnflux, divergence_damping. Also deleted some leftover files that are unused.  The stencils involving heat_source needed some attention. Moving the conditional over a column field inside a stencil made the code that had read offset after write not work (seems like maybe it would have never worked?)

## Code changes:
- d_sw's column_namelist now has column gt4py storages instead of lists so they can be used inside of stencils
- removed some stencils called in a loop over k_bounds and merged some stencil code
- move the TranslateHeatDiss into the file for TranslateD_SW, deleted the heat_diss file
-deleted several translate classes around substeps in FVDynamics that are not tested because they are unnecessary

## Checklist


fv3core/stencils/d_sw.py
```diff
@@ -431,21 +450,34 @@ def get_column_namelist():
...
+        "nord_v",
+        "nord_w",
+        "nord_t",
+        "damp_vt",
+        "damp_w",
+        "damp_t",
+        "d2_divg",
+    ]
     col = {}
     num_k = len(k_bounds())
```
eddie-c-davis, line 463: Is `num_k` any longer needed?

fv3core/stencils/d_sw.py
```diff
@@ -570,33 +602,35 @@ def damp_vertical_wind(w, heat_s, diss_e, dt, column_namelist, kstart, nk):
...
-            w,
-            grid().rarea,
-            heat_s,
-            diss_e,
-            dw,
-            dd8,
-            origin=(grid().is_, grid().js, kstart),
-            domain=(grid().nic, grid().njc, nk),
-        )
+    for kstart, nk in k_bounds():
```
eddie-c-davis, line 605: Nice simplification.

eddie-c-davis (approved):
Excellent work! Especially like the way you were able to lower the k-loops but also move some 
 logic into stencils using k-fields.

eddie-c-davis (approved):
Still LGTM!

PR #355, opened 2021-04-01
Add recent absolute runtime plot

ofuhrer:
## Purpose

Adaption of plots to add a plot of absolute runtime for the past 7 days as well. Should help us in deciding whether we can increase the number of timesteps.


twicki (approved):
lgtm - do we need the full data still? or can we just replace that?

PR #359, opened 2021-04-04
refactor a2b_ord4 with #235 reapplication + OOP + v17

rheacangeo:
## Purpose

refactor a2b_ord4 -- partially reapplying #235 with modifications to work with the v17 gt4py:develop tag where regions are reimplemented to work better for gt backends, but do not allow temporaries(and thus gtscript functions) to be assigned within a region. 
Also converting to the OOP design at the same time because we are doing a few things that slow down the code that OOPing solves. 
A small workaround was needed to deal with the edge_w and edge_e grid variable, which are 1D arrays in the J dimension. They are defined to be IJ with a singleton I, which works in regular stencils, but within regions triggers and error. Here we duplicate the dat in the I dimension for a temporary workaround.  
For gtx86 to pass, temporary variables were made to be temporary storages and passed in, and the stencil was broken into multiple parts. It should be easy to push these together and remove the temporaries once the next regions update is tagged in gt4py.
Because we will revisit this soon and are keenly trying to remove python code from the repo, we have chosen not to update names and docstrings for this PR. 

## Code changes:

- a2b_ord4 is refactored
- python math functions used by a2b_ord4 are stencils and thus removed from gt4py_utils

## Checklist


rheacangeo:
Closing because out of date and regions working better with gt backends unlikely to change anytime soon

PR #360, opened 2021-04-05
enable mypy in pre-commit, fix simplest typing errors

mcgibbon:
## Purpose

While we have type hints in our repo, they are currently untested and in some cases misleading. This PR enables static type checking through mypy to ensure they reflect the code written, and to provide more robustness in the code (by testing typing, which we don't currently do in our unit tests). This can give faster turnaround on feedback when refactoring the codebase than the unit tests.

In this commit type checking is disabled for many files in the repository. They can be progressively enabled by deleting the corresponding ignore line from `pre-commit-config.yaml`.

Gt4py breaks static typing when for some operations, requiring an ignore comment. For example:
```
with interval(0, 1):
    peg = ptop
with interval(1, None):
    peg = peg[0, 0, -1] + 1.  # type: ignore[index]
```

This shows up in only a few places in our repo.

## Code changes:

- Fixed the most straightforward typing errors, particularly for files with very few errors
- `gtstencil()` now only operates as a function which returns a decorator. It was being used this way in almost all cases, and the remaining cases could be modified to use it this way by only adding parentheses. This greatly simplifies type checking for this routine, and makes its usage uniform across the repository.

## Requirements changes:

- Added mypy to `pre-commit-config.yaml`

## Infrastructure changes:

- Added mypy type checking for certain files in the repo

## Checklist


mcgibbon:
launch jenkins

.pre-commit-config.yaml
```diff
@@ -9,6 +9,30 @@ repos:
...
+        fv3core/utils/gt4py_utils.py |
+        fv3core/decorators.py |
+        fv3core/stencils/fillz.py |
+        fv3core/stencils/moist_cv.py |
+        fv3core/stencils/corners.py |
+        fv3core/stencils/rayleigh_super.py |
+        fv3core/stencils/map_single.py |
+        fv3core/stencils/remap_profile.py |
+        fv3core/stencils/fv_subgridz.py |
+        fv3core/testing/parallel_translate.py
```
rheacangeo, line 33: way less than I expected!

CONTRIBUTING.md
```diff
@@ -124,8 +124,6 @@ def pt_adjust(pkz:sd, dp1: sd, q_con: sd, pt: sd):
     with computation(PARALLEL), interval(...):
 ```
 
-Note that `fv3core.gtstencil` can be manually called on an undecorated stencil, but this is currently in general discouraged.
```
rheacangeo, line 127: good call, we don't want to even mention this option any more

examples/wrapped/plotscripts/raw_out_testplots.py
```diff
@@ -51,7 +52,7 @@
 surface_pressure_plots = []
 surface_temperature_plots = []
 
-relative_errors = {}
+relative_errors: Dict[str, np.ndarray] = {}
```
rheacangeo, line 55: was this type-hint required? I thought adding type-hints was still optional, or did you just add it because you were looking at this script and added it?

fv3core/utils/corners.py
```diff
@@ -212,17 +212,15 @@ def definition(q: FloatField):
         func = (
             fill_corners_2cells_mult_x if num_fill == 2 else fill_corners_3cells_mult_x
         )
-        stencil = gtstencil(
-            definition=definition,
-            externals={"func": func},
+        stencil = gtstencil(externals={"func": func})(
+            definition,
```
rheacangeo, line 216: was this something caught by mypy? it seems like it was valid, just a different convention?

fv3core/utils/corners.py
```diff
@@ -212,17 +212,15 @@ def definition(q: FloatField):
...
     elif direction == "y":
         func = (
             fill_corners_2cells_mult_y if num_fill == 2 else fill_corners_3cells_mult_y
         )
-        stencil = gtstencil(
-            definition=definition,
-            externals={"func": func},
+        stencil = gtstencil(externals={"func": func})(
+            definition,
         )
```
rheacangeo, line 224: same question

setup.py
```diff
@@ -37,7 +35,7 @@
     keywords="fv3core",
     name="fv3core",
     packages=find_packages(include=["fv3core", "fv3core.*"]),
-    setup_requires=setup_requirements,
+    setup_requires=[],
```
rheacangeo, line 38: what is this change about?

rheacangeo (commented):
Excellent work. I have a few questions on things, but no major blockers, looks good!

.pre-commit-config.yaml
mcgibbon: Me too! This was a lot longer, but the majority of the files only had the issue that `gtstencil` could not be statically typed.

CONTRIBUTING.md
mcgibbon: The option has been removed by this PR, so I'd hope so!

examples/wrapped/plotscripts/raw_out_testplots.py
mcgibbon: Type hints are required by mypy for dictionaries whose initialization does not indicate the types. If this were `relative_errors = {"foo": np.ndarray([1.0])}` mypy wouldn't ask for a type hint. This is only true about a couple built-in mutable classes, like `list` and `dict`.

If you don't type hint something, mypy secretly type hints it by seeing how you initialized it.

fv3core/utils/corners.py
mcgibbon: This was because I changed gtstencil to have a consistent static typing scheme, so it has to be used in the same way everywhere. This usage is the same as:

```
@gtstencil(externals={"func": func})
def defintiion():
    ...
```

which as you know is how we use as a decorator.

Fixing this made most files pass mypy.

fv3core/utils/corners.py
mcgibbon: See above.

setup.py
mcgibbon: Same as when `my_dict = {}` required a type hint, mypy also wanted a type hint for line 14. Because we don't actually use setup_requirements, I did this change instead to make it clear when reading the `setup` call that we give no setup requirements.

examples/wrapped/plotscripts/raw_out_testplots.py
rheacangeo: seems overly restrictive to me, but we can follow it given we are mostly trying to push things into gt4py

mcgibbon: In practice this comes up rarely enough that I forget it's required. I think I've had to type hint things like this fewer than 10 times, ever.

rheacangeo (approved):
re-approving :)

PR #361, opened 2021-04-05
Switch assert __INLINED to external_assert

jdahm:
## Purpose

Switch to function-based `external_assert` from the keyword `assert`.

## Code changes:

- Switch to function-based `external_assert` from the keyword `assert`.

## Requirements changes:

- Requires new gt4py version, which in turn depends on this.


ofuhrer (approved):
I really like this. Thanks for getting this in so blazingly fast. Two questions:
- What's the "reason" behind *external* in `external_assert()`?
- Why did you decide to remove the additional string comment?

jdahm:
Thanks.

> What's the "reason" behind external in external_assert()?

I didn't want to call it `assert` since it might confuse with the run-time keyword `assert`.

> Why did you decide to remove the additional string comment?

That comment box never worked very well. It only worked with string literals. It prints the source code line with the assertion anyway, so it could be put in a code comment above the assertion.

ofuhrer:
> > What's the "reason" behind external in external_assert()?
> 
> I didn't want to call it `assert` since it might confuse with the run-time keyword `assert`.

That was clear. I was wondering why "external". External to what?

> That comment box never worked very well. It only worked with string literals. It prints the source code line with the assertion anyway, so it could be put in a code comment above the assertion.

👍

jdahm:
> That was clear. I was wondering why "external". External to what?

Ah, yes: because the compile-time constants are called "externals" in gt4py, and I didn't have a better name in mind at the time 😁

PR #364, opened 2021-04-06
remove storage initialization and uncached copies

rheacangeo:
## Purpose

We are still seeing inspect and storage initialization popping up on the performance profile from the copy method and python assignments from init=True for temporary storage allocation. Added a cache key to the copy method, but in most cases remove copy in favor of merging the copy operation with a nearby stencil, or calling copy_stencil directly. Did not change UpdateDzC extensively because it has another refactor waiting. 
Testing on piz daint resulted in a speedup of ~10 seconds

## Code changes:

- for make_storage_data if a cache_key is not set expressly, the storage is not memoized (removed inspect)
- replaced all calls to make_storage_data that had init=True to not, setting to 0 in the stencil when needed
- removed upper_fix and lower_fix as temporaries outside the FillZ stencil as they are not needed outside of it
- replaced all calls to copy() with copy_stencil, merging the copy operation into another stencil, or using a cache_key for the storage created

## Checklist

### DIFF
--- a/fv3core/stencils/remapping_part1.py
+++ b/fv3core/stencils/remapping_part1.py
@@ -214,7 +214,7 @@ def compute(
 
     # do_omega = hydrostatic and last_step # TODO pull into inputs
     domain_jextra = (grid.nic, grid.njc + 1, grid.npz + 1)
-    pe1 = copy(pe, origin=grid.compute_origin(), domain=domain_jextra)
+    pe1 = copy(pe, origin=grid.compute_origin(), domain=domain_jextra, cache_key="remapping_part1_pe1")
     pe2 = utils.make_storage_from_shape(
         pe.shape, grid.compute_origin(), cache_key="remapping_part1_pe2"
     )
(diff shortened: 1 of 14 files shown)


fv3core/stencils/remapping_part1.py
```diff
@@ -116,8 +117,16 @@ def copy_j_adjacent(pe2: FloatField):
 
 
 @gtstencil()
-def pn2_and_pk(pe2: FloatField, pn2: FloatField, pk: FloatField, akap: float):
+def pn2_and_pk(
```
oelbert, line 120: Do we want to rename this to something involving delp as well, since it's also setting delp now?

oelbert (approved):
Good stuff! One minor point about naming, otherwise no notes

fv3core/stencils/remapping_part1.py
rheacangeo: sure, good idea

PR #366, opened 2021-04-06
Add UpdateDeltaZOnDGrid class for updatedzd

mcgibbon:
## Purpose

This PR refactors updatedzd to use a class instead of a compute function.

## Code changes:

- Added UpdateDeltaZOnDGrid
- Refactored updatedzd.compute to use UpdateDeltaZOnDGrid
- decorators.py now passes mypy checks and is removed from the mypy ignore list

## Checklist

- [ ] Docstrings and type hints are added to new and updated routines, as appropriate (this is being done in a follow-up PR)


fv3core/stencils/updatedzd.py
```diff
@@ -233,6 +232,172 @@ def edge_profile_stencil(
...
+    """
+
+    def __init__(self, grid, column_namelist, k_bounds):
+        self.grid = spec.grid
+        self._column_namelist = column_namelist
+        if any(
+            column_namelist["damp_vt"][kstart] <= 1e-5
+            for kstart in range(len(k_bounds))
+        ):
+            raise NotImplementedError("damp <= 1e-5 in column_cols is untested")
```
rheacangeo, line 247: this likely would be better suited in the future delnflux object, always separating these from where they were in the code could potentially cause some future confusion. But since we don't have the delnflux object and we don't want this check running every timestep, here is fine.

fv3core/stencils/updatedzd.py
```diff
@@ -233,6 +232,172 @@ def edge_profile_stencil(
...
+        self._k_bounds = k_bounds  # d_sw.k_bounds()
```
rheacangeo, line 248: looking forward to everything being 3d and not needing this

fv3core/stencils/updatedzd.py
```diff
@@ -233,6 +232,172 @@ def edge_profile_stencil(
...
+        largest_possible_shape = self.grid.domain_shape_full(add=(1, 1, 1))
```
rheacangeo, line 249: we use this for almost every object and we don't usually call it this, will be one of the things we can make consistent in another pass.

fv3core/stencils/updatedzd.py
```diff
@@ -233,6 +232,172 @@ def edge_profile_stencil(
...
+        Args:
+            dp0: ???
+            zs: ???
+            zh: ???
+            crx: Courant number in x-direction (??? what units)
+            cry: Courant number in y-direction (??? what units)
+            xfx: ???
+            yfx: ???
+            wsd: ???
+            dt: ???
```
rheacangeo, line 336: I would argue we know a lot of these, they have been documented in other code.  but can be picked up in a future PR

rheacangeo (approved):
no objections, a few comments

.pre-commit-config.yaml
```diff
@@ -22,7 +22,6 @@ repos:
         (?x)^(
         fv3core/utils/grid.py |
         fv3core/utils/gt4py_utils.py |
-        fv3core/decorators.py |
```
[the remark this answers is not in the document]
mcgibbon: I could have sworn I had, it looks like I didn't apply. I updated it just now so it'll be included when I merge.

fv3core/decorators.py
```diff
@@ -350,6 +350,10 @@ def __init__(self, func, origin, domain, **kwargs):
         )
 
     def __call__(self, *args, **kwargs) -> None:
+        if "origin" in kwargs:
+            raise ValueError("cannot pass origin to FixedOriginStencil at call time")
+        if "domain" in kwargs:
+            raise ValueError("cannot pass domain to FixedOriginStencil at call time")
```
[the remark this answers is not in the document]
mcgibbon: Thanks for the heads up. I'll update it to this version when I merge, since it should be more informative (someone might not realize they're calling a FixedOrigin stencil).

fv3core/stencils/updatedzd.py
mcgibbon: That sounds like I'd agree. This would be done by having the column namelist be an initialization argument for delnflux instead of a call-time argument, so that this assertion can still happen on init. This probably is the same every timestep?

fv3core/stencils/updatedzd.py
mcgibbon: Alternatively, if we're using the same shape for every storage we use, we could have a single static function `make_storage(origin=None)`. It feels a little weird to pass the same shape in every time we're calling this, and I couldn't come up with a name I'm happy with for the shape (including `largest_possible_shape`).

fv3core/stencils/updatedzd.py
mcgibbon: I've documented most of them in the follow-on PR. Trying to document these was what led to doing that PR, because I didn't understand for sure what `zh`, `xfx`, and `yfx` are until yesterday. If it's OK, let's review those docstrings there.

PR #369, opened 2021-04-07
Optimize origin and domain handling

twicki:
## Purpose

Our performance investigations showed that the calls to normalize the origin for each stencil are unnecessarily expensive. This PR solves that problem by moving the normalization process to the initialization-time of stencil objects. Additionally an lru_cache is applied to `grid.axis_offsets` to further reduce the expensive calls to the gt4py_tuple

## Code changes:

- Addition of the `FixedOriginStencil` that requires origin and domain at instantiation-time and stores the normalized origins
- Example usage of the `FixedOriginStencil` in `FVTP2D`
- Addtion of LRU caching to `grid.axis_offsets`

## Infrastructure changes:

- Switch to the GT4Py version that offers the functionality to pass in normalized origins

## Checklist

- [X] Docstrings and type hints are added to new and updated routines, as appropriate
- [X] All relevant documentation has been updated or added (e.g. README, CONTRIBUTING docs)
- [X] Unit tests are added or updated for non-stencil code changes


GT4PY_VERSION.txt
```diff
@@ -1 +1 @@
-v17
+bfd178f35cdbb292067d215e1f64477d3028c029
```
rheacangeo, line 1: I assume you didn't intend to leave this here?

fv3core/decorators.py
```diff
@@ -335,6 +335,32 @@ def decorator(func) -> FV3StencilObject:
...
+
+    def __init__(self, func, origin, domain, **kwargs):
+        self.normalized_origin = gtscript.gt_definitions.normalize_origin_mapping(
+            origin
+        )
+        self.domain = domain
+        self.func = func
+        self.stencil_object = gtscript.stencil(
+            backend=global_config.get_backend(),
+            rebuild=global_config.get_rebuild(),
```
rheacangeo, line 350: so here is where we can add @eddie-c-davis  turning off of synchronize, shouldn't be too hard to do then. He named these objects something else, so checking in with him to coordinate order of operations makes sense.

fv3core/stencils/fvtp2d.py
```diff
@@ -82,10 +83,21 @@ def __init__(self, namelist, hord):
...
+        )
+        self.stencil_q_j = FixedOriginStencil(
+            q_j_stencil,
+            origin=self.grid.full_origin(add=(3, 0, 0)),
+            domain=self.grid.domain_shape_full(add=(-3, 0, 1)),
+        )
+        self.stencil_transport_flux = FixedOriginStencil(
+            transport_flux_xy,
+            origin=self.grid.compute_origin(),
+            domain=self.grid.domain_shape_compute(add=(1, 1, 1)),
```
rheacangeo, line 99: I know there has been a push to put the origin and domain in the init instead of in the call, this does help make sure you use the right axis offsets with your origin. It might make understanding the code in 'call' a bit harder, but I think is worth it for the speedup we can get from removing normalize_origin from __call__

fv3core/decorators.py
```diff
@@ -335,6 +335,32 @@ def decorator(func) -> FV3StencilObject:
...
+            definition=self.func,
+            **kwargs,
+        )
+
+    def __call__(self, *args, **kwargs) -> None:
+        self.stencil_object(
+            *args,
+            **kwargs,
+            normalized_origin=self.normalized_origin,
+            domain=self.domain,
```
rheacangeo, line 360: It would be super nice to also include validate_args==global_config.get_validate_args() here and then remove the self.stencil_runtime_args assignment and use in calling stencils where we apply this. I know this is a little out of scope of your PR, but is a tedious silly thing that this object gives us the capability to remove

fv3core/utils/grid.py
```diff
@@ -419,6 +420,7 @@ def horizontal_starts_from_shape(self, shape):
             return 0, 0
 
 
+@functools.lru_cache
```
rheacangeo, line 423: sweet

rheacangeo (commented):
Yay! This should improve things so much and motivate us to make more objects

fv3core/decorators.py
eddie-c-davis: I created a [stencil_wrapper](https://github.com/VulcanClimateModeling/fv3core/blob/ab86522213ddce95692ac06bbf1ed9f504589d95/fv3core/decorators.py#L340) method to encapsulate this behavior.

jdahm:
@twicki nice work. If we're wrapping StencilObjects in this PR, can we incorporate the `exec_info` timings from https://github.com/VulcanClimateModeling/fv3core/pull/362? Thanks!

fv3core/utils/grid.py
jdahm: Doesn't this assume that the object is immutable? Should we be following https://stackoverflow.com/a/4828108 to ensure this?

twicki: no, this just requires hash-able arguments

GT4PY_VERSION.txt
twicki: so this was to test the new feature without having to build a gt4py release. But it needs a change there (to expose the API of normalized origins

fv3core/decorators.py
twicki: This can be combined

fv3core/decorators.py
twicki: this is done

twicki:
launch jenkins

fv3core/decorators.py
```diff
@@ -335,6 +335,33 @@ def decorator(func) -> FV3StencilObject:
...
+            rebuild=global_config.get_rebuild(),
+            definition=self.func,
+            **kwargs,
+        )
+
+    def __call__(self, *args, **kwargs) -> None:
+        self.stencil_object(
+            *args,
+            **kwargs,
+            validate_args=global_config.get_validate_args(),
```
rheacangeo, line 359: yay! thank you!

rheacangeo (approved):
Approved conditional on the GT4PY_VERSION being updated to a real tag. Yay!

rheacangeo (approved):
Looks good!

PR #370, opened 2021-04-07
a2b_ord4 uses stencils for corners instead of python code

rheacangeo:
## Purpose

While regions with gtbackends is under development, remove the python code that computes corners for a2b_ord4 so we can remove that impact from the performance and profile. A bigger refactor reducing the stencil call count is pending regions working.

## Code changes:

- move extrapolate corner code to stencils in a2b_ord4

## Checklist

### DIFF
--- a/fv3core/stencils/a2b_ord4.py
+++ b/fv3core/stencils/a2b_ord4.py
@@ -597,49 +597,48 @@ def compute_qout(qxx, qyy, qout, kstart, nk):
 
 
 def compute(qin, qout, kstart=0, nk=None, replace=False):
-    grid = spec.grid
     if nk is None:
-        nk = grid.npz - kstart
+        nk = grid().npz - kstart
     corner_domain = (1, 1, nk)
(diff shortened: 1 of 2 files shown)


fv3core/stencils/a2b_ord4.py
```diff
@@ -25,6 +25,203 @@ def grid():
     return spec.grid
 
 
+@gtscript.function
+def great_circle_dist_noradius(p1a, p1b, p2a, p2b):
```
twicki, line 29: seems like we're not using the with-radius one, can we remove the noradius from the name?

twicki (changes requested):

fv3core/stencils/a2b_ord4.py
rheacangeo: sure

twicki (approved):
making this regions-based seems out of scope so I'd suggest we move that to another PR

PR #372, opened 2021-04-07
Remove python code from updatedzc

rheacangeo:
## Purpose

We have a waiting PR  #247 that does all of updatedzc in a single stencil, the problem is it doesn't work for the gtx86 backend, likely due to the corners calls. While we wait for regions to solve this, just doing the minimal here to get validation and remove python functions from this. We are not doing the renaming or the docstrings or that stuff here, as that is part of the other PR that will be done soon. Let's just unmuddy the profile waters

## Code changes:

- updatedzc calls a fill_corners stencil rather than a routine that spawns a gtscript runtime function
- udpatedzc no longer uses the python code overwrite_edges and instead uses copy methods for stencils
- python code for overwriting edges is removed
- updatedzc test code is simplified as we do not need to return different variables than we pass in

## Checklist

### DIFF
--- a/fv3core/stencils/updatedzc.py
+++ b/fv3core/stencils/updatedzc.py
@@ -4,7 +4,7 @@ from gt4py.gtscript import BACKWARD, FORWARD, PARALLEL, computation, interval
 import fv3core._config as spec
 import fv3core.utils.global_constants as constants
 from fv3core.decorators import gtstencil
-from fv3core.stencils.basic_operations import copy
+from fv3core.stencils.basic_operations import copy, copy_stencil
 from fv3core.utils import corners
 from fv3core.utils.typing import FloatField, FloatFieldIJ, FloatFieldK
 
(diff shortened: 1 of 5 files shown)


fv3core/stencils/updatedzc.py
```diff
@@ -4,194 +4,47 @@
...
-
-#   do k = km, 1, -1
-#      do j=js1, je1
-#         do i=is1, ie1
-#            gz(i,j,k) = max( gz(i,j,k), gz(i,j,k+1) + dz_min )
-#         enddo
-#      enddo
-#   enddo
+@gtstencil
+def set_zero_2d(out_field: FloatFieldIJ):
```
eddie-c-davis, line 42: This could potentially be a common stencil like `copy_stencil`, but that can wait until needed.

eddie-c-davis (approved):
One small comment. Impressive refactor!

fv3core/stencils/updatedzc.py
rheacangeo: that's a good point! I can add that if you'd like, though the next refactor of this stencil will not use that, it'll all be 1 stencil and this hacky hack won't be needed

PR #374, opened 2021-04-08
Optimize gt origin

twicki:
## Purpose

Along the lines of https://github.com/VulcanClimateModeling/fv3core/pull/369 we can also normalize the domain at initialization-time

## Code changes:

- Extension of the `FixedOriginStencil` that requires to also store the normalized domain
- Assertions on `FixedOriginStencil` that at call-time no origin and domain are present

## Infrastructure changes:

- Switch to the GT4Py version that offers the functionality to pass in normalized domains

## Checklist


GT4PY_VERSION.txt
```diff
@@ -1 +1 @@
-v21
+4ffecd822891e07ddcb83cf4ecc945dceb03c3fc
```
twicki, line 1: this will change to the tag as soon as it is availabe

fv3core/decorators.py
```diff
@@ -337,8 +337,15 @@ class FixedOriginStencil:
     and using the normalized origins."""
 
     def __init__(self, func, origin, domain, **kwargs):
-        self.normalized_origin = gtscript.gt_definitions.normalize_origin_mapping(
-            origin
+        self.normalized_origin = (
```
eddie-c-davis, line 340: Should we assert that `origin` or `domain` be not `None` in `__init__` since they cannot be passed to `__call__`, or does the `gt4py` code handle that?

twicki: so for the normalized origin it is handled by creating an empty dict, this is as if you would call a gt4py stencil without an origin. And for the domain can be None for the same reason

eddie-c-davis: Sweet!

jdahm, line 340: Does this support passing an origin per field?

eddie-c-davis (approved):
LGTM

jdahm:
`FixedOriginStencil` is no longer a good name if the domain is also frozen. Maybe `FV3FrozenStencil`?

fv3core/decorators.py
twicki: yes of course

jdahm: 👍

PR #380, opened 2021-04-09
Oop/dyn core

rheacangeo:
## Purpose

The purpose is to gain a little speedup and unclutter the performance profile from python code. dyn_core has some python code setting data fields -- moved those into stencils, and turn dyn_core into a class AcousticDynamics to take advantage of the speedups we get from having normalize_origin in the init stages. There also are still some uncached storages here that are removed with the class.  

## Code changes:

- dyn_core now is a class AcousticDynamics
- the DynamicalCore(fv_dynamics) object now as an instance of AcousticDynamics as an attribute
- p_grad_c updated to work as an object attribute
- tests for DynCore and PGradC updated to work

## Checklist

### DIFF
--- a/fv3core/stencils/dyn_core.py
+++ b/fv3core/stencils/dyn_core.py
@@ -230,6 +230,7 @@ class AcousticDynamics:
         }
         self._p_grad_c = FixedOriginStencil(p_grad_c_stencil, **pgradc_kwargs)
         self._initialize_data = FixedOriginStencil(initialize_data, origin=self.grid.full_origin(), domain=self.grid.domain_shape_full(), externals={"hydrostatic":self.namelist.hydrostatic})
+        
     def __call__(self, state):
         # u, v, w, delz, delp, pt, pe, pk, phis, wsd, omga, ua, va, uc, vc, mfxd,
         # mfyd, cxd, cyd, pkz, peln, q_con, ak, bk, diss_estd, cappa, mdt, n_split,
--- a/fv3core/stencils/dyn_core.py
+++ b/fv3core/stencils/dyn_core.py
@@ -36,19 +36,32 @@ from fv3core.utils.typing import FloatField, FloatFieldIJ, FloatFieldK
 
 HUGE_R = 1.0e40
 
-def initialize_data(mfxd: FloatField, mfyd: FloatField, cxd: FloatField, cyd: FloatField, gz: FloatField, pk3: FloatField,diss_estd: FloatField,  init_step: bool):
+
+def initialize_data(
+    mfxd: FloatField,
+    mfyd: FloatField,
+    cxd: FloatField,
+    cyd: FloatField,
+    gz: FloatField,
+    pk3: FloatField,
+    diss_estd: FloatField,
+    init_step: bool,
+):
     with computation(PARALLEL), interval(...):
         from __externals__ import hydrostatic
+
         mfxd = 0.0
         mfyd = 0.0
         cxd = 0.0
         cyd = 0.0
         if init_step:
             gz = HUGE_R
-            with horizontal(region[3: -3, 3:-3]):
+            with horizontal(region[3:-3, 3:-3]):
                 diss_estd = 0.0
             if __INLINED(not hydrostatic):
                 pk3 = HUGE_R
+
+
 # NOTE in Fortran these are columns
 def dp_ref_compute(
     ak: FloatFieldK,
@@ -229,8 +242,13 @@ class AcousticDynamics:
             "externals": {"hydrostatic": self.namelist.hydrostatic, **ax_offsets},
         }
         self._p_grad_c = FixedOriginStencil(p_grad_c_stencil, **pgradc_kwargs)
-        self._initialize_data = FixedOriginStencil(initialize_data, origin=self.grid.full_origin(), domain=self.grid.domain_shape_full(), externals={"hydrostatic":self.namelist.hydrostatic})
-        
+        self._initialize_data = FixedOriginStencil(
+            initialize_data,
+            origin=self.grid.full_origin(),
+            domain=self.grid.domain_shape_full(),
+            externals={"hydrostatic": self.namelist.hydrostatic},
+        )
+
     def __call__(self, state):
         # u, v, w, delz, delp, pt, pe, pk, phis, wsd, omga, ua, va, uc, vc, mfxd,
         # mfyd, cxd, cyd, pkz, peln, q_con, ak, bk, diss_estd, cappa, mdt, n_split,
@@ -269,7 +287,16 @@ class AcousticDynamics:
 
         state.__dict__.update(self._temporaries)
 
-        self._initialize_data(state.mfxd, state.mfyd, state.cxd, state.cyd, state.gz, state.pk3, state.diss_estd, init_step)
+        self._initialize_data(
+            state.mfxd,
+            state.mfyd,
+            state.cxd,
+            state.cyd,
+            state.gz,
+            state.pk3,
+            state.diss_estd,
+            init_step,
+        )
         if not hydrostatic:
             # k1k = akap / (1.0 - akap)
             self._dp_ref_compute(
--- a/fv3core/stencils/dyn_core.py
+++ b/fv3core/stencils/dyn_core.py
@@ -37,29 +37,22 @@ from fv3core.utils.typing import FloatField, FloatFieldIJ, FloatFieldK
 HUGE_R = 1.0e40
 
 
-def initialize_data(
+def zero_data(
     mfxd: FloatField,
     mfyd: FloatField,
     cxd: FloatField,
     cyd: FloatField,
-    gz: FloatField,
-    pk3: FloatField,
     diss_estd: FloatField,
-    init_step: bool,
+    first_timestep: bool,
 ):
     with computation(PARALLEL), interval(...):
-        from __externals__ import hydrostatic
-
         mfxd = 0.0
         mfyd = 0.0
         cxd = 0.0
         cyd = 0.0
-        if init_step:
-            gz = HUGE_R
+        if first_timestep:
             with horizontal(region[3:-3, 3:-3]):
                 diss_estd = 0.0
-            if __INLINED(not hydrostatic):
-                pk3 = HUGE_R
 
 
 # NOTE in Fortran these are columns
@@ -143,18 +136,18 @@ def p_grad_c_stencil(
             )
 
 
-def get_n_con(namelist, grid):
+def get_nk_heat_dissipation(namelist, grid):
     if namelist.convert_ke or namelist.vtdm4 > 1.0e-4:
-        n_con = grid.npz
+        nk_heat_dissipation = grid.npz
     else:
         if namelist.d2_bg_k1 < 1.0e-3:
-            n_con = 0
+            nk_heat_dissipation = 0
         else:
             if namelist.d2_bg_k2 < 1.0e-3:
-                n_con = 1
+                nk_heat_dissipation = 1
             else:
-                n_con = 2
-    return n_con
+                nk_heat_dissipation = 2
+    return nk_heat_dissipation
 
 
 def dyncore_temporaries(shape, namelist, grid):
@@ -210,13 +203,19 @@ class AcousticDynamics:
     def __init__(self, comm: fv3gfs.util.CubedSphereCommunicator, namelist):
         self.comm = comm
         self.namelist = namelist
+        assert self.namelist.d_ext == 0
+        assert self.namelist.beta == 0
+        assert not self.namelist.use_logp
         self.grid = spec.grid
         self.do_halo_exchange = global_config.get_do_halo_exchange()
-        self.n_con = get_n_con(namelist, self.grid)
+        self._nk_heat_dissipation = get_nk_heat_dissipation(namelist, self.grid)
         self.nonhydrostatic_pressure = nh_p_grad.NonHydrostaticPressureGradient()
         self._temporaries = dyncore_temporaries(
             self.grid.domain_shape_full(add=(1, 1, 1)), self.namelist, self.grid
         )
+        self._temporaries["gz"][:] = HUGE_R
+        if not namelist.hydrostatic:
+            self._temporaries["pk3"][:] = HUGE_R
         self._dp_ref_compute = FixedOriginStencil(
             dp_ref_compute,
             origin=self.grid.full_origin(),
@@ -236,33 +235,29 @@ class AcousticDynamics:
         pgradc_origin = self.grid.compute_origin()
         pgradc_domain = self.grid.domain_shape_compute(add=(1, 1, 0))
         ax_offsets = axis_offsets(self.grid, pgradc_origin, pgradc_domain)
-        pgradc_kwargs = {
-            "origin": pgradc_origin,
-            "domain": pgradc_domain,
-            "externals": {"hydrostatic": self.namelist.hydrostatic, **ax_offsets},
-        }
-        self._p_grad_c = FixedOriginStencil(p_grad_c_stencil, **pgradc_kwargs)
-        self._initialize_data = FixedOriginStencil(
-            initialize_data,
+        self._p_grad_c = FixedOriginStencil(
+            p_grad_c_stencil,
+            origin=pgradc_origin,
+            domain=pgradc_domain,
+            externals={"hydrostatic": self.namelist.hydrostatic, **ax_offsets},
+        )
+        self._zero_data = FixedOriginStencil(
+            zero_data,
             origin=self.grid.full_origin(),
             domain=self.grid.domain_shape_full(),
-            externals={"hydrostatic": self.namelist.hydrostatic},
         )
 
     def __call__(self, state):
         # u, v, w, delz, delp, pt, pe, pk, phis, wsd, omga, ua, va, uc, vc, mfxd,
         # mfyd, cxd, cyd, pkz, peln, q_con, ak, bk, diss_estd, cappa, mdt, n_split,
         # akap, ptop, pfull, n_map, comm):
-        grid = self.grid
-        init_step = state.n_map == 1
         end_step = state.n_map == self.namelist.k_split
         akap = constants.KAPPA
         dt = state.mdt / self.namelist.n_split
         dt2 = 0.5 * dt
-        hydrostatic = self.namelist.hydrostatic
         rgrav = 1.0 / constants.GRAV
         n_split = self.namelist.n_split
-        # TODO: Put defaults into code.
+        # TODO: When the namelist values are set to 0, use these instead:
         # m_split = 1. + abs(dt_atmos)/real(k_split*n_split*abs(p_split))
         # n_split = nint( real(n0split)/real(k_split*abs(p_split)) * stretch_fac + 0.5 )
         ms = max(1, self.namelist.m_split / 2.0)
@@ -277,27 +272,25 @@ class AcousticDynamics:
                 "pt_quantity",
             ]:
                 reqs[halovar] = self.comm.start_halo_update(
-                    state.__getattribute__(halovar), n_points=utils.halo
+                    state.__getattribute__(halovar), n_points=self.grid.halo
                 )
             reqs_vector = self.comm.start_vector_halo_update(
-                state.u_quantity, state.v_quantity, n_points=utils.halo
+                state.u_quantity, state.v_quantity, n_points=self.grid.halo
             )
             reqs["q_con_quantity"].wait()
             reqs["cappa_quantity"].wait()
 
         state.__dict__.update(self._temporaries)
 
-        self._initialize_data(
+        self._zero_data(
             state.mfxd,
             state.mfyd,
             state.cxd,
             state.cyd,
-            state.gz,
-            state.pk3,
             state.diss_estd,
-            init_step,
+            state.n_map == 1,
         )
-        if not hydrostatic:
+        if not self.namelist.hydrostatic:
             # k1k = akap / (1.0 - akap)
             self._dp_ref_compute(
                 state.ak,
@@ -331,10 +324,10 @@ class AcousticDynamics:
             remap_step = False
             if self.namelist.breed_vortex_inline or (it == n_split - 1):
                 remap_step = True
-            if not hydrostatic:
+            if not self.namelist.hydrostatic:
                 if self.do_halo_exchange:
                     reqs["w_quantity"] = self.comm.start_halo_update(
-                        state.w_quantity, n_points=utils.halo
+                        state.w_quantity, n_points=self.grid.halo
                     )
                 if it == 0:
                     self._set_gz(
@@ -344,7 +337,7 @@ class AcousticDynamics:
                     )
                     if self.do_halo_exchange:
                         reqs["gz_quantity"] = self.comm.start_halo_update(
-                            state.gz_quantity, n_points=utils.halo
+                            state.gz_quantity, n_points=self.grid.halo
                         )
             if it == 0:
                 if self.do_halo_exchange:
@@ -360,7 +353,7 @@ class AcousticDynamics:
                     )
             if self.do_halo_exchange:
                 reqs_vector.wait()
-                if not hydrostatic:
+                if not self.namelist.hydrostatic:
                     reqs["w_quantity"].wait()
 
             # compute the c-grid winds at t + 1/2 timestep
@@ -383,26 +376,26 @@ class AcousticDynamics:
 
             if self.namelist.nord > 0 and self.do_halo_exchange:
                 reqs["divgd_quantity"] = self.comm.start_halo_update(
-                    state.divgd_quantity, n_points=utils.halo
+                    state.divgd_quantity, n_points=self.grid.halo
                 )
-            if not hydrostatic:
+            if not self.namelist.hydrostatic:
                 if it == 0:
                     if self.do_halo_exchange:
                         reqs["gz_quantity"].wait()
                     copy_stencil(
                         state.gz,
                         state.zh,
-                        origin=grid.full_origin(),
-                        domain=grid.domain_shape_full(add=(0, 0, 1)),
+                        origin=self.grid.full_origin(),
+                        domain=self.grid.domain_shape_full(add=(0, 0, 1)),
                     )
                 else:
                     copy_stencil(
                         state.zh,
                         state.gz,
-                        origin=grid.full_origin(),
-                        domain=grid.domain_shape_full(add=(0, 0, 1)),
+                        origin=self.grid.full_origin(),
+                        domain=self.grid.domain_shape_full(add=(0, 0, 1)),
                     )
-            if not hydrostatic:
+            if not self.namelist.hydrostatic:
                 updatedzc.compute(
                     state.dp_ref, state.zs, state.ut, state.vt, state.gz, state.ws3, dt2
                 )
@@ -423,8 +416,8 @@ class AcousticDynamics:
                 )
 
             self._p_grad_c(
-                grid.rdxc,
-                grid.rdyc,
+                self.grid.rdxc,
+                self.grid.rdyc,
                 state.uc,
                 state.vc,
                 state.delpc,
@@ -433,12 +426,12 @@ class AcousticDynamics:
                 dt2,
             )
             if self.do_halo_exchange:
-                reqc_vector = self.comm.start_vector_halo_update(
-                    state.uc_quantity, state.vc_quantity, n_points=utils.halo
+                req_vector_c_grid = self.comm.start_vector_halo_update(
+                    state.uc_quantity, state.vc_quantity, n_points=self.grid.halo
                 )
                 if self.namelist.nord > 0:
                     reqs["divgd_quantity"].wait()
-                reqc_vector.wait()
+                req_vector_c_grid.wait()
             # use the computed c-grid winds to evolve the d-grid winds forward
             # by 1 timestep
             d_sw.compute(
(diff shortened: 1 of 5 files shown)


fv3core/stencils/dyn_core.py
```diff
@@ -26,17 +26,43 @@
...
 
+def initialize_data(
+    mfxd: FloatField,
+    mfyd: FloatField,
+    cxd: FloatField,
+    cyd: FloatField,
+    gz: FloatField,
+    pk3: FloatField,
+    diss_estd: FloatField,
+    init_step: bool,
```
mcgibbon, line 48: Should these actually be 2 functions, one of which gets called in an object `__init__`? Note this might need to get called in the `DynamicalCore` init, or wherever these temporaries are allocated.

If not, because "initialize" is used for two different meanings (in this variable name, and in the function name), it would be helpful to document what `init_step` means, and why the initialization is different on the init step. Alternatively, you could update one or both names (`first_timestep: bool`? `zero_data`?).

fv3core/stencils/dyn_core.py
```diff
@@ -179,381 +197,415 @@ def dyncore_temporaries(shape):
...
 
-        # To write in parallel region, these need to be 3D first
-        state.dp_ref = utils.make_storage_from_shape(
-            shape, grid.full_origin(), cache_key="dyn_core_dp_ref"
+    def __init__(self, comm: fv3gfs.util.CubedSphereCommunicator, namelist):
+        self.comm = comm
+        self.namelist = namelist
+        self.grid = spec.grid
+        self.do_halo_exchange = global_config.get_do_halo_exchange()
+        self.n_con = get_n_con(namelist, self.grid)
```
mcgibbon, line 215: Please update `n_con` to be either private or plain English.

fv3core/stencils/dyn_core.py
```diff
@@ -179,381 +197,415 @@ def dyncore_temporaries(shape):
...
-                    domain=grid.domain_shape_full(add=(0, 0, 1)),
+        pgradc_origin = self.grid.compute_origin()
+        pgradc_domain = self.grid.domain_shape_compute(add=(1, 1, 0))
+        ax_offsets = axis_offsets(self.grid, pgradc_origin, pgradc_domain)
+        pgradc_kwargs = {
+            "origin": pgradc_origin,
+            "domain": pgradc_domain,
+            "externals": {"hydrostatic": self.namelist.hydrostatic, **ax_offsets},
+        }
+        self._p_grad_c = FixedOriginStencil(p_grad_c_stencil, **pgradc_kwargs)
```
mcgibbon, line 244: This feels like a lot of boilerplate to set up `pgradc_kwargs` to be used for one stencil. Can you refactor this to follow the same format you used for `self._initialize_data` below?

fv3core/stencils/dyn_core.py
```diff
@@ -179,381 +197,415 @@ def dyncore_temporaries(shape):
...
+        self._initialize_data = FixedOriginStencil(
+            initialize_data,
+            origin=self.grid.full_origin(),
+            domain=self.grid.domain_shape_full(),
+            externals={"hydrostatic": self.namelist.hydrostatic},
+        )
+
+    def __call__(self, state):
+        # u, v, w, delz, delp, pt, pe, pk, phis, wsd, omga, ua, va, uc, vc, mfxd,
```
mcgibbon, line 253: While this is dead code, I'll say it's helpful to keep because the call signature doesn't currently say what gets used.

fv3core/stencils/dyn_core.py
```diff
@@ -179,381 +197,415 @@ def dyncore_temporaries(shape):
...
+        end_step = state.n_map == self.namelist.k_split
+        akap = constants.KAPPA
+        dt = state.mdt / self.namelist.n_split
+        dt2 = 0.5 * dt
+        hydrostatic = self.namelist.hydrostatic
+        rgrav = 1.0 / constants.GRAV
+        n_split = self.namelist.n_split
+        # TODO: Put defaults into code.
+        # m_split = 1. + abs(dt_atmos)/real(k_split*n_split*abs(p_split))
+        # n_split = nint( real(n0split)/real(k_split*abs(p_split)) * stretch_fac + 0.5 )
```
mcgibbon, line 267: This TODO and commented code is confusing to me, can you either clarify the TODO, implement it, or delete this?

fv3core/stencils/dyn_core.py
```diff
@@ -179,381 +197,415 @@ def dyncore_temporaries(shape):
...
+                            state.pkc_quantity, n_points=2
+                        )
+                    else:
+                        reqs["pkc_quantity"] = self.comm.start_halo_update(
+                            state.pkc_quantity, n_points=utils.halo
+                        )
+                if remap_step:
+                    pe_halo.compute(state.pe, state.delp, state.ptop)
+                if self.namelist.use_logp:
+                    raise Exception("unimplemented namelist option use_logp=True")
```
mcgibbon, line 536:
```suggestion
                    raise NotImplementedError("unimplemented namelist option use_logp=True")
```
The base class `Exception` should never be raised, since it can only ever be caught with `except Exception:`, which catches every exception, including things like system errors you shouldn't catch. Some people go as far as to say you should never raise base exception types so that someone can always catch your particular exception, which is debatable. At least it makes sense to use `NotImplementedError` over `Exception` for features which aren't implemented.

fv3core/stencils/dyn_core.py
```diff
@@ -179,381 +197,415 @@ def dyncore_temporaries(shape):
...
+                    state.zh,
+                    state.gz,
+                    constants.GRAV,
+                    origin=(grid.is_ - 2, grid.js - 2, 0),
+                    domain=(grid.nic + 4, grid.njc + 4, grid.npz + 1),
+                )
+                if grid.npx == grid.npy and self.do_halo_exchange:
+                    reqs["pkc_quantity"].wait()
+                if self.namelist.beta != 0:
+                    raise Exception(
```
mcgibbon, line 554:
```suggestion
                    raise NotImplementedError(
```

fv3core/stencils/dyn_core.py
```diff
@@ -179,381 +197,415 @@ def dyncore_temporaries(shape):
...
+            origin=self.grid.full_origin(),
+            domain=self.grid.domain_shape_full(),
+            externals={"hydrostatic": self.namelist.hydrostatic},
+        )
+
+    def __call__(self, state):
+        # u, v, w, delz, delp, pt, pe, pk, phis, wsd, omga, ua, va, uc, vc, mfxd,
+        # mfyd, cxd, cyd, pkz, peln, q_con, ak, bk, diss_estd, cappa, mdt, n_split,
+        # akap, ptop, pfull, n_map, comm):
+        grid = self.grid
```
mcgibbon, line 256: Can you delete this line and use `self.grid` below instead? That would avoid a reader needing to question whether this came from an argument, global state, or self.

fv3core/stencils/dyn_core.py
```diff
@@ -179,381 +197,415 @@ def dyncore_temporaries(shape):
...
-                remap_step,
-                dt,
-                akap,
-                state.cappa,
-                state.ptop,
-                state.zs,
-                state.w,
-                state.delz,
-                state.q_con,
+            # compute the c-grid winds at t + 1/2 timestep
```
mcgibbon, line 366: Thought for later: we should be able to communicate this comment by naming c_sw to make it clear it's integrating winds (and only winds) forward on the c-grid, and passing `dt / 2.0` as the timestep.

fv3core/stencils/dyn_core.py
```diff
@@ -179,381 +197,415 @@ def dyncore_temporaries(shape):
...
+                state.omga,
+                dt2,
             )
 
-            if global_config.get_do_halo_exchange():
-                reqs["zh_quantity"] = comm.start_halo_update(
-                    state.zh_quantity, n_points=utils.halo
+            if self.namelist.nord > 0 and self.do_halo_exchange:
+                reqs["divgd_quantity"] = self.comm.start_halo_update(
+                    state.divgd_quantity, n_points=utils.halo
```
mcgibbon, line 386: `
```suggestion
                    state.divgd_quantity, n_points=self.grid.halo
```

fv3core/stencils/dyn_core.py
```diff
@@ -179,381 +197,415 @@ def dyncore_temporaries(shape):
...
+                    )
+                if it == 0:
+                    self._set_gz(
+                        state.zs,
+                        state.delz,
+                        state.gz,
+                    )
+                    if self.do_halo_exchange:
+                        reqs["gz_quantity"] = self.comm.start_halo_update(
+                            state.gz_quantity, n_points=utils.halo
```
mcgibbon, line 347:
```suggestion
                            state.gz_quantity, n_points=self.grid.halo
```

fv3core/stencils/dyn_core.py
```diff
@@ -179,381 +197,415 @@ def dyncore_temporaries(shape):
...
+            #
+            # The pressure gradient force and elastic terms are then evaluated
+            # backwards-in-time, to improve stability.
+            remap_step = False
+            if self.namelist.breed_vortex_inline or (it == n_split - 1):
+                remap_step = True
+            if not hydrostatic:
+                if self.do_halo_exchange:
+                    reqs["w_quantity"] = self.comm.start_halo_update(
+                        state.w_quantity, n_points=utils.halo
```
mcgibbon, line 337:
```suggestion
                        state.w_quantity, n_points=self.grid.halo
```

fv3core/stencils/dyn_core.py
```diff
@@ -179,381 +197,415 @@ def dyncore_temporaries(shape):
...
-                state.phis,
-                state.omga,
-                state.ptc,
-                state.q_con,
-                state.delpc,
-                state.gz,
-                state.pkc,
-                state.ws3,
+            reqs_vector = self.comm.start_vector_halo_update(
+                state.u_quantity, state.v_quantity, n_points=utils.halo
```
mcgibbon, line 283:
```suggestion
                state.u_quantity, state.v_quantity, n_points=self.grid.halo
```

fv3core/stencils/dyn_core.py
```diff
@@ -179,381 +197,415 @@ def dyncore_temporaries(shape):
...
+        reqs = {}
+        if self.do_halo_exchange:
+            for halovar in [
+                "q_con_quantity",
+                "cappa_quantity",
+                "delp_quantity",
+                "pt_quantity",
+            ]:
+                reqs[halovar] = self.comm.start_halo_update(
+                    state.__getattribute__(halovar), n_points=utils.halo
```
mcgibbon, line 280:
```suggestion
                    state.__getattribute__(halovar), n_points=self.grid.halo
```

fv3core/stencils/dyn_core.py
```diff
@@ -179,381 +197,415 @@ def dyncore_temporaries(shape):
...
+            # (the rest of the routines).
+            #
+            # Along-surface flux terms (mass, heat, vertical momentum, vorticity,
+            # kinetic energy gradient terms) are evaluated forward-in-time.
+            #
+            # The pressure gradient force and elastic terms are then evaluated
+            # backwards-in-time, to improve stability.
+            remap_step = False
+            if self.namelist.breed_vortex_inline or (it == n_split - 1):
+                remap_step = True
```
mcgibbon, line 333:
```suggestion
            remap_step = self.namelist.breed_vortex_inline or (it == n_split - 1)
```

fv3core/stencils/dyn_core.py
```diff
@@ -179,381 +197,415 @@ def dyncore_temporaries(shape):
...
-                state.ptop,
-                akap,
+                dt2,
             )
-
-        if spec.namelist.rf_fast:
-            # TODO: Pass through ks, or remove, inconsistent representation vs Fortran.
-            ray_fast.compute(
+            if self.do_halo_exchange:
+                reqc_vector = self.comm.start_vector_halo_update(
```
mcgibbon, line 436: `reqc` made me do a double take, can you simply name this `req_vector` (I think this is the important part of the intent), or more verbosely `req_vector_c_grid`? even `req` is fine, here, since you're waiting on the request immediately.

fv3core/stencils/dyn_core.py
```diff
@@ -179,381 +197,415 @@ def dyncore_temporaries(shape):
...
+                    self.comm.halo_update(
+                        state.__getattribute__(halovar), n_points=utils.halo
+                    )
+
+            # Not used unless we implement other betas and alternatives to nh_p_grad
+            # if self.namelist.d_ext > 0:
+            #    raise 'Unimplemented namelist option d_ext > 0'
+            # else:
+            #    divg2 = utils.make_storage_from_shape(delz.shape,
+            #   grid.compute_origin())
```
mcgibbon, line 485: Are we going to do this in the next 8 months? If not I'd suggest deleting it, it won't be too bad reasoning about this through the Fortran code.

fv3core/stencils/dyn_core.py
```diff
@@ -179,381 +197,415 @@ def dyncore_temporaries(shape):
...
+                    state.pkc,
+                    state.pk3,
+                    state.pk,
+                    state.peln,
+                    state.wsd,
+                )
+
+                if self.do_halo_exchange:
+                    reqs["zh_quantity"] = self.comm.start_halo_update(
+                        state.zh_quantity, n_points=utils.halo
```
mcgibbon, line 523:
```suggestion
                        state.zh_quantity, n_points=self.grid.halo
```

fv3core/stencils/dyn_core.py
```diff
@@ -179,381 +197,415 @@ def dyncore_temporaries(shape):
...
                     )
+                    if grid.npx == grid.npy:
+                        reqs["pkc_quantity"] = self.comm.start_halo_update(
+                            state.pkc_quantity, n_points=2
+                        )
+                    else:
+                        reqs["pkc_quantity"] = self.comm.start_halo_update(
+                            state.pkc_quantity, n_points=utils.halo
```
mcgibbon, line 531:
```suggestion
                            state.pkc_quantity, n_points=self.grid.halo
```

fv3core/stencils/dyn_core.py
```diff
@@ -179,381 +197,415 @@ def dyncore_temporaries(shape):
...
+        # u, v, w, delz, delp, pt, pe, pk, phis, wsd, omga, ua, va, uc, vc, mfxd,
+        # mfyd, cxd, cyd, pkz, peln, q_con, ak, bk, diss_estd, cappa, mdt, n_split,
+        # akap, ptop, pfull, n_map, comm):
+        grid = self.grid
+        init_step = state.n_map == 1
+        end_step = state.n_map == self.namelist.k_split
+        akap = constants.KAPPA
+        dt = state.mdt / self.namelist.n_split
+        dt2 = 0.5 * dt
+        hydrostatic = self.namelist.hydrostatic
```
mcgibbon, line 262: Same with `self.grid`, can we use `self.namelist.hydrostatic` below?

fv3core/stencils/dyn_core.py
```diff
@@ -179,381 +197,415 @@ def dyncore_temporaries(shape):
...
+            if it == n_split - 1 and end_step:
+                if self.namelist.use_old_omega:  # apparently True
+                    self._set_pem(
+                        state.delp,
+                        state.pem,
+                        state.ptop,
+                    )
+            if self.do_halo_exchange:
+                reqs_vector.wait()
+                if not hydrostatic:
```
mcgibbon, line 363:
```suggestion
                if not self.namelist.hydrostatic:
```

fv3core/stencils/dyn_core.py
```diff
@@ -179,381 +197,415 @@ def dyncore_temporaries(shape):
...
-                reqs["zh_quantity"] = comm.start_halo_update(
-                    state.zh_quantity, n_points=utils.halo
+            if self.namelist.nord > 0 and self.do_halo_exchange:
+                reqs["divgd_quantity"] = self.comm.start_halo_update(
+                    state.divgd_quantity, n_points=utils.halo
                 )
-                if grid.npx == grid.npy:
-                    reqs["pkc_quantity"] = comm.start_halo_update(
-                        state.pkc_quantity, n_points=2
+            if not hydrostatic:
```
mcgibbon, line 388:
```suggestion
            if not self.namelist.hydrostatic:
```

fv3core/stencils/dyn_core.py
```diff
@@ -179,381 +197,415 @@ def dyncore_temporaries(shape):
...
-                constants.GRAV,
-                origin=(grid.is_ - 2, grid.js - 2, 0),
-                domain=(grid.nic + 4, grid.njc + 4, grid.npz + 1),
-            )
-            if grid.npx == grid.npy and global_config.get_do_halo_exchange():
-                reqs["pkc_quantity"].wait()
-            if spec.namelist.beta != 0:
-                raise Exception(
-                    "Unimplemented namelist option -- we only support beta=0"
+            if not hydrostatic:
```
mcgibbon, line 405:
```suggestion
            if not self.namelist.hydrostatic:
```

fv3core/stencils/dyn_core.py
```diff
@@ -179,381 +197,415 @@ def dyncore_temporaries(shape):
...
-                state.ptop,
-                state.ks,
             )
+            # note that uc and vc are not needed at all past this point.
+            # they will be re-computed from scratch on the next acoustic timestep.
+
+            if self.do_halo_exchange:
+                for halovar in ["delp_quantity", "pt_quantity", "q_con_quantity"]:
+                    self.comm.halo_update(
+                        state.__getattribute__(halovar), n_points=utils.halo
```
mcgibbon, line 477:
```suggestion
                        state.__getattribute__(halovar), n_points=self.grid.halo
```

fv3core/stencils/dyn_core.py
```diff
@@ -179,381 +197,415 @@ def dyncore_temporaries(shape):
...
+                basic.multiply_constant(
+                    state.zh,
+                    state.gz,
+                    constants.GRAV,
+                    origin=(grid.is_ - 2, grid.js - 2, 0),
+                    domain=(grid.nic + 4, grid.njc + 4, grid.npz + 1),
+                )
+                if grid.npx == grid.npy and self.do_halo_exchange:
+                    reqs["pkc_quantity"].wait()
+                if self.namelist.beta != 0:
```
mcgibbon, line 553: Can you replicate or move the logic for the `NotImplementedError`s to the init?

fv3core/stencils/dyn_core.py
```diff
@@ -179,381 +197,415 @@ def dyncore_temporaries(shape):
...
-                state.delz,
-                state.delp,
-                state.cappa,
-                n_con,
-                dt,
-            )
+            if self.do_halo_exchange:
+                if it != n_split - 1:
+                    reqs_vector = self.comm.start_vector_halo_update(
+                        state.u_quantity, state.v_quantity, n_points=utils.halo
```
mcgibbon, line 586:
```suggestion
                        state.u_quantity, state.v_quantity, n_points=self.grid.halo
```

fv3core/stencils/dyn_core.py
```diff
@@ -179,381 +197,415 @@ def dyncore_temporaries(shape):
...
+                        )
+
+        if self.n_con != 0 and self.namelist.d_con > 1.0e-5:
+            nf_ke = min(3, self.namelist.nord + 1)
+
+            if self.do_halo_exchange:
+                self.comm.halo_update(state.heat_source_quantity, n_points=utils.halo)
+            cd = constants.CNST_0P20 * grid.da_min
+            del2cubed.compute(state.heat_source, nf_ke, cd, grid.npz)
+            if not hydrostatic:
```
mcgibbon, line 601:
```suggestion
            if not self.namelist.hydrostatic:
```

fv3core/testing/parallel_translate.py
```diff
@@ -176,13 +176,13 @@ def compute_sequential(self, a, b):
 
 
 class ParallelTranslate2PyState(ParallelTranslate2Py):
-    def compute_parallel(self, inputs, communicator):
+    def compute_parallel(self, inputs):
```
mcgibbon, line 179: Changing the call signature of this can be confusing for people who are used to the previous signature, and when reading other objects using this class or similar parallel translate classes.

Instead of changing the API, could you instead initialize the object inside `compute_parallel`, rather than inside the translate class `init`? It does look like this is what you're actually doing, since while you changed the base class API, the subclass you refactored still takes in `communicator` in its compute. Since these are being called once, it shouldn't matter whether the class gets made on each call or on each init.

mcgibbon (changes requested):
I tried to catch all the calls of `utils.halo`, but it's probably worth doing a text search for more.

The most substantial code comment is that I'd ask we not change the API of ParallelTranslate2PyState in this PR, since the testing logic is fairly complex and right now the APIs of all parallel classes are at least fairly coherent with one another.

If you're up to it, the refactor I suggested for `initialize_data` is fairly relevant for this PR since it impacts the init/compute decomposition of responsibilities, but I'm OK if you want to leave it for later.

fv3core/stencils/dyn_core.py
rheacangeo: I think if we expand the model, this has potential to be one of the first places we do so. I think as we rename everything and reorganizing, going back to the fortran code will be harder and harder. This particular one seems more important than the bigger ones like hydrostatic or grid type >=3, this is unsupported behavior on a combination of namelist options... I can uncomment the exception though and put it in the init

fv3core/stencils/dyn_core.py
rheacangeo: never say never :). I agree it's good to be specific with error types and I'll make the change here. I disagree that one should never raise Exception. You may want to have an exception that can't be caught, you may not have the time to go with a better one.

fv3core/stencils/dyn_core.py
rheacangeo: I can replicate it, though prefer not to move it for the use_logp one since the control flow logic being in place will make it easier to expand the options when we do so. the beta one can be moved easily because it has nothing to do with anything around it

tests/translate/translate_map1_ppm_2d.py
```diff
@@ -53,6 +53,7 @@ def compute(self, inputs):
         inputs["j1"] = inputs["j_2d"]
         inputs["j2"] = inputs["j_2d"]
         del inputs["j_2d"]
+        print("qs", inputs["qs"].shape)
```
mcgibbon, line 56:
```suggestion
```

fv3core/stencils/dyn_core.py
```diff
@@ -132,29 +136,36 @@ def p_grad_c_stencil(
             )
 
 
-def get_n_con():
-    if spec.namelist.convert_ke or spec.namelist.vtdm4 > 1.0e-4:
-        n_con = spec.grid.npz
+def get_nk_heat_dissipation(namelist, grid):
```
mcgibbon, line 139: A comment for devs would be useful here based on what you told me earlier.

```suggestion
def get_nk_heat_dissipation(namelist, grid):
    # determines whether to convert dissipated kinetic energy into heat in the full
    # column, not at all, or in 1 or 2 of the top of atmosphere sponge layers
```

fv3core/stencils/dyn_core.py
```diff
@@ -132,29 +136,36 @@ def p_grad_c_stencil(
...
+def dyncore_temporaries(shape, namelist, grid):
     tmps = {}
     utils.storage_dict(
         tmps,
         ["ut", "vt", "gz", "zh", "pem", "pkc", "pk3", "heat_source", "divgd"],
         shape,
         grid.full_origin(),
     )
+    if not namelist.hydrostatic:
+        # To write in parallel region, these need to be 3D first
```
mcgibbon, line 162: This comment doesn't make sense to me, could you please either expand it or delete it? What is a "parallel region" and why does it mean these have to be 3D?

fv3core/stencils/dyn_core.py
```diff
@@ -132,29 +136,36 @@ def p_grad_c_stencil(
...
+        utils.storage_dict(
+            tmps,
+            ["dp_ref", "zs"],
+            shape,
+            grid.full_origin(),
+        )
```
mcgibbon, line 168: Do these need to be initialized here, or is there somewhere lower down where they're used that they could be initialized? Having it here makes it unclear from the code whether it's used in one or multiple sub-components (it seems like only one, since this code wasn't here before?). Because of this, ideally only temporaries that get passed to more than one sub-component of DynamicalCore would be initialized here.

fv3core/stencils/dyn_core.py
```diff
@@ -179,381 +190,413 @@ def dyncore_temporaries(shape):
...
+    Peforms the Lagrangian acoustic dynamics described by Lin 2004
+    """
 
-        # To write in parallel region, these need to be 3D first
-        state.dp_ref = utils.make_storage_from_shape(
-            shape, grid.full_origin(), cache_key="dyn_core_dp_ref"
+    def __init__(self, comm: fv3gfs.util.CubedSphereCommunicator, namelist):
+        self.comm = comm
+        self.namelist = namelist
+        assert self.namelist.d_ext == 0
```
mcgibbon, line 206:
```suggestion
        assert self.namelist.d_ext == 0, "d_ext != 0 is not implemented"
```

fv3core/stencils/dyn_core.py
```diff
@@ -179,381 +190,413 @@ def dyncore_temporaries(shape):
...
+        assert self.namelist.beta == 0
```
mcgibbon, line 207:
```suggestion
        assert self.namelist.beta == 0, "beta != 0 is not implemented"
```

fv3core/stencils/dyn_core.py
```diff
@@ -179,381 +190,413 @@ def dyncore_temporaries(shape):
...
+        assert not self.namelist.use_logp
```
mcgibbon, line 208:
```suggestion
        assert not self.namelist.use_logp, "use_logp=True is not implemented"
```

fv3core/stencils/dyn_core.py
```diff
@@ -179,381 +190,413 @@ def dyncore_temporaries(shape):
...
+        self.grid = spec.grid
+        self.do_halo_exchange = global_config.get_do_halo_exchange()
+        self._nk_heat_dissipation = get_nk_heat_dissipation(namelist, self.grid)
+        self.nonhydrostatic_pressure = nh_p_grad.NonHydrostaticPressureGradient()
```
mcgibbon, line 212: Can you rename this to something like `nonhydrostatic_pressure_gradient`, `nh_pressure_gradient`, or `nonhydrostatic_pg`? The "gradient" part is an important distinction, "nonhydrostatic_pressure" could be misleading. If this name is too long, it is probably better to shorten parts of it, and at least have it documented in the class name/attribute type. "nonhydrostatic pressure" to me is `p'` where pressure is defined as `p = p_hs + p'` and `p_hs` is the hydrostatic component of pressure.

Not necessarily required right now, but also nh_p_grad applies the pressure gradient force/PGF, as opposed to computing a pressure gradient, so a full name would probably be something like `NonHydrostaticPressureGradientForce`. Maybe `PressureGradientForce` is enough for us, since we are explicitly writing a nonhydrostatic model? `self.pressure_gradient_force`? It makes it clearer that this is the full hydrostatic + nonhydrostatic component, whereas "nonhydrostatic pressure gradient force" kind of says you're leaving out the hydrostatic part.

mcgibbon: Interesting follow-up question: is `p_grad_c` nonhydrostatic?

fv3core/stencils/dyn_core.py
```diff
@@ -179,381 +190,413 @@ def dyncore_temporaries(shape):
...
+                        reqs["gz_quantity"] = self.comm.start_halo_update(
+                            state.gz_quantity, n_points=self.grid.halo
+                        )
+            if it == 0:
+                if self.do_halo_exchange:
+                    reqs["delp_quantity"].wait()
+                    reqs["pt_quantity"].wait()
+
+            if it == n_split - 1 and end_step:
+                if self.namelist.use_old_omega:  # apparently True
```
mcgibbon, line 348: Not from your PR so this is optional, but do you know what "apparently True" means? Could you expand the comment or delete it?

mcgibbon (commented):
The main thing I'd ask to change is initializing dp_ref and zs further down the object heirarchy, if you can. I'd also appreciate if you can merge the line suggestions. `nonhydrostatic_pressure` should be renamed but I'm OK if you don't answer all the questions I had about it right away, and use a "temporary" name before looking into the answers. I think in the end the right name is `self.pressure_gradient_force` but it may take some digging around to know for sure.

fv3core/stencils/dyn_core.py
rheacangeo: sure

fv3core/stencils/dyn_core.py
rheacangeo: deleted

fv3core/stencils/dyn_core.py
rheacangeo: They are used in this class, and are used to have temporary 3d storages so they can be assigned values, and then recast as a 1d and 2d variable for use lower down. We will definitely want to revisit this, but I think they make sense being here for now.

fv3core/stencils/dyn_core.py
rheacangeo: updated comment

tests/translate/translate_map1_ppm_2d.py
rheacangeo: thanks for the catch

mcgibbon (approved):
Thanks for the updates! LGTM.

PR #385, opened 2021-04-12
Oop/remap profile

oelbert:
## Purpose

This PR turns the remap_profile module into a class

## Code changes:

- remap_profile.py: `compute` has been replaced by the `RemapProfile` class, `kord` and `iv` are now externals, and references to `kord` and `iv` are `__INLINED` where possible. Documentation has been added
- translate_remap_profile_2d.py initializes and calls `RemapProfile` as it's compute function
- mapn_tracer.py: now calls `RemapProfile` instead of `remap_profile.compute`
- map_single.py: `RemapProfile` instead of `remap_profile.compute`
- translate_map1_ppm_2d.py: now passes `qs` as a full 3D variable to map_single in order to keep the shape consistent with the handling for j1 and j2.

## Checklist

### DIFF
--- a/fv3core/stencils/map_single.py
+++ b/fv3core/stencils/map_single.py
@@ -4,10 +4,10 @@ import numpy as np
 from gt4py.gtscript import FORWARD, PARALLEL, computation, interval
 
 import fv3core._config as spec
-from fv3core.stencils.remap_profile import RemapProfile
 import fv3core.utils.gt4py_utils as utils
 from fv3core.decorators import gtstencil
 from fv3core.stencils.basic_operations import copy_stencil
+from fv3core.stencils.remap_profile import RemapProfile
 from fv3core.utils.typing import FloatField, FloatFieldIJ
 
 
--- a/fv3core/stencils/map_single.py
+++ b/fv3core/stencils/map_single.py
@@ -4,7 +4,7 @@ import numpy as np
 from gt4py.gtscript import FORWARD, PARALLEL, computation, interval
 
 import fv3core._config as spec
-import fv3core.stencils.remap_profile as remap_profile
+from fv3core.stencils.remap_profile import RemapProfile
 import fv3core.utils.gt4py_utils as utils
 from fv3core.decorators import gtstencil
 from fv3core.stencils.basic_operations import copy_stencil
@@ -107,11 +107,14 @@ def compute(
     qmin: float = 0.0,
     version: str = "stencil",
 ):
+    remap_profile_k = utils.cached_stencil_class(RemapProfile)(
+        kord, mode, cache_key="map_single_profile"
+    )
     dp1, q4_1, q4_2, q4_3, q4_4, origin, domain, i_extent, j_extent = setup_data(
         q1, pe1, i1, i2, j1, j2
     )
-    q4_1, q4_2, q4_3, q4_4 = remap_profile.compute(
-        qs, q4_1, q4_2, q4_3, q4_4, dp1, spec.grid.npz, i1, i2, j1, j2, mode, kord, qmin
+    q4_1, q4_2, q4_3, q4_4 = remap_profile_k(
+        qs, q4_1, q4_2, q4_3, q4_4, dp1, i1, i2, j1, j2, qmin
     )
     do_lagrangian_contributions(
         q1,
--- a/fv3core/stencils/map_single.py
+++ b/fv3core/stencils/map_single.py
@@ -108,7 +108,7 @@ def compute(
     version: str = "stencil",
 ):
     remap_profile_k = utils.cached_stencil_class(RemapProfile)(
-        kord, mode, cache_key="map_single_profile"
+        kord, mode, cache_key=f"map_profile_{kord}_{mode}"
     )
     dp1, q4_1, q4_2, q4_3, q4_4, origin, domain, i_extent, j_extent = setup_data(
         q1, pe1, i1, i2, j1, j2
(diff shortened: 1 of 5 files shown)


rheacangeo:
launch jenkins

fv3core/stencils/map_single.py
```diff
@@ -107,11 +107,14 @@ def compute(
...
+    remap_profile_k = utils.cached_stencil_class(RemapProfile)(
+        kord, mode, cache_key=f"map_profile_{kord}_{mode}"
+    )
     dp1, q4_1, q4_2, q4_3, q4_4, origin, domain, i_extent, j_extent = setup_data(
         q1, pe1, i1, i2, j1, j2
     )
-    q4_1, q4_2, q4_3, q4_4 = remap_profile.compute(
-        qs, q4_1, q4_2, q4_3, q4_4, dp1, spec.grid.npz, i1, i2, j1, j2, mode, kord, qmin
+    q4_1, q4_2, q4_3, q4_4 = remap_profile_k(
+        qs, q4_1, q4_2, q4_3, q4_4, dp1, i1, i2, j1, j2, qmin
```
rheacangeo, line 117: yay! removing arguments from runtime is awesome

fv3core/stencils/mapn_tracer.py
```diff
@@ -40,6 +40,10 @@ def compute(
     kord: int,
     version: str = "stencil",
 ):
+    remapping_calculation = utils.cached_stencil_class(RemapProfile)(
+        kord, 0, cache_key=f"map_profile_{kord}_0"
```
rheacangeo, line 44: oh very neat, can share this same cached object with other places it is used

fv3core/stencils/remap_profile.py
```diff
@@ -117,12 +116,12 @@ def set_vals(
...
     qs: FloatField,
-    iv: int,
-    kord: int,
 ):
+    from __externals__ import iv, kord
+
     with computation(PARALLEL), interval(0, 1):
         # set top
-        if iv == -2:
+        if __INLINED(iv == -2):
```
rheacangeo, line 124: I am excited to see runtime conditionals become compile-time!

fv3core/stencils/remap_profile.py
```diff
@@ -469,158 +467,207 @@ def set_inner(
...
+        i1: int,
+        i2: int,
+        j1: int,
+        j2: int,
+        qmin: float = 0.0,
+    ):
+        """
+        Calculates the interpolation coefficients for a cubic-spline which models the
+        distribution of the remapped field within each deformed grid cell.
+        The constraints on the spline are set by kord and iv.
```
rheacangeo, line 590: excellent docstring, you could potentially go into more detail about what kord and iv are, but don't have to for this pr

fv3core/stencils/remap_profile.py
```diff
@@ -469,158 +467,207 @@ def set_inner(
...
             a4_3,
             a4_4,
-            extm,
-            iv,
-            origin=(i1, j1, km - 2),
-            domain=(i_extent, j_extent, 2),
+            self._q_bot,
+            qs,
+            origin=orig,
+            domain=dom_extend,
```
rheacangeo, line 621: potentially in a future version we might be able to identify the variations if i1/i2/j1/j2 that exist and make a different object for each, then compile the stencils using the  FixedOriginStencil, giving the origin and domain in compile time, reducing the time spend in normalize_origin and normalize_domain. But the impact may be small, and can be a future exploration. Making this an object is a huge and great step, and that would make sense as a another PR.

tests/translate/translate_remap_profile_2d.py
```diff
@@ -50,6 +49,7 @@ def make_storage_data_input_vars(self, inputs, storage_vars=None):
                 del inputs[serialname]
 
     def compute(self, inputs):
+        self.compute_func = profile.RemapProfile(inputs["kord"], inputs["iv"])
```
rheacangeo, line 52: ah yes, when it depends on input data arguments, we can't initialize in the init.

tests/translate/translate_map1_ppm_2d.py
```diff
@@ -37,9 +43,15 @@ def __init__(self, grid):
         self.out_vars = {"var_inout": {}}
         self.max_error = 5e-13
         self.write_vars = ["qs"]
+        self.nj = grid.npy
 
     def compute(self, inputs):
+        print(self.nj)
+        print("test_code")
```
rheacangeo, line 50: remove print statements?

tests/translate/translate_map1_ppm_2d.py
```diff
@@ -4,6 +4,12 @@
 from fv3core.testing import TranslateFortranData2Py, TranslateGrid
 
 
+def pad_field_in_j(field, nj):
+    outfield = np.tile(field[:, 0, :], [nj, 1, 1]).transpose(1, 0, 2)
+    np.testing.assert_array_equal(outfield[:, 0, :], field[:, 0, :])
+    return outfield
```
rheacangeo, line 10: This is a different method than used in the remap_profile_2d to achieve the same thing, should they share the same approach for less confusion next time we revisit these? Or are they different?

rheacangeo (approved):
Awesome job -- made some comments. Nothing is blocking if you want to merge asap, but I think the test code suggestion might be worth considering before merging.

fv3core/stencils/remap_profile.py
oelbert: In principle that makes sense, but right now the only place where the i and j extents aren't the compute domain are in the test code, I believe? Regardless, I agree that that probably makes sense as a separate PR

tests/translate/translate_map1_ppm_2d.py
oelbert: Good question! I initially wrote `pad_field_in_j` in translate_remap_profile_2d.py, but it seems it wasn't necessary there or in the other remapping classes. I agree that we should be using the same method for both, but I was nervous to implement that without potentially breaking more things

rheacangeo: I think it's fine to leave as-is for now

PR #387, opened 2021-04-13
update to v23 with constant folding

twicki:
## Purpose

Update to gt4py version 23 (with constant folding, reducing temporaries and adding the option to have non-synchronizing stencils).
Note that none of the fv3core stencils make use of disabling synchronization yet.

## Requirements changes:

- change gt4py to version 23

### DIFF
--- a/GT4PY_VERSION.txt
+++ b/GT4PY_VERSION.txt
@@ -1,1 +1,1 @@
-v22
+d2ca5ab750fa601aeb0506faf4bc159b38fc1530


GT4PY_VERSION.txt
```diff
@@ -1 +1 @@
-v22
+d2ca5ab750fa601aeb0506faf4bc159b38fc1530
```
twicki, line 1: this will be a tag as soon as it is there

PR #388, opened 2021-04-13
Update gt4py

twicki:
## Purpose

Update to v24 with constant folding, reducing temp size and enabling non-syncing stencils (not yet used in fv3core)


eddie-c-davis (approved):
LGTM

PR #389, opened 2021-04-13
validate zh in UpdateDzD on only compute domain, remove temporary

mcgibbon:
## Purpose

Currently our validation tests check that the output arrays are identical to the Fortran outputs, including in halo regions. For some stencils this is valid because halo values are updated and then used by other stencils, but in many cases it is only necessary to test a subset of the output for validation, because the halo values are overwritten without being read.

This PR adds a solution to validate only a subset of the output, while also guaranteeing the non-validated data is unused by setting it to NaN during validation tests (including larger tests like fv_dynamics and dyn_core). In this PR that solution is used to validate zh in UpdateDzD on only the compute domain, allowing us to remove a temporary copy of zh whose compute domain is never modified.

## Code changes:

- Validate zh on only the compute domain in UpdateDeltaZOnDGrid, setting halo output values to nan during tests
- Remove z2/zh_tmp temporary from UpdateDeltaZOnDGrid, using zh directly instead
- Added SelectiveValidation class which provides a slice to subset arrays for validation, a method to set arrays to NaN when the class is in test mode, and a dictionary for translate class output specifications
- Updated test_translate to check translate classes for a `subset_output(varname, array) -> array` method which subsets outputs for validation (or returns the unmodified array). If the method is not present, the behavior is unchanged.

## Infrastructure changes:

- parallel test command is updated so that they fail when any worker fails, instead of deadlocking if only a subset of workers fail

## Checklist


tests/test_translate.py
```diff
@@ -204,9 +206,9 @@ def test_sequential_savepoint(
                 print_failures,
                 failure_stride,
                 test_name,
-                ignore_near_zero,
-                testobj.near_zero,
-                xy_indices,
+                ignore_near_zero_errors=ignore_near_zero,
```
mcgibbon, line 209: This is updated because I was getting weird behaviors like variable values not propagating through this call. I wanted to be sure they weren't getting crossed.

tests/translate/translate_updatedzd.py
```diff
@@ -24,11 +24,19 @@ def __init__(self, grid):
...
-        return self.slice_output(inputs)
+        self.updatedzd(**inputs)
+        self.out_vars["zh"].update(self.updatedzd._zh_validator.translate_dict)
+        outputs = self.slice_output(inputs)
+        outputs["zh"] = self.subset_output("zh", outputs["zh"])
+        return outputs
+
+    def subset_output(self, varname, output):
+        if varname == "zh":
+            output = output[self.updatedzd._zh_validator.validation_slice]
```
rheacangeo, line 41: ah you do refer to the updatedzd object's zh_validator

fv3core/utils/validation.py
```diff
@@ -0,0 +1,42 @@
...
+
+    @property
+    def translate_dict(self):
+        return {
+            "i_start": self.origin[0],
+            "j_start": self.origin[1],
+            "k_start": self.origin[2],
+            "i_end": self.origin[0] + self.domain[0],
+            "j_end": self.origin[1] + self.domain[1],
+            "k_end": self.origin[2] + self.domain[2],
```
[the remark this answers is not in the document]
rheacangeo: On second though after seeing how it is used, I think you don't need to have a translate_dict

rheacangeo (commented):
Awesome work, this will be a nice feature to have . Adding a test was good. Is there a way we can verify in a test or manually, that if we set validator bounds too small the parent test does indeed not work? It looks like it will behave as expected and discussed, however.

tests/translate/translate_updatedzd.py
rheacangeo, line 41: This works for this case, but want to point out the validation_slice here is defined from the point of reference of the full domain. e.g. because zh starts at isd, jsd, or at the start of the full domain including halos, setting the validation region using 'compute_domain' works just fine. But some data comparisons are of a fields that start at different halo lines, e.g, is-2 or is - 1, and grid.compute_domain() as a slice would be further into the interior than you likely want, all of the familiar grid indexes would be off when slicing the array.

fv3core/utils/validation.py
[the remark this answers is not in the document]
rheacangeo: Since this is not being applied, I think you don't need it

tests/translate/translate_updatedzd.py
rheacangeo: Alternatively I wonder if it would make sense to fill the reference data with nans rather than slicing, but slicing is probably easier to debug

fv3core/utils/validation.py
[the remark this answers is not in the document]
mcgibbon: I actually don't know what these values are supposed to be, I thought I was using the same ones as the translate class and took the tests passing to mean this was correct. Would this be correct if I rename these keys to e.g. `istart` as suggested? Is this necessary at all?

tests/translate/translate_updatedzd.py
```diff
@@ -24,11 +24,19 @@ def __init__(self, grid):
...
 
     def compute(self, inputs):
         self.make_storage_data_input_vars(inputs)
-        updatedzd = fv3core.stencils.updatedzd.UpdateDeltaZOnDGrid(
-            self.grid, d_sw.get_column_namelist(), d_sw.k_bounds()
-        )
-        updatedzd(**inputs)
-        return self.slice_output(inputs)
+        self.updatedzd(**inputs)
+        self.out_vars["zh"].update(self.updatedzd._zh_validator.translate_dict)
```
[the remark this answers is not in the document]
mcgibbon: OK, I'll delete this part of the logic. That's great.

fv3core/utils/validation.py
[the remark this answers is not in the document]
mcgibbon: Will remove this as suggested below.

fv3core/stencils/d_sw.py
```diff
@@ -608,7 +608,7 @@ def compute(
     xfx,
     yfx,
     q_con,
-    zh,
+    zh,  # input only
```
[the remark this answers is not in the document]
mcgibbon: Yes, but there currently is no docstring and this isn't a module I'm editing, so I did this as a faster way to add the information. It should be put into the docstring when that gets added.

Makefile
```diff
@@ -36,7 +36,8 @@ TEST_TYPE=$(word 3, $(subst _, ,$(EXPERIMENT)))
 THRESH_ARGS=--threshold_overrides_file=$(FV3_PATH)/tests/translate/overrides/$(TEST_TYPE).yaml
 TEST_ARGS_USE=$(TEST_ARGS) $(THRESH_ARGS)
 PYTEST_SEQUENTIAL=pytest --data_path=$(TEST_DATA_RUN_LOC) $(TEST_ARGS_USE) $(FV3_PATH)/tests
-PYTEST_PARALLEL=$(MPIRUN_CALL) pytest --data_path=$(TEST_DATA_RUN_LOC) $(TEST_ARGS_USE) -m parallel $(FV3_PATH)/tests
+# we can't rule out a deadlock if one test fails, so we must set maxfail=1 for parallel tests
+PYTEST_PARALLEL=$(MPIRUN_CALL) python -m mpi4py -m pytest --maxfail=1 --data_path=$(TEST_DATA_RUN_LOC) $(TEST_ARGS_USE) -m parallel $(FV3_PATH)/tests
```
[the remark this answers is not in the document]
mcgibbon: We need something like this for daint tests, though there is a potential use case if you run locally and know the tests you're running won't deadlock and want to see if more than the first test fails. Right now you'd edit the Makefile and delete the maxfail temporarily to do that.

The alternative is to somehow incorporate it into one of the "passable" arguments like TEST_ARGS, but this is complicated by the fact that we want this by default for only parallel tests. So we'd have to make `TEST_ARGS_PARALLEL?= --maxfail=1 $(TEST_ARGS)` and then pass TEST_ARGS_PARALLEL from the command line to override the maxfail. Learning these two arguments for this very particular use case seems complex enough to me that it's OK to say you have to edit the Makefile for this use case. I'm willing to do it the other way if you think this use case comes up more often, though.

tests/translate/translate_updatedzd.py
mcgibbon: Sort of... test code is allowed to access private members of the tested code, though I do still feel weird about anything accessing a private member of a class instance. The tiebreaker here for me was that I really don't think this belongs as a documented attribute shown to an external user of the `fv3core` package, so because the class of `updatedzd` is public the attribute needs to be private.

mcgibbon: Filling "not this slice" with nans is also a fairly complicated operation, since there's no way to "invert" a slice.

The impact of this I think is that `validation_slice` has to be the slice on the reference data. If the user sets that in the class, then the test data passed here will already be made to be the reference data shape by `self.slice_output`.

Ah but it will break/require re-writing the test because the NaNs on the array we use in the model should no longer be applied in the same way.

Maybe for now we say this object only works for reference data with 3 halo points, and we can fix this if we run into this issue? I think it can be done by adding a new keyword argument about the validation origin and then changing the tests and the code inside the validation class. But it's also possible we won't need it for that use case.

fv3core/utils/validation.py
```diff
@@ -0,0 +1,39 @@
...
+
+    @property
+    def validation_slice(self) -> Tuple[slice, ...]:
+        """
+        A slice based on the origin and domain used to initialize this object.
+        """
+        return self._validation_slice
+
+    def set_nans_if_test_mode(self, array):
+        if SelectiveValidation.TEST_MODE:
```
mcgibbon, line 33: Add `if __debug__ and`

Makefile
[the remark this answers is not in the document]
rheacangeo: go ahead and add this, I have heard of stuck tests hitting a few people since the first review of this, we can always change again if the workflow doesn't work for us. I personally seem to run into stuck parallel tests very infrequently, but if it is hitting other people we should do it. Luckily there are only a handful of parallel tests.

fv3core/utils/validation.py
rheacangeo: I don't think we want that necessarily, it's not just debug, but normal tests we want this to happen for

fv3core/utils/validation.py
```diff
@@ -0,0 +1,39 @@
...
+            fv3core.utils.gt4py_utils.device_sync()
```
rheacangeo, line 34: we should not longer need this I believe

fv3core/utils/validation.py
```diff
@@ -0,0 +1,39 @@
...
+            validation_data = np.copy(array[self.validation_slice])
+            array[:] = np.nan
+            fv3core.utils.gt4py_utils.device_sync()
+            array[self.validation_slice] = validation_data
+            fv3core.utils.gt4py_utils.device_sync()
```
rheacangeo, line 39: same question, we've removed the explicit device syncs from most of the model, do we still need them here?

tests/translate/translate_updatedzd.py
```diff
@@ -25,18 +30,43 @@ def __init__(self, grid):
...
             d_sw.get_column_namelist(spec.namelist, self.grid.npz),
             d_sw.k_bounds(),
         )
         inputs["x_area_flux"] = inputs.pop("xfx")
         inputs["y_area_flux"] = inputs.pop("yfx")
-        updatedzd(**inputs)
+        inputs["surface_height"] = inputs.pop("zs")
+        inputs["height"] = inputs.pop("zh")
+        inputs["courant_number_x"] = inputs.pop("crx")
+        inputs["courant_number_y"] = inputs.pop("cry")
```
rheacangeo, line 49: to use different names in the model than what was defined for serialization, we set the 'serialname' attribute to the in_vars dictionary. e.g. 
instead of 
"zs": {},
use
"surface_height": {"serialname": "zs"},

fv3core/utils/validation.py
```diff
@@ -0,0 +1,39 @@
+from typing import Tuple
+
+import numpy as np
+
+import fv3core.utils.gt4py_utils
+
+
+class SelectiveValidation:
```
rheacangeo, line 8: this should probably live in the fv3core/testing or tests/ folder rather than utils, as it is just used for tests

tests/translate/translate_updatedzd.py
```diff
@@ -25,18 +30,43 @@ def __init__(self, grid):
...
+        inputs["ws"] = inputs.pop("wsd")
+        self.updatedzd(**inputs)
         inputs["xfx"] = inputs.pop("x_area_flux")
         inputs["yfx"] = inputs.pop("y_area_flux")
-        return self.slice_output(inputs)
+        inputs["zh"] = inputs.pop("height")
+        inputs["crx"] = inputs.pop("courant_number_x")
+        inputs["cry"] = inputs.pop("courant_number_y")
+        inputs["wsd"] = inputs.pop("ws")
```
rheacangeo, line 57: same as above for all the variables, there is a convention for this use-case

rheacangeo (commented):
On re-review, looks good, just a couple of small points.

rheacangeo (approved):
Great work. I am happy to have this feature and start removing some unnecessary regions! Two discussion points for potential future iterations are 1) moving this test related code out of the mode code and 2) making the validation slice more intuitive to set, as the case here is just lucky the field being applied is a full domain field

mcgibbon:
launch jenkins

PR #394, opened 2021-04-14
update to the latest daint_venv

rheacangeo:


eddie-c-davis (approved):
Yay, both are up to date.

PR #395, opened 2021-04-14
update fv3gs-util to latest

rheacangeo:
## Purpose

We are wondering if the no-sync validation issues are from not having updated to the update the fv3gfs-util that enabled halo updates to work on gt backends


eddie-c-davis (approved):
Looks great!

PR #396, opened 2021-04-15
hotfix for Map1_PPm_2d tests with 54 ranks

rheacangeo:
## Purpose

A recent commit broke Map1_PPM_2d tests for 54 ranks (just test code, so doesn't impact results of the full model), this fixes it.


eddie-c-davis (approved):
Looks good!

PR #405, opened 2021-04-21
Simplify StencilWrapper call logic

eddie-c-davis:
## Purpose

This PR simplifies the `StencilWrapper.__call__` method so that the `stencil_object.run` method is only called in one place and removes the unnecessary domain casts to `gtscript.Shape`.


rheacangeo (approved):
Wow, great find! Nice work.

PR #406, opened 2021-04-21
Hot fix for rebuild flag in FV3StencilObject

eddie-c-davis:
## Purpose

Modifies `FV3StencilObject.__call__` to use `global_config.get_rebuild()` rather than `self.rebuild` to prevent object version from being set before global version.


rheacangeo (approved):
yay!

PR #407, opened 2021-04-21
Ensure global rebuild flag is always used

eddie-c-davis:
## Purpose

This PR sets `self.rebuild` to `global_config.get_rebuild()` in every call to ensure it is always set.

### DIFF
--- a/fv3core/decorators.py
+++ b/fv3core/decorators.py
@@ -394,7 +394,8 @@ class FV3StencilObject(StencilWrapper):
         # Can optimize this by marking stencils that need these
         axis_offsets = fv3core.utils.axis_offsets(spec.grid, origin, domain)
 
-        regenerate_stencil = not self.built or global_config.get_rebuild()
+        self.rebuild = global_config.get_rebuild()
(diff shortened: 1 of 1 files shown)


twicki (approved):

fv3core/decorators.py
```diff
@@ -394,7 +394,8 @@ def __call__(
         # Can optimize this by marking stencils that need these
         axis_offsets = fv3core.utils.axis_offsets(spec.grid, origin, domain)
 
-        regenerate_stencil = not self.built or global_config.get_rebuild()
+        self.rebuild = global_config.get_rebuild()
+        regenerate_stencil = not self.built or self.rebuild
```
jdahm, line 398: Is it important that line 397 be here instead of inside the `__init__` method?

jdahm: This also changes the internal state of the StencilWrapper in the __call__ method which is unintuitive to me.

eddie-c-davis: The issue is that the `@gtstencil` decorated stencils are loaded at import time before the variables are set in `global_config` so they need to be reset here (for now).

PR #408, opened 2021-04-21
Remove uneeded dirs from caches before saving

ofuhrer:
## Purpose

We continually run out of inodes. This PR reduces the number of files and directories (inodes) significantly (down to ~10%) by removing uneeded artifacts of Python (`__pycache__` directory) and GT4Py compilation (`*_pyext_BUILD` directory)

## Code changes:

- Only changes in `.jenkins/actions/run_standalone`

## Notes:

- This is a temporary solution. The better solution is introduce a common decorator to all stencils and make sure we pass the `clean=True` flag when compiling stencils.
- This PR is really hard to test. I suggest merging this in the evening after work hours and then try it.

### DIFF
--- a/.jenkins/actions/run_standalone.sh
+++ b/.jenkins/actions/run_standalone.sh
@@ -100,8 +100,11 @@ fi
 
 # store cache artifacts (and remove caches afterwards)
 if [ "${SAVE_CACHE}" == "true" ] ; then
-    mkdir -p ${CACHE_DIR}
+    echo "Pruning cache to make sure no __pycache__ and *_pyext_BUILD dirs are present"
+    find .gt_cache* -type d -name \*_pyext_BUILD -exec \rm -rf {} \;
+    find .gt_cache* -type d -name __pycache__ -exec \rm -rf {} \;
     echo "Copying GT4Py cache directories to ${CACHE_DIR}"
+    mkdir -p ${CACHE_DIR}
     cp ${ROOT_DIR}/GT4PY_VERSION.txt ${CACHE_DIR}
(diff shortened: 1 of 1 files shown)


twicki (approved):
lgtm. Are we sure this works?

ofuhrer:
> lgtm. Are we sure this works?

I tried it in a standalone test and it worked (did not recompile stencil). I feel like using Jenkins to do the fv3core testing for me, though. Too adventurous?

rheacangeo:
I think letting jenkins tell us after merging this after-hours is a good approach!

twicki:
As you pointed out this will not do any meaningful tests in jenkins (as it only affects a newly generated cache). So in order to test this we have 2 options:
- run this branch locally to create a cache (somewhere other then what we're using), and then launch tests pointing to that cache manually
- merge this in, **delete the existing cache**, wait overnight to have a full rebuild and see if plans picking this updated cache up are fine

ofuhrer:
> As you pointed out this will not do any meaningful tests in jenkins (as it only affects a newly generated cache). So in order to test this we have 2 options:
> 
> * run this branch locally to create a cache (somewhere other then what we're using), and then launch tests pointing to that cache manually
> * merge this in, **delete the existing cache**, wait overnight to have a full rebuild and see if plans picking this updated cache up are fine

Yes. Are you fine with the latter?

And again, having at least some kind of diagnostic when a stencil is being compiled to stdout would be awesome for testing stuff like this. It feels like we are using runtime and/or cProfile as our indicator whether we are compiling stencils. Or am I missing something?

twicki:
I am fine with the latter version

PR #411, opened 2021-04-22
Hotfix: Forgot to add the -prune argument

ofuhrer:


jdahm:
Just a word of warning. I think `-prune` is specific to GNU find, and so might not be supported on other Unix OSs. OK in this case because it's a temporary change.

PR #416, opened 2021-04-26
OOPify Rayleigh Damping (i.e ray_fast)

FlorianDeconinck:
## Purpose

OOPification of the ray_fast stencil: Rayleigh Damping  in acoustic step

## Code changes:

- ray_fast has now RayleighDamping object
- call switched in dyncore
- translate utest adjusted
- namelist doc & general doc from the FV3 namelist doc

## Checklist

### DIFF
--- a/fv3core/stencils/dyn_core.py
+++ b/fv3core/stencils/dyn_core.py
@@ -292,6 +292,9 @@ class AcousticDynamics:
             domain=self.grid.domain_shape_full(),
         )
 
+        if self.namelist.rf_fast:
+            self.rayleigh_damping = ray_fast.RayleighDamping(self.grid, self.namelist)
+
     def __call__(self, state):
         # u, v, w, delz, delp, pt, pe, pk, phis, wsd, omga, ua, va, uc, vc, mfxd,
         # mfyd, cxd, cyd, pkz, peln, q_con, ak, bk, diss_estd, cappa, mdt, n_split,
@@ -601,7 +604,7 @@ class AcousticDynamics:
             if self.namelist.rf_fast:
                 # TODO: Pass through ks, or remove, inconsistent representation vs
                 # Fortran.
-                ray_fast.compute(
+                self.rayleigh_damping(
                     state.u,
                     state.v,
                     state.w,
--- a/fv3core/stencils/dyn_core.py
+++ b/fv3core/stencils/dyn_core.py
@@ -161,10 +161,7 @@ def dyncore_temporaries(shape, namelist, grid):
         grid.full_origin(),
     )
     utils.storage_dict(
-        tmps,
-        ["ws3"],
-        shape[0:2],
-        grid.full_origin()[0:2],
+        tmps, ["ws3"], shape[0:2], grid.full_origin()[0:2],
     )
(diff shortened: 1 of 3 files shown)


fv3core/stencils/dyn_core.py
```diff
@@ -292,6 +292,9 @@ def __init__(
             domain=self.grid.domain_shape_full(),
         )
 
+        if self.namelist.rf_fast:
+            self.rayleigh_damping = ray_fast.RayleighDamping(self.grid, self.namelist)
+
```
rheacangeo, line 297: by convention we have been naming these starting with an underscore to indicate they are private to the class, not a huge deal.

fv3core/stencils/ray_fast.py
```diff
@@ -27,29 +27,28 @@ def dm_layer(rf, dp, wind):
...
-def ray_fast_wind(
+def ray_fast_wind_compute(
     u: FloatField,
     v: FloatField,
     w: FloatField,
     dp: FloatFieldK,
     pfull: FloatFieldK,
     dt: float,
     ptop: float,
+    rf_cutoff: float,
```
rheacangeo, line 38: this can still be an external constant rather than a runtime argument (add to externals in the StencilWrapper initialization and then add here`from___externals__ import rf_cutoff`', but since it doesn't help remove any code compile time, perhaps doesn't matter, but I think it does make sense to keep constant namelist values as externals

fv3core/stencils/ray_fast.py
```diff
@@ -27,29 +27,28 @@ def dm_layer(rf, dp, wind):
...
     rf_cutoff_nudge: float,
     ks: int,
     hydrostatic: bool,
+    tau: float,
```
rheacangeo, line 42: same comment as for rf_cutoff, can be an external to import rather than a runtime argument

fv3core/stencils/ray_fast.py
```diff
@@ -108,34 +107,68 @@ def ray_fast_wind(
...
+    Apply Rayleigh damping (for tau > 0).
+
+    Namelist:
+        - tau [Float]: time scale (in days) for Rayleigh friction applied to horizontal
+                       and vertical winds; lost kinetic energy is converted to heat,
+                       except on nested grids.
+        - rf_cutoff [Float]: pressure below which no Rayleigh damping is applied
+                             if tau > 0.
+
+    Fotran name: rf_fast.
```
rheacangeo, line 125: I think in the Fortran this is actually Ray_fast

rheacangeo (commented):
Looks great, only a couple of minor comments

fv3core/stencils/dyn_core.py
FlorianDeconinck: Good catch

fv3core/stencils/ray_fast.py
FlorianDeconinck: So `rf_cutoff` is also used at realtime to calculate the nudge which depends on another live arg. So I kept it as a class cached var for that, and passed it to the stencils as an external

rheacangeo (approved):
Looks great!

PR #420, opened 2021-04-27
3D Delnflux

oelbert:
## Purpose

Delnflux now operates in 3D over the full k-range

## Code changes:

- delnflux.py: now runs in 3D instead of inside k-splits. This requires conditional copies and conditional corner copies, which are reimplemented here.

- fvtp2d.py: now calls delnflux.compute_delnflux_no_sg directly instead of inside a k-loop

- updatezdz.py: now calls delnflux.compute_no_sg directly instead of inside a k-loop, from 0 to npz+1

- d_sw.py:  now calls delnflux.compute_no_sg directly instead of inside a k-loop

- we have removed the conditional checks on the damping factors in d_sw and updatedzd that determine whether or not to compute for each k-level. Instead we assert that every k-level passes this check, but this may need to be refactored later

## Checklist


fv3core/stencils/delnflux.py
```diff
@@ -86,136 +907,242 @@ def diffusive_damp(
...
+            rebuild=global_config.get_rebuild(),
         )
-    f1_ny = grid.je - grid.js + 1 + 2 * nord
-    f1_nx = grid.ie - grid.is_ + 2 + 2 * nord
-    fx_origin = (grid.is_ - nord, grid.js - nord, kstart)
+        new_copy_stencil(q, d2, origin=origin_d2, domain=domain_d2)
 
-    fx2_order(
-        d2, grid.del6_v, fx2, order=1, origin=fx_origin, domain=(f1_nx, f1_ny, nk)
+    conditional_corner_copy_x = gtscript.stencil(
```
rheacangeo, line 1036: I hate to make you spend more time on this incredible PR that already does so much, but this runtime call to gtscript.stencil will cause a slowdown I believe... we may want to make a DelnFlux class, though we could indeed knowingly slow down master and open a new PR to turn it into a class

eddie-c-davis: I have to agree, we should avoid calling `gtscript.stencil` directly and opt instead for `StencilWrapper`.

fv3core/stencils/delnflux.py
```diff
@@ -1,82 +1,903 @@
...
+import fv3core.utils.global_config as global_config
 import fv3core.utils.gt4py_utils as utils
 from fv3core.decorators import gtstencil
-from fv3core.stencils.basic_operations import copy_stencil
-from fv3core.utils.typing import FloatField, FloatFieldIJ
+from fv3core.utils.grid import axis_offsets
+from fv3core.utils.typing import FloatField, FloatFieldIJ, FloatFieldK
+
+
+def copy_corners_x_nord(q: FloatField):
```
eddie-c-davis, line 23: Tests indicate we need to split these into multiple stencils to prevent template depth errors with the GT backends, something like:
```python
conditional_corner_copy_x_funcs = [
        copy_corners_x_nord0,
        copy_corners_x_nord1,
        copy_corners_x_nord2,
        copy_corners_x_nord3,
    ]
    conditional_corner_copy_x_stencils = [
        StencilWrapper(
            conditional_corner_copy_x_func,
            externals=full_ax_offsets,
        )
        for conditional_corner_copy_x_func in conditional_corner_copy_x_funcs
    ]
...
    for i, conditional_corner_copy_x in enumerate(conditional_corner_copy_x_stencils):
        if nord[i] > 0:
            conditional_corner_copy_x(
                d2, origin=(grid.isd, grid.jsd, 0), domain=(grid.nid, grid.njd, nk)
            )
```

fv3core/stencils/delnflux.py
```diff
@@ -1,82 +1,903 @@
...
+                region[i_start - 2, j_end + 3], region[i_end + 3, j_end + 2]
+            ):
+                q = q[-1, -4, 0]
+            with horizontal(
+                region[i_start - 1, j_end + 3], region[i_end + 3, j_end + 1]
+            ):
+                q = q[-2, -3, 0]
+
+
+def copy_corners_y_nord(q: FloatField):
```
eddie-c-davis, line 324: Same as above...

fv3core/stencils/delnflux.py
```diff
@@ -86,136 +907,242 @@ def diffusive_damp(
...
-    if damp_c <= 1e-4:
+    if (damp_c <= 1e-4).all():
         return fx, fy
-    damp = (damp_c * grid.da_min) ** (nord + 1)
+    elif (damp_c[:-1] <= 1e-4).any():
+        raise NotImplementedError(
+            "damp_c currently must be always greater than 10^-4 for delnflux"
+        )
+
+    damp_3d = utils.make_storage_from_shape(
```
eddie-c-davis, line 952: Repeat cached storages comment...

eddie-c-davis (changes requested):
Epic PR, will done! Just a few small suggestions...

oelbert:
Hmm, all of these changes to fv3gfs-utils and daint-venv were not intended...

eddie-c-davis:
I have made that same mistake at least twice, `git add -u` is dangerous...

eddie-c-davis (approved):
Excellent work!

PR #425, opened 2021-04-28
FxAdv is OOped

rheacangeo:
## Purpose

Another module OOpified, FxAdv is now a class named FiniteVolumeFluxPrep

## Code changes:

- update stencils/fxadv.py to use the model class conventions
- updated call from d_sw
- updated test code

## Checklist

### DIFF
--- a/fv3core/stencils/fxadv.py
+++ b/fv3core/stencils/fxadv.py
@@ -288,7 +288,7 @@ def vt_corners(
 
 
 """
-# Single stencil version
+# Single stencil version to use when possible with gt backends
 def fxadv_stencil(
     cosa_u: FloatFieldIJ,
     cosa_v: FloatFieldIJ,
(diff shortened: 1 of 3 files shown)


fv3core/stencils/fxadv.py
```diff
@@ -368,117 +338,155 @@ def fxadv_fluxes_stencil(
     with computation(PARALLEL), interval(...):
         prod = dt * ut
         with horizontal(region[local_is : local_ie + 2, :]):
-            crx_adv = prod * rdxa[-1, 0] if prod > 0 else prod * rdxa
-            xfx_adv = dy * prod * sin_sg3[-1, 0] if prod > 0 else dy * prod * sin_sg1
+            crx = prod * rdxa[-1, 0] if prod > 0 else prod * rdxa
+            x_area_flux = (
+                dy * prod * sin_sg3[-1, 0] if prod > 0 else dy * prod * sin_sg1
+            )
```
FlorianDeconinck, line 344: Minor:
``` if prod>0:
        crx = prod * rdxa[-1, 0]
        x_area_flux = dy * prod * sin_sg3[-1, 0]
     else:
        crx = prod * rdxa
        x_area_flux = dy * prod * sin_sg1
```
Depending on the codegen that will result in (slightly) less thread-split. Likewise below.

FlorianDeconinck:
Looks good to me.
One minor comment: there's a bunch of commented code that look like deprecated previous implementation. Could be a good time to remove it entirely?

FlorianDeconinck (approved):

fv3core/stencils/fxadv.py
rheacangeo: thanks! yeah this was written back in the days when we *only* could do terniary if statements :)

rheacangeo:
I agree generally we shouldn't leave commented out code -- but doing so for this particular one already saved me a huge amount of time, as I had previously implemented it a couple of different ways. When we realized the regions wasn't going to work with the gt backends, having the other version made it super fast and easy to revert to a version that worked. I didn't need to hunt the history or remember that I had it, it was right in my face. In this case, we still expect to eventually be able to have a single stencil version of this, and I'd like to keep the code here until we have that version, if that's ok with you?

FlorianDeconinck:
Sound fair. Maybe but it directly as a header to the commented out section so that the commented code is motivated and we can remove it later when we are happy with our "final" results

PR #428, opened 2021-04-28
del2cubed is an object

elynnwu:
## Purpose

Make del2cubed a class

## Code changes:

- del2cubed now has HyperdiffusionDamping class
- Remove commented code in original compute function

HyperdiffusionDamping is initialized in both dyn_core and fv_dynamics. A private variable is added in dyn_core to initialize only when the configuration calls for this damping. Could potentially move this so that only fv_dynamics initializes and passes to dyn_core. 

## Checklist

- [x] The names of all the new contributors have been added to CONTRIBUTORS.md


fv3core/stencils/del2cubed.py
```diff
@@ -73,47 +70,98 @@ def corner_fill(q: FloatField):
...
+
+    def __init__(self, grid, qdel: FloatField, km: int):
+        """
+        Args:
+            grid: fv3core grid object
+            qdel (inout): Variable to be filterd
+            km: Number of vertical levels (grid.npz)
+        """
+        self.grid = spec.grid
+        self.origin = (self.grid.isd, self.grid.jsd, 0)
```
rheacangeo, line 86: we usually call this self.grid.full_origin()

fv3core/stencils/del2cubed.py
```diff
@@ -73,47 +70,98 @@ def corner_fill(q: FloatField):
...
+        self.domain = (self.grid.nid, self.grid.njd, km)
```
rheacangeo, line 87: when km is npz, this is usually self.grid.domain_shape_full(), but because it needs to be a custom km, this is perfect as-is.

fv3core/stencils/del2cubed.py
```diff
@@ -73,47 +70,98 @@ def corner_fill(q: FloatField):
...
-    )
-
-    # set up the temporal loop
-    ntimes = min(3, nmax)
-    for n in range(1, ntimes + 1):
-        nt = ntimes - n
-        origin = (grid.is_ - nt, grid.js - nt, 0)
+class HyperdiffusionDamping:
+    """
+    Fortran name is del2cubed
```
rheacangeo, line 75: is del2_cubed in Fortran

fv3core/stencils/del2cubed.py
```diff
@@ -73,47 +70,98 @@ def corner_fill(q: FloatField):
...
+    """
+
+    def __init__(self, grid, qdel: FloatField, km: int):
```
rheacangeo, line 78: I'm glad you are leaving km as an argument -- one might remove it since the current python code only calls it with one possible value, but there are namelist flags in the fortran code that trigger using different values. you may want to assign self._km = km so you can use it in the call method.

fv3core/stencils/del2cubed.py
```diff
@@ -73,47 +70,98 @@ def corner_fill(q: FloatField):
...
+        """
+        Args:
+            grid: fv3core grid object
+            qdel (inout): Variable to be filterd
+            km: Number of vertical levels (grid.npz)
```
rheacangeo, line 83: I don't think you need (grid.npz) because in the full model that won't always be true, just is for our current namelist we have implemented

fv3core/stencils/del2cubed.py
rheacangeo, line 87: tiny point which you can leave as-is if preferred. because you aren't using self.origin or self.domain outside of __init__ these don't need to be class instance attributes, they could just be local to this method as origin and domain. but it doesn't hurt anything to leave it like this.

fv3core/stencils/del2cubed.py
```diff
@@ -73,47 +70,98 @@ def corner_fill(q: FloatField):
...
+        self.grid = spec.grid
+        self.origin = (self.grid.isd, self.grid.jsd, 0)
+        self.domain = (self.grid.nid, self.grid.njd, km)
+        ax_offsets = axis_offsets(spec.grid, self.origin, self.domain)
+        self._fx = utils.make_storage_from_shape(
+            qdel.shape, origin=self.origin, cache_key="del2cubed_fx"
+        )
+        self._fy = utils.make_storage_from_shape(
+            qdel.shape, origin=self.origin, cache_key="del2cubed_fy"
+        )
```
rheacangeo, line 94: you don't need to cache these anymore inside of the __init__, so can remove the cache keys

fv3core/stencils/del2cubed.py
```diff
@@ -73,47 +70,98 @@ def corner_fill(q: FloatField):
...
+
+    def __call__(self, qdel: FloatField, nmax: int, cd: float, km: int):
+        """
+        Perform hyperdiffusion damping/filtering
+
+        Args:
+            qdel (inout): Variable to be filterd
+            nmax: Number of times to apply filtering
+            cd: Damping coeffcient
+            km: Number of vertical levels (grid.npz)
```
rheacangeo, line 116: km doesn't need to both be a runtime argument and an init argument -- you can remove km from here and use self._km

fv3core/stencils/del2cubed.py
```diff
@@ -73,47 +70,98 @@ def corner_fill(q: FloatField):
...
 
-        # Update q values
-        ny = grid.njc + 2 * nt  # (grid.je+nt) - (grid.js-nt) + 1
-        update_q(qdel, grid.rarea, fx, fy, cd, origin=origin, domain=(nx, ny, km))
+
+def compute(qdel: FloatField, nmax: int, cd: float, km: int):
+    hyperdiffusion = utils.cached_stencil_class(HyperdiffusionDamping)(
+        spec.grid, qdel, km
+    )
+    hyperdiffusion(qdel, nmax, cd, km)
```
rheacangeo, line 167: ah very interesting choice to put this in the compute method! This way you don't have to change the test translate code and other code that calls it. What we have been doing with the other ones is remove the compute method alltogether and then initialize the new class where it gets called. The places this gets called, dyn_core and fv_dynamics, are already model objects, and so this initialization of HyperdiffusionDamping can be done in each of their __init__ methods, and we no longer need to cache the object in the runtime explicitly. If you do this, though you will also need to initialize HyperdiffusionDamping in tests/translate/translate_del2cubed.py

rheacangeo (commented):
Awesome work! Lucky you getting one of the rare cases of really needing runtime origin and domains. This looks great. I made a few small comments just related to the conventions we have been adopting for the other OOPifications, overall stellar work.

fv3core/stencils/del2cubed.py
rheacangeo: actually that's incorrect, km is only ever grid.npz so can be removed :)

fv3core/stencils/del2cubed.py
rheacangeo: update -- km is always grid.npz so this can be self.grid.domain_shape_full(),

fv3core/stencils/del2cubed.py
elynnwu: Thanks for pointing this out!

fv3core/stencils/del2cubed.py
elynnwu: Good catch

fv3core/stencils/del2cubed.py
elynnwu: km is now removed and replaced with `grid.npz`

fv3core/stencils/del2cubed.py
elynnwu: Changed to local variables

fv3core/stencils/del2cubed.py
elynnwu: Thanks for pointing this out! HyperdiffusionDamping is now initialized in both dyn_core and fv_dynamics.

rheacangeo (approved):
Great work!

PR #431, opened 2021-04-28
Increase number of timesteps for benchmarking & profiling

ofuhrer:
## Purpose

Increase number of timesteps that we run for the benchmarks. This should have no side-effects since it only affects the performance benchmarking and profiling (cache building is not affected). The performance metrics are already pretty stable, so increasing further will mainly show us whether the system is stable and can sustain a significant number of timesteps without a lot of jitter.


twicki (commented):
lgtm

twicki (approved):

PR #440, opened 2021-04-30
Feature/oop neg adj3

FlorianDeconinck:
## Purpose

OOP-ify neg_adj3

## Code changes:

- Introduce `AdjustNegativeTracerMixingRatio` object
- Fix call in `fv_dynamics` in `dynamics` run file

## Requirements changes:

N/A

## Infrastructure changes:

N/A

## Checklist

- [x ] You have followed the coding standards guidelines established at [Code Review Checklist](https://paper.dropbox.com/doc/Code-Review-Checklist--BD7zigBMAhMZAPkeNENeuU2UAg-IlsYffZgTwyKEylty7NhY).

- [x] The names of all the new contributors have been added to CONTRIBUTORS.md

### DIFF
--- a/fv3core/stencils/fv_dynamics.py
+++ b/fv3core/stencils/fv_dynamics.py
@@ -505,8 +505,8 @@ def fv_dynamics(
         state["atmosphere_hybrid_a_coordinate"],
         state["atmosphere_hybrid_b_coordinate"],
         state["surface_geopotential"],
-        state["specific_humidity"],
-        state["graupel_mixing_ratio"],
+        state["specific_humidity"].data,
+        state["graupel_mixing_ratio"].data,
     )
     dycore.step_dynamics(
         state,
--- a/fv3core/stencils/fv_dynamics.py
+++ b/fv3core/stencils/fv_dynamics.py
@@ -4,7 +4,6 @@ from gt4py.gtscript import PARALLEL, computation, interval, log
 
 import fv3core._config as spec
 import fv3core.stencils.moist_cv as moist_cv
-import fv3core.stencils.neg_adj3 as neg_adj3
 import fv3core.stencils.rayleigh_super as rayleigh_super
 import fv3core.stencils.remapping as lagrangian_to_eulerian
 import fv3core.utils.global_config as global_config
@@ -16,6 +15,7 @@ from fv3core.stencils import c2l_ord
 from fv3core.stencils.basic_operations import copy_stencil
 from fv3core.stencils.del2cubed import HyperdiffusionDamping
 from fv3core.stencils.dyn_core import AcousticDynamics
+from fv3core.stencils.neg_adj3 import AdjustNegativeTracerMixingRatio
 from fv3core.stencils.tracer_2d_1l import Tracer2D1L
 from fv3core.utils.typing import FloatField, FloatFieldK
 
@@ -168,11 +168,16 @@ def post_remap(hyperdiffusion, state, comm, grid, namelist):
         hyperdiffusion(state.omga, namelist.nf_omega, 0.18 * grid.da_min)
 
 
-def wrapup(state, comm: fv3gfs.util.CubedSphereCommunicator, grid):
+def wrapup(
+    state,
+    comm: fv3gfs.util.CubedSphereCommunicator,
+    grid,
+    adjust_stencil: AdjustNegativeTracerMixingRatio,
+):
     if __debug__:
         if grid.rank == 0:
             print("Neg Adj 3")
-    neg_adj3.compute(
+    adjust_stencil(
         state.qvapor,
         state.qliquid,
         state.qrain,
@@ -296,6 +301,8 @@ class DynamicalCore:
         ak: fv3gfs.util.Quantity,
         bk: fv3gfs.util.Quantity,
         phis: fv3gfs.util.Quantity,
+        qvapor: fv3gfs.util.Quantity,
+        qgraupel: fv3gfs.util.Quantity,
     ):
         """
         Args:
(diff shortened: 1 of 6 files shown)


fv3core/stencils/fv_dynamics.py
```diff
@@ -447,7 +457,7 @@ def _compute(
                         self.namelist,
                     )
                 state.wsd[:] = state.wsd_3d[:, :, 0]
-        wrapup(state, self.comm, self.grid)
+        wrapup(state, self.comm, self.grid, self._adjust_tracer_mixing_ratio)
```
eddie-c-davis, line 460: Passing stencil object as a parameter, nice touch...

tests/translate/translate_neg_adj3.py
```diff
@@ -33,3 +33,23 @@ def __init__(self, grid):
         }
         for qvar in utils.tracer_variables:
             self.ignore_near_zero_errors[qvar] = True
+
+    def compute(self, inputs):
+        self.make_storage_data_input_vars(inputs)
+        compute_fn = AdjustNegativeTracerMixingRatio(
```
eddie-c-davis, line 39: Not a big deal, but curious why not just set `self.compute_func` to `AdjustNegativeTracerMixingRatio(...)`?

profiler/nsys_id_kernel.py
```diff
@@ -1,5 +1,5 @@
-import sqlite3
 import re
+import sqlite3
```
eddie-c-davis, line 2: Ah, `pre-commit`...

eddie-c-davis (approved):
One question, but otherwise looks great.

fv3core/stencils/neg_adj3.py
```diff
@@ -299,70 +298,80 @@ def compute(qvapor, qliquid, qrain, qsnow, qice, qgraupel, qcld, pt, delp, delz,
...
+        namelist,
+        qvapor,
+        qgraupel,
+    ):
+
+        shape_ij = qgraupel.shape[0:2]
+        self._sum1 = utils.make_storage_from_shape(shape_ij, origin=(0, 0))
+        self._sum2 = utils.make_storage_from_shape(shape_ij, origin=(0, 0))
+        self._upper_fix = utils.make_storage_from_shape(qvapor.shape, origin=(0, 0, 0))
+        self._lower_fix = utils.make_storage_from_shape(qvapor.shape, origin=(0, 0, 0))
```
rheacangeo, line 314: I don't think upper_fix and lower_fix are needed as storages outside of the stencil. I think they had been removed awhile ago, but I think slipped back in

fv3core/stencils/neg_adj3.py
```diff
@@ -299,70 +298,80 @@ def compute(qvapor, qliquid, qrain, qsnow, qice, qgraupel, qcld, pt, delp, delz,
...
+            qsnow,
+            qice,
+            qgraupel,
+            self._lv00,
+            self._d0_vap,
+        )
+        self._fillq(qgraupel, delp, self._sum1, self._sum2)
+        self._fillq(qrain, delp, self._sum1, self._sum2)
+        self._fix_water_vapor_down(qvapor, delp, self._upper_fix, self._lower_fix)
+        self._fix_neg_cloud(delp, qcld)
```
rheacangeo, line 377: These could be all one stencil. Chris Kung I think had previously refactored it to be so but we were worried about losing the physical meaning so never merged it. No need to change it here, but something to keep in mind if we need to reduce the stencil call count, these can be joined.

tests/translate/translate_neg_adj3.py
rheacangeo: His init routine relied on model data, qvapor and qgraupel, which isn't known at init time of the translate class. Otherwise I'd suggest the same thing!

fv3core/stencils/fv_dynamics.py
```diff
@@ -495,6 +505,8 @@ def fv_dynamics(
         state["atmosphere_hybrid_a_coordinate"],
         state["atmosphere_hybrid_b_coordinate"],
         state["surface_geopotential"],
+        state["specific_humidity"].data,
+        state["graupel_mixing_ratio"].data,
```
rheacangeo, line 509: one of these days we'll make it more seamless to always use quantity or gt4pystorage

rheacangeo (approved):
Great job! I have one small suggestion of temporary fields that can be removed. Well on our way!

profiler/nsys_id_kernel.py
FlorianDeconinck: The best kind of commit ;)

fv3core/stencils/neg_adj3.py
FlorianDeconinck: Well to keep the physical meaning you can keep that structure as `@gtstencil.function` into a merged stencil

fv3core/stencils/fv_dynamics.py
FlorianDeconinck: It seems like an implementation detail that should be removed from users view indeed.

fv3core/stencils/neg_adj3.py
FlorianDeconinck: So you would make them a tmp ? They are accessed as a 3d field (accessing offset in z) wouldn't that be a slow tmp ?

rheacangeo: yeah let gt4py allocate them rather than us. I doubt there would be much performance impact from it. we've wanted to reduce the number of temporary fields we allocate in the python space, and how many fields need to be passed to stencils. The exception is our workarounds for regions, but I thought in general it was still true the less we can avoid declaring temporary storages, the better.

fv3core/stencils/neg_adj3.py
rheacangeo: no, you can't have with computations in a function. you need separate computations in this algorithm.

fv3core/stencils/neg_adj3.py
FlorianDeconinck: Sounds good, done.

PR #441, opened 2021-04-30
Compile asserts and gtscript updates

jdahm:
## Purpose

Incorporate latest gt4py master into the develop branch. Changes:

1. `Field[float, IJK]` becomes `Field[IJK, float]`
2. `external_assert` becomes `compile_assert`

Blocked by: #442.


GT4PY_VERSION.txt
```diff
@@ -1 +1 @@
-v26
+3053f82e5ab1acc6108b739df9cc6cb178ae78b3
```
rheacangeo, line 1: we both know this, but just want to add the comment for anyone else's benefit, this is meant to be temporary until a new gt4py tag is generated, at which point we will update fv3core master to that

fv3core/decorators.py
```diff
@@ -97,518 +63,221 @@ def get_namespace(arg_specs, state):
...
+                    )
+                    field_origins[name] = (
+                        *field_origin,
+                        *((0,) * len(field_info.data_dims)),
+                    )
+            elif all_origin is not None:
+                field_origins[name] = (
+                    *gt4py.utils.filter_mask(all_origin, field_info.domain_mask),
+                    *((0,) * len(field_info.data_dims)),
+                )
```
rheacangeo, line 200: would you mind describing in the PR description what this change is about, what motivates it and what it fixes etc? comparing the diff from jeremy's PR, it is still not clear to me

rheacangeo (commented):
I think these changes look good. I would love some more clarity on the changes to the field origin computations, but once jeremy's pr is in master I am willing to accept this one.

rheacangeo (approved):

fv3core/decorators.py
jdahm: For now I tried removing this change again. We will need it at some point soon when we introduce higher dimensional fields, but for now I think it's ok...

PR #442, opened 2021-04-30
Simplify stencil wrapper classes and caching logic

mcgibbon:
## Purpose

This PR simplifies the classes used for wrapping stencils, removing features we no longer use and adding tests for the features we do use.

It also adds a requirement that stencil origins must be explicitly defined, for example if an input has a singleton dimension its origin must be zero for that dimension. Code which relied on gt4py having a side-effect of changing this origin to `shape[i] - 1` has been updated to explicitly pass the origin, or no longer use a singleton-dimension array. In other words, when we need the origin to be different for different inputs, we now explicitly pass those origins.

## Code changes:

- Added `StencilConfig` class which contains all configuration used to initialize a gt4py stencil
- stencils now get all global gt4py configuration from a new method global_config.get_stencil_config(), instead of a separate getter for each configuration option
- "stencil report" functionality which was used for initial validation but has not been used in months is removed
- when using gtstencil as a decorator, the syntax is now `@gtstencil` with no parentheses. To decorate with explicit externals, a separate assignment line must be used.
- The error produced when trying to use a stencil which has fields come after parameters is now more informative

## Checklist


fv3core/stencils/dyn_core.py
```diff
@@ -238,7 +238,7 @@ def __init__(
                 self.grid.domain_shape_full(add=(1, 1, 1)), self.grid.full_origin()
             )
 
-            dp_ref_stencil = StencilWrapper(
+            dp_ref_stencil = FrozenStencil(
```
mcgibbon, line 241: I kind of like making FrozenStencil explicitly instead of relying on gtstencil, but I'm interested in hearing what reviewers think. I'm happy to switch it as well.

fv3core/decorators.py
```diff
@@ -96,502 +63,177 @@ def get_namespace(arg_specs, state):
...
+            if field_info is not None:
+                field_origin_list = []
+                for ax in field_info.axes:
+                    origin_index = {"I": 0, "J": 1, "K": 2}[ax]
+                    field_origin_list.append(origin_tuple[origin_index])
+                field_origin = tuple(field_origin_list)
+            else:
+                field_origin = origin_tuple
+            field_origins[field_name] = field_origin
+    return field_origins
```
jdahm, line 178: If `stencil_object._make_origin_dict` is called before this call (or `stencil_object` is passed in and called), then the following code can replace this and will be very close to what gt4py does. 

Note that code like `origin_index = {"I": 0, "J": 1, "K": 2}[ax]` is not guaranteed to aways work.

```python
    field_origins = stencil_object._make_origin_dict(origin)  # optional: if not done before call
    all_origin = field_origins.get("_all_", None)

    for field_name, field_info in field_info_mapping.items():
        if field_info is not None:
            field_origin = field_origins.get(name, None)
            if field_origin is not None:
                field_origin_ndim = len(field_origin)
                if field_origin_ndim != field_info.ndim:
                    assert field_origin_ndim == field_info.domain_ndim, (
                        f"Invalid origin specification ({field_origin})"
                        f" for '{name}' field."
                    )
                    field_origins[name] = (
                        *field_origin,
                        *((0,) * len(field_info.data_dims)),
                    )
            elif all_origin is not None:
                field_origins[name] = (
                    *gt_utils.filter_mask(all_origin, field_info.domain_mask),
                    *((0,) * len(field_info.data_dims)),
                )
            else:
                raise ValueError(f"Missing origin for {name}")
    return field_origins
```

jdahm: Actually, now that I think about it I believe this requires the latest gt4py, so hold off for now and I'll change it after.

tests/translate/translate_d_sw.py
```diff
@@ -68,23 +68,24 @@ def compute(self, inputs):
...
+def ubke(
+    uc: FloatField,
+    vc: FloatField,
+    cosa: FloatFieldIJ,
+    rsina: FloatFieldIJ,
+    ut: FloatField,
+    ub: FloatField,
+    dt4: float,
+    dt5: float,
+):
```
jdahm, line 81: Good find! Can you also switch these from using a decorator to the functional version inside the `__init__`? Then these won't be generated as pytest scans through the file.

jdahm:
I'm not as familiar with the test code changes, so I'll leave those to @rheacangeo to review. This is definitely going in the right direction with the removal of the storages from the origin computation. There will be more changes to that coming soon when updating to a later gt4py version.

tests/translate/translate_d_sw.py
mcgibbon: Done when merging with master.

rheacangeo (commented):
Wow what a PR. I have to say the changes to the decorator a quite dramatic and I can't really tell whether it is ok to remove all of the behavior that you removed by looking at the code, but given the test still pass, that is a good sign. I think @eddie-c-davis should make sure to check on this before merging, as I think many of the removed features he had added. Have you tried to run run_on_daint with this branch? 
Like I mentioned a few different ways, I thought we wanted to eliminate the use of the decorator gtstencil as part of our current refactors, but this reintroduces it as a parent construct to the FrozenStencil. I think would be nice to have a parallel class to FrozenStencil to be used when not specifying the origin and domain in init. But either way, hopefully in a day near us soon, everything will be FrozenStencils and there's no longer a strange mix of conventions. This does help align our older convention with the new one, which is nice. I like making the global config an object, that is useful.

fv3core/decorators.py
```diff
@@ -97,518 +63,223 @@ def get_namespace(arg_specs, state):
...
-            self.run(domain, *args, **kwargs)
-            self.is_cached = True
-
-    def run(self, domain: Optional[Index3D], *args, **kwargs):
-        """Runs the stencil with the provided domain and args."""
-        kwargs = self._process_kwargs(domain, *args, **kwargs)
-        self.stencil_object.run(**kwargs)
-        if "cuda" in self.backend:
-            for write_field in self.write_fields:
-                kwargs[write_field]._set_device_modified()
```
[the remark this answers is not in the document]
eddie-c-davis: Thank you @rheacangeo , it looks like the tests pass without this code, but the MPI Recv time increases in the absence of the intermediate syncs,
```
master:   26.061    0.000 {method 'Recv' of 'mpi4py.MPI.Comm' objects}
p442:    224.419    0.001 {method 'Recv' of 'mpi4py.MPI.Comm' objects}
```

mcgibbon (commented):
Just comments for now, about to implement these changes.

rheacangeo (approved):
Ok looks good to me! might be worthwhile to check the performance using run_on_daint before merging. I was wondering if there are any other removed features we might need, but nothing stands out. Thanks for bringing back in the write_fields behavior

eddie-c-davis (approved):
Looks good to me.

PR #443, opened 2021-05-01
Cleanup dead code

rheacangeo:
## Purpose

In the initial port we had a slightly different namelist and we implemented a few namelist options that we no longer use since we shifted to a namelist that is more similar to what the ML team is using. This code initially validated, but we no longer have that data (and it would be way out of date with out latest fortran code environment anyways and would not validate anymore) and this untested dead code has been sitting around for over a year now. Although it is hard to delete the effort that went into these, we still need to focus on making one namelist configuration faster and better before expanding the other options. When we do so, we'll want to generate a variety of other datasets based on different namelist options and implement more things. But we are actively drastically changing this code's conventions and we shouldn't keep around code we can't test refactors on. We could make the data right now and support it as well, but this is not the current focus. 
If this misses any dead code, we can always remove it when we turn it into class objects. 

## Code changes:

- slash and burn


jdahm (approved):
Looks good to me!

PR #452, opened 2021-05-04
OOP-ify Fast Corners

twicki:
## Purpose

Copy Corners showed up as one of the slowest stencils we run. This was due to it creating a kernel call for each region executed. This PR speeds up the corner computation. 
As this functionality  is used in three spots throughout the code, we create an object to re-use the logic.

## Code changes:

- Create a `CopyCorners` class that handles the logic of copying to the temporary in the separate kernel call
- Applies the object in all two spots (ignoring delnflux as is is currently in the process of being refactored)

## Checklist


fv3core/stencils/delnflux.py
```diff
@@ -131,9 +131,6 @@ def compute_delnflux_no_sg(
     else:
         # TODO: To join these stencils you need to overcompute, making the edges
         # 'wrong', but not actually used, separating now for comparison sanity.
-
-        # diffusive_damp(fx, fx2, fy, fy2, mass, damp, origin=diffuse_origin,
-        # domain=(grid.nic + 1, grid.njc + 1, nk))
```
rheacangeo, line 136: I think you were cleaning up and didn't need to delete this, but seems like it would minimally impact the delnflux refactor so is fine to leave.

rheacangeo (approved):
looks great!

PR #455, opened 2021-05-04
PK3Halo is an object

elynnwu:
## Purpose

pk3_halo is an object

## Code changes:

- `PK3Halo` is a class in `pk3_halo` 
Note that even though this is a single stencil, it is still made into an object because we need to make storage for this.

- Changed how `PK3Halo` is called in `dyn_core`

## Checklist


rheacangeo (approved):
Looks great! I still think this makes more sense as just a FrozenStencil rather than a separate object, I but can accept we don't need the separate temporary 2d variable floating around dyn_core confusing people. It only exists because to write to a 2d field in a stencil it needs to be an API field. So someday can be eliminated, and is easier to do inside of an object where its scope is clear. 
The diff looks great!

PR #456, opened 2021-05-04
Object Orientation of Saturation Adjustment

gmao-ckung:
## Purpose

This PR creates an object out of the saturation adjustment stencil.

## Code changes:

`saturation_adjustment.py` : Created `SatAdjust3d` class and moved the relevant parts from the previous `compute` routine into `__init__` and `__call__`.

`translate_satadjust3d.py` : `self.compute_func` calls `SatAdjust3d()`

`remapping_part2.py` : Using `utils.cached_stencil_class` to call the OO version of `satadjust`


fv3core/stencils/remapping_part2.py
```diff
@@ -87,6 +87,10 @@ def compute(
     consv: float,
     do_adiabatic_init: bool,
 ):
+    saturation_adjustment = utils.cached_stencil_class(SatAdjust3d)(
+        cache_key="satadjust3d"
```
rheacangeo, line 91: perfect! could even put this inside of the if condition so it only gets compiled if the namelist says it is used. but this will get refactored into the init method of the parent object in a future pr anyways

fv3core/stencils/saturation_adjustment.py
```diff
@@ -886,103 +885,105 @@ def satadjust(
...
-    fac_i2s = 1.0 - math.exp(-mdt / spec.namelist.tau_i2s)
-    fac_v2l = 1.0 - math.exp(-sdt / spec.namelist.tau_v2l)
-    fac_r2g = 1.0 - math.exp(-mdt / spec.namelist.tau_r2g)
-    fac_l2r = 1.0 - math.exp(-mdt / spec.namelist.tau_l2r)
-
-    fac_l2v = 1.0 - math.exp(-sdt / spec.namelist.tau_l2v)
-    fac_l2v = min(spec.namelist.sat_adj0, fac_l2v)
-
-    fac_imlt = 1.0 - math.exp(-sdt / spec.namelist.tau_imlt)
-    fac_smlt = 1.0 - math.exp(-mdt / spec.namelist.tau_smlt)
```
rheacangeo, line 929: I want to double check with folks, on whether we want these calculations to happen inside of stencils. in theory, yes we do, so probably fine to do. but if there's any lag in toolchain. logic that accidentally makes 3d fields out of these scalars, we might want to reconsider.

fv3core/stencils/saturation_adjustment.py
```diff
@@ -886,103 +885,105 @@ def satadjust(
...
-        spec.namelist.rad_rain,
-        spec.namelist.rad_graupel,
-        spec.namelist.tintqs,
-        origin=origin,
-        domain=domain,
-    )
+class SatAdjust3d:
+    def __init__(self):
+        self.grid = spec.grid
+        self._satadjust_stencil = gtstencil(func=satadjust)
```
rheacangeo, line 891: we can use FrozenStencil here and pass the origin and domain at this point so we don't incur runtime pythong overhead processing the origin and domain.

fv3core/stencils/saturation_adjustment.py
```diff
@@ -886,103 +885,105 @@ def satadjust(
...
+        pkz: FloatField,
+        cappa: FloatField,
+        r_vir: float,
+        mdt: float,
+        fast_mp_consv: bool,
+        last_step: bool,
+        akap: float,
+        kmp: int,
+    ):
+        hydrostatic = spec.namelist.hydrostatic
```
rheacangeo, line 918: this works! in our new conventions, which maybe you haven't heard all of (there are many), we are moving towards relying less on global state and for the namelist variables, adding the specific variables needed in the __Init__ method. e.g. self.hydrostatic = namelist.hydrostatic (passing the namelist object in as an init argument). There are a lot of namelist values though, so might be ok for now to say self.namelist  = namelist and use self.namelist.hydrostatic rather than making a variable for every single one. 
similarly there are a few namelist variables inside the stencil (e.g. sat_adj0, ql_gen) that should also be passed that way rather than referring to spec.namelist directly. conventions always changing :)

fv3core/stencils/saturation_adjustment.py
```diff
@@ -886,103 +885,105 @@ def satadjust(
...
+        fac_i2s = 1.0 - math.exp(-mdt / spec.namelist.tau_i2s)
+        fac_v2l = 1.0 - math.exp(-sdt / spec.namelist.tau_v2l)
+        fac_r2g = 1.0 - math.exp(-mdt / spec.namelist.tau_r2g)
+        fac_l2r = 1.0 - math.exp(-mdt / spec.namelist.tau_l2r)
+
+        fac_l2v = 1.0 - math.exp(-sdt / spec.namelist.tau_l2v)
+        fac_l2v = min(spec.namelist.sat_adj0, fac_l2v)
+
+        fac_imlt = 1.0 - math.exp(-sdt / spec.namelist.tau_imlt)
+        fac_smlt = 1.0 - math.exp(-mdt / spec.namelist.tau_smlt)
```
rheacangeo, line 930: same as above with all of the namelist values.

fv3core/stencils/saturation_adjustment.py
```diff
@@ -886,103 +885,105 @@ def satadjust(
...
+            lv00,
+            fac_v2l,
+            fac_l2v,
+            last_step,
+            spec.namelist.rad_snow,
+            spec.namelist.rad_rain,
+            spec.namelist.rad_graupel,
+            spec.namelist.tintqs,
+            origin=(self.grid.is_, self.grid.js, kmp),
+            domain=(self.grid.nic, self.grid.njc, (self.grid.npz - kmp)),
```
rheacangeo, line 988: these can be set in init using FrozenStencil, woot!

fv3core/stencils/saturation_adjustment.py
```diff
@@ -886,103 +885,105 @@ def satadjust(
...
+            fac_imlt,
+            d0_vap,
+            lv00,
+            fac_v2l,
+            fac_l2v,
+            last_step,
+            spec.namelist.rad_snow,
+            spec.namelist.rad_rain,
+            spec.namelist.rad_graupel,
+            spec.namelist.tintqs,
```
rheacangeo, line 986: instead of having these be runtime arguments, we should pass these as externals in the initilaization of the stencil. then in the stencil you can do `from __exernals__ import rad_snow, rad_rain` etc. and use them directly there that way.

rheacangeo (commented):
Looks awesome! just a couple of minor changes needed to adjust to our ever-changing conventions for these refactors.

fv3core/stencils/saturation_adjustment.py
rheacangeo: ah currently origin and domain are runtime specified. we should change that in the near future, but don't have to for this PR

rheacangeo (approved):

PR #460, opened 2021-05-05
Refine requirements

twicki:
## Purpose

This moves requirements for the different purposes into the `requirements` folder.
The default behaviour is unchanged as the base requirement file includes what it included before. 
We also have a specific requirement file for out Piz Daint development that pins the MPI4Py version so it does not need to rebuild that bare metal every time

## Checklist

- [X] All relevant documentation has been updated or added (e.g. README, CONTRIBUTING docs)


requirements.txt
```diff
@@ -1,27 +1 @@
...
-pytest-regressions
-gitpython
-
-# GT4Py Tests
-clang-format
-factory-boy
-pytest-factoryboy
-pytest-cache
-pytest-cov
+-r requirements/requirements_container.txt
```
rheacangeo, line 1: it's a little confusing for the central requirements.txt to be specifically for the container, do we need this file at all?

rheacangeo (commented):
Awesome! I like moving the requirements files to their own folders and the daint_venv simplification. I am a little confused on the daint_venv diffs, this change to its install.sh here https://github.com/VulcanClimateModeling/daint_venv/compare/e26d69ff3dc3cc700ff8d45250ac5a3bed0d25ac...bbf6e829f7128a31d6e54a33e777c5eb09235cf2#diff-043df5bdbf6639d7a77e1d44c5226fd7371e5259a1e4df3a0dd5d64c30dca44f doesn't seem to match the open PR in daint_venv here: https://github.com/VulcanClimateModeling/daint_venv/pull/18/files#diff-043df5bdbf6639d7a77e1d44c5226fd7371e5259a1e4df3a0dd5d64c30dca44f

I approve, but before merging we should make sure the relevant daint_venv PR is merged into its master and the submodule in this PR is updated to reflect it.

requirements/requirements_container.txt
```diff
@@ -0,0 +1,2 @@
+-r requirements_base.txt
+git+https://github.com/mpi4py/mpi4py.git@aac3d8f2a56f3d74a75ad32ac0554d63e7ef90ab
```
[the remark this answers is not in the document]
twicki: I think this is an interesting question. Does it have to match?

requirements.txt
twicki: Well, it is more that the central file should be the one that also works in most cases. And the thing we do in containers is what works locally, it is just on daint - with modules etc, we need something differently. I am open to renaming it if you think the intent gets clearer?

twicki: and going this route the "default" behaivour we had so far does not change

requirements/requirements_container.txt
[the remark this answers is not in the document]
rheacangeo: Not necessarily, in some cases might not be able to, but the less version divergence we have in different environments, the less confusing it is to support multiple environments.

requirements.txt
rheacangeo: ok we can leave it then!

rheacangeo (approved):
Not sure why it is not passing on docker, but this looks great

requirements/requirements_container.txt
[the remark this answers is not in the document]
FlorianDeconinck: I don't think this is too much of an issue. Remember that this is temporary because we are using a feature in the non-released `mpi4py`. E.g we don't need this commit as much as "this commit at least'. Locking it in a commit that works is good practice but this should go away if they ever decide to release 3.1.0

PR #462, opened 2021-05-05
make pfull init time

rheacangeo:
## Purpose

Remapping computes a k origin of kmp using pfull in the runtime. This could be resolved by pushing the pfull comparison as a runtime if condition inside of stencils, but this will not look great. pfull actually only depends on things known at the init time of the FVDynamics objects that don't change, so it also can be computed in init. Then when we make remapping an object, we can make the computation of kmp also init time, and use FrozenStencil for stencils that use kmp as an origin. 

## Code changes:

- Updated fv_dynamics and dyn_core to use init time pfull. 

## Checklist

### DIFF
--- a/fv3core/stencils/dyn_core.py
+++ b/fv3core/stencils/dyn_core.py
@@ -231,6 +231,7 @@ class AcousticDynamics:
         namelist,
         ak: FloatFieldK,
         bk: FloatFieldK,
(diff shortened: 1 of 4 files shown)


gmao-ckung:
Looks good!  Thanks for getting this together quickly!

gmao-ckung (approved):

PR #463, opened 2021-05-05
fix bug introduced in 54 rank sequential tests

rheacangeo:
## Purpose

Fix the issues caught on the daily cron that caused failures in c54ranks. One issue was using grid.npy, which is tile global, instead of grid.njd, the local array range. The other was caching the non frozen stencils, when running the sequential tests, the cache survives the move to a new rank (used to be cleared, but that was removed in a recent PR). With 54 ranks, this changes the stencil whether or not a cell is on a tile edge. So, added the rank to the hash key as a short-term solution, knowing eventually everything will be a FrozenStencil and we don't need a cache of non-frozen ones. 

## Code changes:

- change occurrences of grid.npy to grid.njd in translate classes j slicing
- update fv3core/decorators to use the rank as part of the non-frzone stencil cache hash

## Checklist

### DIFF
--- a/fv3core/decorators.py
+++ b/fv3core/decorators.py
@@ -267,7 +267,7 @@ def get_non_frozen_stencil(func, externals) -> Callable[..., None]:
         else:
             origin_key = origin
             origin_tuple = origin
-        key: Hashable = (origin_key, domain, stencil_config)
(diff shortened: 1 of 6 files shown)


mcgibbon (approved):
LGTM! Thanks for fixing this!

PR #466, opened 2021-05-05
Lagrangian contributions using while loop

jdahm:
## Purpose

Using new `while` loop gt4py feature.


fv3core/stencils/map_single.py
```diff
@@ -7,172 +7,70 @@
 from fv3core.decorators import FrozenStencil, gtstencil
 from fv3core.stencils.basic_operations import copy_defn
 from fv3core.stencils.remap_profile import RemapProfile
-from fv3core.utils.typing import FloatField, FloatFieldIJ
+from fv3core.utils.typing import FloatField, IntFieldIJ
 
 
+r3 = 1.0 / 3.0
+r23 = 2.0 / 3.0
```
jdahm, line 14: Remove these before merging.

rheacangeo:
This is really freaking exciting!

rheacangeo (approved):
Probably someone who has battled this algorith more, e.g. @oelbert , @eddie-c-davis should also review it, but this looks great to me!

fv3core/stencils/map_single.py
```diff
@@ -181,30 +75,49 @@ class MapSingle:
...
 
         self.dp1 = utils.make_storage_from_shape(shape, origin=origin)
         self.q4_1 = utils.make_storage_from_shape(shape, origin=origin)
         self.q4_2 = utils.make_storage_from_shape(shape, origin=origin)
         self.q4_3 = utils.make_storage_from_shape(shape, origin=origin)
         self.q4_4 = utils.make_storage_from_shape(shape, origin=origin)
 
-        self.i_extent = i2 - i1 + 1
-        self.j_extent = j2 - j1 + 1
+        self.lev = utils.make_storage_from_shape(
```
eddie-c-davis, line 87: Temporary fields do not need to be cached in OOP-ified stencils.

eddie-c-davis (approved):
One minor comment, great to see this in action!

fv3core/stencils/map_single.py
jdahm: But we don't want this re-allocated on each call, right? I'm not opposed to switching it to a function-local temporary, and I suppose right now everything is cached anyway, but this might not be how it works in the future. I'll leave the changes to others who are more familiar with fv3core's up-to-date standards.

PR #467, opened 2021-05-06
Reduce configuration-related overhead of gtstencil decorator

mcgibbon:
## Purpose

The initial implementation of the gtstencil decorator refactor has an expensive to compute key, and gets global configuration on each call instead of once (as was documented). This PR greatly reduces the expense of computing the key, and grabs global config the first time a gtstencil-decorated function is called.

## Code changes:

- try: except is used in favor of isinstance in StencilConfig's equality operator to check if the other argument is hashable
- stencil_config is retrieved only the first time a gtstencil-decorated function is called, instead of every time, and is removed from the cache key since it is now constant for each call

## Checklist


rheacangeo:
Totally great to do something that will give us a speedup, but we are so close on removing the unfrozen stencil everywhere, attention might be well spent finishing those off?

rheacangeo:
Ah, I see get_stencil_config at 5% in the profile, gotchya, now I get the motivation.

fv3core/decorators.py
```diff
@@ -260,14 +277,20 @@ def decorated(
...
+            origin_key: Hashable = tuple(
+                sorted(origin.items(), key=lambda x: x[0])  # type: ignore
+            )
+            origin_tuple: Tuple[int, ...] = origin["_all_"]  # type: ignore
+        except AttributeError:  # assume origin is a tuple
             origin_key = origin
-            origin_tuple = origin
-        key: Hashable = (origin_key, domain, stencil_config)
+            origin_tuple = cast(Index3D, origin)
+        key: Hashable = (origin_key, domain)
```
rheacangeo, line 293: can you be sure to incorporate the spec.grid.rank in the key so 54 rank sequential tests will still work? I assume you will when you update to master.

mcgibbon: Yeah, I'll keep an eye out for that when I resolve merge conflicts.

fv3core/decorators.py
```diff
@@ -261,17 +278,23 @@ def decorated(
...
-        stencil_config = global_config.get_stencil_config()
-        if isinstance(origin, Mapping):
-            origin_key: Hashable = tuple(sorted(origin.items(), key=lambda x: x[0]))
-            origin_tuple: Tuple[int, ...] = origin["_all_"]
-        else:
+        try:
+            stencil_config = stencil_config_holder[0]
+        except IndexError:
+            stencil_config = global_config.get_stencil_config()
+            stencil_config_holder.append(stencil_config)
```
jdahm, line 285: Could `stencil_config_holder` get initialized with `global_config.get_stencil_config()` as its only element?

mcgibbon: Unfortunately not, the reason it's a holder and not the config itself is that the configuration isn't available until the first call happens.

twicki (approved):
lgtm

jdahm (approved):
I wish we would not have to try-except in regular code like this, but alas this should get taken care of at some point in the future and is orthogonal to this work. Looks good to me!

PR #479, opened 2021-05-07
Reapply OOP remapping part1 plus small fix

rheacangeo:
## Purpose

reapply the awesome OOP of remapping_part1 (#478), with all uses of gtstencil removed in remapping_part1.py


oelbert:
Might change Rapply to Reapply ;-)

PR #484, opened 2021-05-11
Moist CV and FVDynamics stencils as FrozenStencils 

rheacangeo:
## Purpose

Using unfrozen stencils is confusing with 3 styles of stencils. In our refactoring cycle, here we update/remove the stencils decorated with @gtstencil  left in the moist_cv file, as well as those in fv_dynamics and xppm/yppm. 

## Code changes:

- fv3core/stencils/moist_cv has frozen all @gtstencil 's and the compute methods that called them. computing pkz, pt and fv_setup were changed this way. 
- tests that called removes moist_cv stencils are updated translate_moistcvpluspkz_2d.py, ranslate_moistcvpluspt_2d.py to run the single j directly rather than a method in the model code  that would either run a single j or 3d depending on what was passed. 
- fv_dynamics adjust_pt and set_omega stencils are frozen
- fv_dynamics call to fv_setup uses a FrozenStencil
- added asserts to fv_dynamics to make sure supported namelist options of moist_phys and nwat are used
- fv_setup test is removed because there is no data 
- Remapping_part1 calls the compute_pt FrozenStencil
- XPPM and YPPM don't use gtstencil, pert_ppm code was removed because it is not called. 

## Checklist


rheacangeo:
launch jenkins

tests/translate/translate_moistcvpluspkz_2d.py
```diff
@@ -52,18 +56,16 @@ def __init__(self, grid):
...
 
-    def compute(self, inputs):
-        self.make_storage_data_input_vars(inputs)
-        inputs["j_2d"] = self.grid.global_to_local_y(
-            inputs["j_2d"] + TranslateGrid.fpy_model_index_offset
+    def compute_from_storage(self, inputs):
+        moist_cv_pkz = FrozenStencil(
+            moist_cv.moist_pkz,
+            origin=(self.grid.is_, self.grid.js, 0),
+            domain=(self.grid.nic, 1, self.grid.npz),
```
gmao-ckung, line 63: I'm curious why the Frozen Stencil is here in compute but not in init.

gmao-ckung (approved):
Looks good!  I'm getting a better sense of the conventions that the group is using for the OOness.

tests/translate/translate_moistcvpluspkz_2d.py
rheacangeo: good catch! it can be in init, I'll move it there. When the init relies on serialized data it needs to happen in the compute function, but you are right, this is not needed here.

rheacangeo: ah, it was there because initially I was going to use j_2d to set the origin. but if we end up returning to the fortran code to write out a different j, we should probably do the step we previously decided not do, of updating serialbox to accunulate 3d variables in the j direction like it does for the k direction when k blocked. Anyways, here decided to just use the jslice value we picked to make it simpler. But then forgot to move the FrozenStencil init back to the translate init.

gmao-ckung: Thanks for the explanation!  I was thinking maybe there was a particular reason why the stencil was in compute, so that's why I asked.

PR #491, opened 2021-05-12
remove compile step

twicki:
## Purpose

Removes the single compilation timestep that was done in run_on_daint


FlorianDeconinck (approved):

rheacangeo:
You could also change dynamics.py to have a new flag for whether or not to save the timing data rather than using the disable-halo-flags, but is not necessary. This should also solve the fv3core-performance-plots errors caused by the compile step producing a single timestep json file.

PR #492, opened 2021-05-12
overriding MapnTracer numpy threshold 

rheacangeo:
## Purpose
Since the addition of the while loop in remapping (which is super awesome and super speeds that code), MapN_Tracer_2d fails for a few tracers at a few points with the c48_6ranks_standard dataset. The current threshold is set to 3.5e-11, and passing validation appears to require 9e-9. Given the jump in 2 orders of magnitude, we may want to investigate this use-case. But it does not appear to propagate to require relaxing of parent tests, so that's good. 

## Code changes:

- update the threshold overrides file

### DIFF
--- a/tests/translate/overrides/standard.yaml
+++ b/tests/translate/overrides/standard.yaml
@@ -1,8 +1,3 @@
-C_SW:
-  - backend: numpy
-    platform: docker
-    max_error: 2e-10
-
 CS_Profile_2d:
   - backend: gtcuda
     max_error: 2.5e-9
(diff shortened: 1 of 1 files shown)


tests/translate/overrides/standard.yaml
```diff
@@ -1,8 +1,3 @@
-C_SW:
-  - backend: numpy
-    platform: docker
-    max_error: 2e-10
```
rheacangeo, line 4: removing this because it was only added as a demonstration of the format before we had overrides

tests/translate/overrides/standard.yaml
```diff
@@ -24,6 +19,10 @@ MapN_Tracer_2d:
     near_zero: 1e-17
     ignore_near_zero_errors:
       - qtracers
+  - backend: numpy
```
eddie-c-davis, line 22: Very precise!

eddie-c-davis (approved):
Looks good!

PR #494, opened 2021-05-13
Surface vertical velocity is always 2d

rheacangeo:
## Purpose

There was a conversion for wsd between 2d and 3d every timestep as parts of the code optimized to use it for the 2d nature it really is, and it was much more involved to unravel the 3d nature from this variable in remapping at the time. This PR unravels that, so we can always use wsd in 2d. The time converting is minimal, but as we investigate GPU profiles, we are seeing it as CPU work chunks that don't need to happen anymore. 

## Code changes:
- removed CPU assignment statements for wsd_3d and wsd in fv_dynamics
- wsd and the qs variable is becomes at lower levels is converted to a FloatIJ
- Since remapping uses wsd at the deepest level, remap_profile, and sometimes a different variable, gz, which is not 2d was passed to remap_profile, this touches many of the remapping files.
- In the fortran code 'gz' is passed to map1_ppm for iv modes where qs is not actually used, so has no impact. Instead define a temporary variable in the MapSingle class that can be used when an actual variable is not passed it. gz is no longer passed in, it was never used. 
- test code is updated to make wsd 2d, but there may be an option to do this more generally rather than duplicating the logic. I am open to putting more effort here to make that cleaner. 

## Checklist

### DIFF
--- a/fv3core/stencils/remapping_part1.py
+++ b/fv3core/stencils/remapping_part1.py
@@ -348,7 +348,7 @@ class VerticalRemapping1:
         self._copy_j_adjacent(self._pe2)
 
         self._pn2_pk_delp(self._dp2, delp, self._pe2, self._pn2, pk, akap)
-        
+
         self._map_single_pt(pt, peln, self._pn2, qmin=self._t_min)
 
         # TODO if self._nq > 5:
--- a/fv3core/stencils/remapping_part1.py
+++ b/fv3core/stencils/remapping_part1.py
@@ -303,7 +303,7 @@ class VerticalRemapping1:
         omga: FloatField,
         ak: FloatFieldK,
         bk: FloatFieldK,
-        gz: FloatFieldIJ,
+        gz: FloatField,
         cvm: FloatField,
         ptop: float,
     ):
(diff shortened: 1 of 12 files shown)


fv3core/stencils/remapping_part1.py
```diff
@@ -299,11 +299,11 @@ def __call__(
         hs: FloatFieldIJ,
         te: FloatField,
         ps: FloatFieldIJ,
-        wsd: FloatField,
+        wsd: FloatFieldIJ,
         omga: FloatField,
         ak: FloatFieldK,
         bk: FloatFieldK,
-        gz: FloatField,
+        gz: FloatFieldIJ,
```
gmao-ckung, line 306: I don't know if there is upcasting/downcasting of fields/storages, but I figure I mention just in case.  In `remapping.py`, line 51, `gz` is defined as a `FloatField`, and the call to `_moist_cv_pt_pressure` also has `gz` as a `FloatField`, but it's typed here as `FloatFieldIJ`.  Perhaps it'll still work, but I wanted to double check.

gmao-ckung (approved):
Looks good!  I only had one small question.

fv3core/stencils/remapping_part1.py
rheacangeo: good catch, this was an accident -- at first though gz was supposed to be 2d, but it's not

PR #496, opened 2021-05-13
fixing linting fail in last pr

rheacangeo:
## Purpose

Fixing linting on a new unit test. So that all subsequent PRs don't run into my mistake in parallel. Currently when you introduce a new file our linting test step doesn't catch errors in it.


rheacangeo:
Fixing linting on a new unit test. So that all subsequent PRs don't run into my mistake in parallel. Currently when you introduce a new file our linting test step doesn't catch errors in it.

PR #497, opened 2021-05-14
stencils_varied_bounds tests works with 54 ranks

rheacangeo:
Small fix so that the stencils-varied_bounds unit test added (that will be deleted when the feature is removed in a few months) passes 54 ranks.


elynnwu (approved):
Great fix to get around the last grid not having istart, thanks Rhea! This should go away when we don't have grid as global variable anymore.

PR #506, opened 2021-05-18
Fix `nvtx` marking in `nsys` jenkins run

FlorianDeconinck:
## Purpose

`--profile` option on run on `daint` consume the profile hook and remove the entry point we use to do nvtx marking

## Code changes:

- Remove `--profile` for `nsys` run
- Go from 2 steps to 3 to get a bit more data

## Infrastructure changes:

- `fv3core-profile` plan should now have nvtx markings on the `.qdrep`


twicki (approved):
good catch!

PR #507, opened 2021-05-18
Selective Validation for UpdateDzC

oelbert:
## Purpose

This PR changes the selective validation framework to allow multiple variables to be selectively validated with different origins and domains for validation, and uses that to simplify updatedzc

## Code changes:

- validation.py: `get_selective_class` now takes as an argument dictionary mapping a variable name to its serialized name and the function that generates the origin and domain to validate over, in order to allow each variable to have a separate origin/domain for selective validation, as well as internal refactors to handle this dictionary argument.
- test_selective_validation.py: Updated `test_wrapped_class` to use new `get_selective_class` structure
- updatedzc.py: Consolidated all stencils into one `updatedzc` stencil

## Checklist


fv3core/testing/validation.py
```diff
@@ -27,11 +26,38 @@ class SelectivelyValidated:
         """
 
         def __init__(self, *args, **kwargs):
+
+            selective_arg_names = []
+            origin_domain_funcs = []
```
mcgibbon, line 31: Instead of these lists, can you use a Mapping[str, Callable] in the argument above and do `for arg_name, func in name_to_function.items():`?

fv3core/testing/validation.py
```diff
@@ -9,8 +9,7 @@
 
 def get_selective_class(
     cls: type,
-    selective_arg_names,
-    origin_domain_func: Callable[..., Tuple[Tuple[int, ...], Tuple[int, ...]]],
+    name_to_function_dict,
```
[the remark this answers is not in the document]
mcgibbon: With a type hint you also don't need `_dict` in the name.
```suggestion
    name_to_function: Mapping[str, Callable[..., Tuple[Tuple[int, ...], Tuple[int, ...]]]],
```

mcgibbon (changes requested):

rheacangeo:
launch jenkins

rheacangeo:
launch jenkins

fv3core/testing/validation.py
```diff
@@ -114,14 +122,22 @@ def _set_nans(self, tracers: Mapping[str, Quantity]):
...
     origin = instance.grid.compute_origin()
     domain = instance.grid.domain_shape_compute(add=(0, 0, 1))
     return origin, domain
 
 
+def get_compute_domain_2d(
+    instance,
+) -> Tuple[Tuple[int, ...], Tuple[int, ...]]:
+    origin = (instance.grid.is_, instance.grid.js)
+    domain = (instance.grid.npx, instance.grid.npy)
```
mcgibbon, line 137: nit: if possible, use the compute/full origin/domain functions for this, since it describes the intent of these indices. I'm not actually sure if the domain argument here is accurate or not, because I can't remember if npx and npy are n_cells or n_cells + 1, which is kind of the issue.
```suggestion
    origin = instance.grid.compute_origin()[:2]
    domain = instance.grid.domain_shape_compute()[:2]
```

fv3core/testing/validation.py
```diff
@@ -114,14 +122,22 @@ def _set_nans(self, tracers: Mapping[str, Quantity]):
...
+    return origin, domain
```
mcgibbon, line 138: Or maybe a better code-reuse way to write this function.
```suggestion
    return get_compute_domain_k_interfaces(instance)[:2]
```

fv3core/testing/validation.py
```diff
@@ -114,14 +122,22 @@ def _set_nans(self, tracers: Mapping[str, Quantity]):
...
-def get_compute_domain(
+def get_compute_domain_k_interfaces(
     instance,
 ) -> Tuple[Tuple[int, ...], Tuple[int, ...]]:
     origin = instance.grid.compute_origin()
     domain = instance.grid.domain_shape_compute(add=(0, 0, 1))
     return origin, domain
 
 
+def get_compute_domain_2d(
```
mcgibbon, line 133: I'll keep the code-edit comments i wrote below for reference/feedback, but I see now that this function isn't actually used - can you delete it?

tests/test_selective_validation.py
```diff
@@ -41,7 +41,10 @@ def test_wrapped_class(all_arg_names, selective_arg_names, shape, origin, domain
             with no nans
     """
     origin_domain_func = MagicMock(return_value=(origin, domain))
-    Wrapped = get_selective_class(DummyClass, selective_arg_names, origin_domain_func)
+    name_to_function = {}
```
mcgibbon, line 44: Can you add tests of the new feature? This may be easiest as a new test, because you need two different origin_domain_func's and you need to look at `origin_domain_func.call_count` to make sure that each one is called once. You don't need to cover al the origin/domain/argument combinations covered by this existing test, one case should be enough.

tests/translate/translate_updatedzc.py
```diff
@@ -5,11 +5,11 @@
 class TranslateUpdateDzC(TranslateFortranData2Py):
     def __init__(self, grid):
         super().__init__(grid)
-        update_gz_on_c_grid = updatedzc.UpdateGeopotentialHeightOnCGrid(grid)
+        self.update_gz_on_c_grid = updatedzc.UpdateGeopotentialHeightOnCGrid(grid)
 
         def compute(**kwargs):
             kwargs["dt"] = kwargs.pop("dt2")
-            update_gz_on_c_grid(**kwargs)
+            self.update_gz_on_c_grid(**kwargs)
```
mcgibbon, line 12: This change is not necessary, can you revert it? `compute` both before and after is referencing a variable from the outer scope of the function, the variable being referenced has just been changed from `update_gz_on_c_grid` to `self` (since `self` is undefined in the local scope of this `def compute`).

mcgibbon (changes requested):
Needs tests. Great work though!  I'm surprised we just didn't need to operate on one of the halos.

fv3core/testing/validation.py
oelbert: I don't like this as much as the other suggestion. Code reuse is good, but I like having each function be independent conceptually.

mcgibbon: I liked this suggestion more when the other function was called `get_compute_domain` and this one was called `get_compute_domain_2d`.

tests/test_selective_validation.py
```diff
@@ -16,45 +16,100 @@ def __call__(self, *args, **kwargs):
...
+    "all_arg_names, selective_arg_names_shape1, selective_arg_names_shape2",
     [
-        pytest.param(["a"], [], id="no_selective_validation"),
-        pytest.param(["a"], ["a"], id="one_out_of_one_arg"),
-        pytest.param(["a", "b", "c"], ["b"], id="one_out_of_three_args"),
-        pytest.param(["a", "b", "c"], ["a", "b"], id="two_out_of_three_args"),
+        pytest.param(["a"], [], [], id="no_selective_validation"),
+        pytest.param(["a"], ["a"], [], id="one_out_of_one_arg"),
+        pytest.param(["a", "b", "c"], ["a"], [], id="one_shape1_arg"),
+        pytest.param(["a", "b", "c"], [], ["a"], id="one_shape2_arg"),
```
mcgibbon, line 50:
```suggestion
        pytest.param(["a", "b", "c"], ["a"], [], id="one_shape1_arg"),
```
These two tests are identical from the standpoint of the tested code, shape1 and shape2 are implementation details of the test and not features of the tested API. you should be able to delete one case.

tests/test_selective_validation.py
```diff
@@ -16,45 +16,100 @@ def __call__(self, *args, **kwargs):
...
+    origin2,
+    domain2,
+):
     """
     Tests behavior of the returned class from get_selective_class. Tests that it:
         - sets the correct number of gridcells to NaN
         - returns a validation subset that is the right shape, and has no nans
         - subsets only the arguments it was told to, and leaves others untouched
             with no nans
+        - uses the correct domain and origin for each variable if they differ
```
mcgibbon, line 76: This test has gotten a lot bigger, can you split it into some helper functions?

fv3core/testing/validation.py
rheacangeo: npz and npy also are global tile indices, so this wouldn't have worked for 54 ranks

rheacangeo:
the fill corners can't be run as a function inside a stencil, need to call the separate stencil, but the copy calls could be combined into a single stencil at least.

fv3core/stencils/updatedzc.py
```diff
@@ -126,24 +124,7 @@ def __call__(
...
-            origin=self.grid.full_origin(),
-            domain=self.grid.domain_shape_full(add=(0, 0, 1)),
-        )
-        basic_operations.copy_stencil(
-            gz,
-            self._gz_y,
-            origin=self.grid.full_origin(),
-            domain=self.grid.domain_shape_full(add=(0, 0, 1)),
-        )
+        self._double_copy_stencil(gz, self._gz_x, self._gz_y)
```
mcgibbon, line 127: Should add a TODO with the plans you currently have for this object, since that information is easy to lose over time. Feel free to use this or write your own comment.

```suggestion
        # TODO: delete gz_x and gz_y and just use gz when regions bug is fixed
        # currently doing so breaks validation
        self._double_copy_stencil(gz, self._gz_x, self._gz_y)
```

fv3core/testing/validation.py
```diff
@@ -52,15 +62,15 @@ def subset_output(self, varname: str, output: np.ndarray) -> np.ndarray:
...
             return output
 
         def _set_nans(self, kwargs):
             for name in set(kwargs.keys()).intersection(self._selective_argument_names):
                 array = kwargs[name]
-                validation_data = np.copy(array[self._validation_slice])
+                validation_data = np.copy(array[self._validation_slice[name]])
                 array[:] = np.nan
-                array[self._validation_slice] = validation_data
+                array[self._validation_slice[name]] = validation_data
```
mcgibbon, line 73: If you use the following for this block and the suggestion for subset_output, you can delete `self._selective_argument_names` altogether:
```
            for name, validation_slice in self._validation_slice.items():
                if name in kwargs:
                    array = kwargs[name]
                    validation_data = np.copy(array[validation_slice])
                    array[:] = np.nan
                    array[validation_slice] = validation_data
```

fv3core/testing/validation.py
```diff
@@ -52,15 +62,15 @@ def subset_output(self, varname: str, output: np.ndarray) -> np.ndarray:
             like to validate against reference data
             """
             if varname in self._selective_argument_names:
```
mcgibbon, line 64:
```suggestion
            if varname in self._validation_slice:
```

tests/test_selective_validation.py
mcgibbon: Please merge this suggestion.

fv3core/testing/validation.py
```diff
@@ -27,16 +28,25 @@ class SelectivelyValidated:
...
+                )
+
             self._all_argument_names = tuple(
                 inspect.getfullargspec(self.wrapped).args[1:]
             )
             assert "self" not in self._all_argument_names
-            self._selective_argument_names = selective_arg_names
+            self._selective_argument_names = tuple(
+                name_to_origin_domain_function.keys()
+            )
```
mcgibbon, line 49: I might as well write this as a code suggestion along with the other two (see suggestion in set_nans).
```suggestion
```

mcgibbon (approved):
Suggested to remove a private variable in favor of using a dictionary's keys, then two even more minor line changes. Shouldn't need another review.

Thanks for this refactor!

PR #511, opened 2021-05-24
fvsubgridz is 3d and OOPed

rheacangeo:
## Purpose

fv_subgridz is outside of the main 'fv_dynamics' loop and often gets left behind in refactors, and is the last piece preventing us from completing the transition from stencils decorated in the file and compiled JIT to one using objects to initialize/compile the whole model before running it. This PR catches it up, though a choice was made to unroll all of the tracers operations, with the idea we'll be adding a gt4py feature that can do this loop inside of the stencil. It was necessary to do so for the m_loop to be consolidated, and changed the other stencils to make them consistent. 
The main purpose of this PR:
  -- make a single stencil inside of the n-loop rather than looping over k to adjust both this layer and the one below it simultaneously. 
  -- make FVSubgridZ an object that matches the conventions of our other stencils. 
What this does not do:
 -- m_loop has a lot of repetition. We could make more effort to use more functions, but most variations tried resulted in failed validation. More could be done here, but in favor of getting this code up to date I suggest we delay that to a future optimization, when we revisit it to apply the loops over tracers. 

## Code changes:

- fv_subgridz is rewritten to be 3d and as a class
- removed remaining references to 'sd' and 'si'
- updated example runfiles to call the FvSubgridZ object

## Checklist

### DIFF
--- a/fv3core/stencils/fv_subgridz.py
+++ b/fv3core/stencils/fv_subgridz.py
@@ -132,7 +132,7 @@ def recompute_qcon(
             qcon = qcon_func(qcon, q0_liquid, q0_ice, q0_snow, q0_rain, q0_graupel)
 
 @gtscript.function
-def adjust_cvm( cpm, cvm, q0_vapor, q0_liquid, q0_rain, q0_ice, q0_snow, q0_graupel, gz, u0, v0, w0, t0, te):
+def adjust_cvm( cpm, cvm, q0_vapor, q0_liquid, q0_rain, q0_ice, q0_snow, q0_graupel, gz, u0, v0, w0, t0, te, hd):
     cpm, cvm = standard_cm(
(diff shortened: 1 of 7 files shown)


fv3core/stencils/fv_subgridz.py
```diff
@@ -54,659 +57,920 @@ def tvol(gz, u0, v0, w0):
...
+            state.qrain,
+            state.qice,
+            state.qsnow,
+            state.qgraupel,
+            state.qo3mr,
+            state.qsgs_tke,
+            state.qcld,
+        )
+
+        ratios = {0: 0.25, 1: 0.5, 2: 0.999}
```
rheacangeo, line 917: oops -- I'll delete this in the next push

oelbert (approved):
Looks good! I agree the repetition is a bit frustrating, but it does validate, and it's really exciting to get fvsubgridz caught up!

PR #513, opened 2021-05-24
Add GridIndexing, use explicit init args in xppm, yppm, fvtp2d

mcgibbon:
## Purpose

We're planning to refactor all initialization routines to take in minimal data, organized into neat objects to reduce arguments. As a first step, we plan to pass all data explicitly to pass minimal data, so that we can reason about what data is used where and package this configuration into objects.

Grid indexing is used throughout the repository in almost every class, so this PR defines a GridIndexing class to reduce the rather large number of arguments which would need to be passed for this down to one object.

The PR starts the refactoring process off with xppm, yppm, and fvtp2d.

## Code changes:

- Introduced GridIndexing class which provides grid indexing functionality from the existing Grid class.
- Refactored XPiecewiseParabolic, YPiecewiseParabolic, and FiniteVolumeTransport to use new GridIndexing class instead of the old Grid class, and remove all references to global state, taking in only the required information as init arguments instead

## Requirements changes:

- typing_extensions is added to requirements.txt and setup.py to provide the `Literal` type hint.

## Infrastructure changes:

- Added documentation to README on how to run pytest with `--pdb` in docker

## Checklist


fv3core/utils/grid.py
```diff
@@ -412,3 +602,46 @@ def axis_offsets(
...
         "local_je": gtscript.J[-1] + grid.je - origin[1] - domain[1] + 1,
     }
+
+
+@functools.lru_cache(maxsize=None)
+def _grid_indexing_axis_offsets(
+    grid: GridIndexing,
+    origin: Tuple[int, ...],
+    domain: Tuple[int, ...],
+) -> Mapping[str, gtscript._AxisOffset]:
```
mcgibbon, line 612: @jdahm This should be a public class in gtscript, because it's the type of a publicly available variable.

fv3core/stencils/d_sw.py
```diff
@@ -546,30 +546,53 @@ def __init__(self, namelist, column_namelist):
         self.delnflux_nosg_w = DelnFluxNoSG(self._column_namelist["nord_w"])
         self.delnflux_nosg_v = DelnFluxNoSG(self._column_namelist["nord_v"])
         self.fvtp2d_dp = FiniteVolumeTransport(
-            namelist,
-            namelist.hord_dp,
-            self._column_namelist["nord_v"],
-            self._column_namelist["damp_vt"],
+            grid=GridIndexing.from_legacy_grid(self.grid),
```
rheacangeo, line 549: lol, the name 'legacy' grid is funny.

fv3core/stencils/fvtp2d.py
```diff
@@ -80,16 +82,28 @@ class FiniteVolumeTransport:
...
+        area,
+        grid_type: int,
+        hord,
+        nord=None,
+        damp_c=None,
+    ):
+        # TODO: add a docstring
+        self.grid = grid
+        self._area = area
+        shape = self.grid.domain_full(add=(1, 1, 1))
```
rheacangeo, line 99: since we are making a new object anyways, how about we add an abstraction for the shape, this is the shape used by all 3d fields in the model, could be something like domain_max, or field_shape_3d or whatever you prefer.

fv3core/stencils/tracer_2d_1l.py
```diff
@@ -170,7 +170,14 @@ def __init__(self, comm: fv3gfs.util.CubedSphereCommunicator, namelist):
             domain=self.grid.domain_shape_compute(),
             externals=local_axis_offsets,
         )
-        self.finite_volume_transport = FiniteVolumeTransport(namelist, namelist.hord_tr)
+        self.finite_volume_transport = FiniteVolumeTransport(
+            grid=self.grid.grid_indexing,
```
rheacangeo, line 174: Why are you setting this differently here compared to the call in d_sw? e.g. grid=GridIndexing.from_legacy_grid(self.grid), vs. self.grid.grid_indexing. Either way is fine, but is less confusing if we use the same convention

fv3core/stencils/xppm.py
```diff
@@ -244,12 +244,20 @@ class XPiecewiseParabolic:
...
-        self._dxa = grid.dxa
-        origin = (grid.is_, jfirst, 0)
-        domain = (grid.nic + 1, jlast - jfirst + 1, grid.npz + 1)
+    def __init__(
+        self,
+        grid: GridIndexing,
+        dxa,
+        grid_type: int,
+        iord,
+        j_domain: Literal["compute", "full"],
```
rheacangeo, line 253: adding special strings can sometimes be more confusing than passing numbers. Also the fortran code is flexible to any jstart and jend. I'm on the fence, maybe we can keep it, but should consider implications for other modules that also pass through index numbers.

fv3core/stencils/xppm.py
```diff
@@ -264,6 +272,19 @@ def __init__(self, namelist, iord, jfirst, jlast):
...
+    ):
+        if j_domain == "compute":
+            add_origin = (0, 0, 0)
+            add_domain = (1, 1, 1)
+        elif j_domain == "full":
+            add_origin = (0, -grid.n_halo, 0)
+            add_domain = (1, 1 + 2 * grid.n_halo, 1)
+        origin = grid.origin_compute(add=add_origin)
+        domain = grid.domain_compute(add=add_domain)
+        return origin, domain
```
rheacangeo, line 286: This pattern is generic enough it would make sense to live in the grid indexing object. We do have other computations in the model that are compute in the i dimension and full in the j dimension and vice versa.

fv3core/utils/grid.py
```diff
@@ -366,14 +367,203 @@ def horizontal_starts_from_shape(self, shape):
...
+            west_edge: whether the current rank is on the west edge of a tile
+            east_edge: whether the current rank is on the east edge of a tile
+        """
+        self.origin = origin
+        self.domain = domain
+        self.n_halo = n_halo
+        # TODO: split edge booleans into their own class, they aren't needed
+        # for any of the indexing operations, only needed for axis_offsets
+        # and they make the API pretty big
+        # maybe give this responsibility to CubedSphereCommunicator
```
rheacangeo, line 411: I agree there is a TODO here, but I don't think they necessarily need their own class. Once axis_offsets is the only thing using them, we can just pass a CubedSphereCommunicator to axis_offsets. We should preserve the south_edge/west_edge nomenclature as that was a concrete bit of feedback we got from potential users that they would like to see.

fv3core/utils/grid.py
```diff
@@ -366,14 +367,203 @@ def horizontal_starts_from_shape(self, shape):
...
+            domain=grid.domain_shape_compute(),
+            n_halo=grid.halo,
+            south_edge=grid.south_edge,
+            north_edge=grid.north_edge,
+            west_edge=grid.west_edge,
+            east_edge=grid.east_edge,
+        )
+
+    @property
+    def isc(self):
```
rheacangeo, line 453: it is strange to me that the origin and domain initializing this object can be whatever, and the indices then are relative to that provided origin and domain. The compute domain should be the same indices always in this model. It might be more clear to pass in the 'c' value, e.g. 12, 48, whatever, and then set all of these attributes base on that. I am open to discussion on this. Is there a potential future use case where we'd want to pass in a different origin and domain to this object?

fv3core/utils/grid.py
```diff
@@ -366,14 +367,203 @@ def horizontal_starts_from_shape(self, shape):
...
+        return (
+            self.iec - self.isc + add[0],
+            self.jec - self.jsc + add[1],
+            self.ke - self.ks + add[2],
+        )
+
 
-@functools.lru_cache(maxsize=None)
 def axis_offsets(
+    grid: Union[Grid, GridIndexing],
```
rheacangeo, line 536: should axis_offsets live in grid.py? since they take in a grid, maybe it should belong to the GridIndexing class. The logic has already been duplicated to work with the new class anyways.

fv3core/utils/grid.py
```diff
@@ -412,3 +602,46 @@ def axis_offsets(
...
     }
+
+
+@functools.lru_cache(maxsize=None)
+def _grid_indexing_axis_offsets(
+    grid: GridIndexing,
+    origin: Tuple[int, ...],
+    domain: Tuple[int, ...],
+) -> Mapping[str, gtscript._AxisOffset]:
+    print(type(gtscript.I[0]))
```
rheacangeo, line 613: stray print statement

rheacangeo (commented):
Interesting proposal! I'd like to consider the impact on more complicated modules, but I think this could work well.

fv3core/stencils/fvtp2d.py
```diff
@@ -98,33 +112,43 @@ def __init__(self, namelist, hord, nord=None, damp_c=None):
...
         self.stencil_transport_flux = FrozenStencil(
             transport_flux_xy,
-            origin=self.grid.compute_origin(),
-            domain=self.grid.domain_shape_compute(add=(1, 1, 1)),
+            origin=self.grid.origin_compute(),
+            domain=self.grid.domain_compute(add=(1, 1, 1)),
         )
         if (self._nord is not None) and (self._damp_c is not None):
-            self.delnflux = DelnFlux(self._nord, self._damp_c)
+            self.delnflux: Optional[DelnFlux] = DelnFlux(self._nord, self._damp_c)
```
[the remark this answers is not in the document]
mcgibbon: The type hint is necessary here because mypy can't determine the type is Optional (can be None). Without a hint on this line, it gives an error.

fv3core/stencils/d_sw.py
mcgibbon: I could just name it `from_grid` if you like, but I'm not sure whether that's clear.

fv3core/stencils/fvtp2d.py
mcgibbon: Should it actually be the shape used by all 3d fields? I know it works, and that we (including me) did this because it's easy/the same everywhere and you don't have to think hard about it, but in theory all cell-centered variables should only need `domain_full` and interface variables only need an add in one or the other direction. If we did this the code would also make it clearer what's on cell centers or interfaces. I also suspect we're currently doing some computations on the last z-level that aren't needed but are never checked by the validation framework (which slices to the fortran-equivalent part of the array).

The GridIndexing object is currently written as "this provides indexing for cell-centered variables, use add if you need to use it for interface variables" which I find fairly easy to think about. I'm a little worried about adding an attribute on top of that which should only be needed for cell corner variables.

fv3core/stencils/tracer_2d_1l.py
mcgibbon: Only because I hadn't written the `grid_indexing` method when I wrote the change for d_sw. I'll refactor it there.

fv3core/stencils/xppm.py
mcgibbon: I'll refactor this to an `origin` and `domain` argument, which is general and also gets around the indexing issue I ran into.

fv3core/utils/grid.py
mcgibbon: I think we're on the same page. I'll update the todo to read "into a different class" instead of "into their own class".

>Once axis_offsets is the only thing using them, we can just pass a CubedSphereCommunicator to axis_offsets.

axis_offsets will need a GridIndexing as well as a CubedSphereCommunicator, because the CSC does not have information about the size of the domain (it gets this at call time based on the size of input arguments).

>We should preserve the south_edge/west_edge nomenclature as that was a concrete bit of feedback we got from potential users that they would like to see.

Yes 100%. "giving this responsibility to CubedSphereCommunicator" will mean adding `is_south_edge` etc. methods to CubedSphereCommunicator. We should also refactor the "on_tile_bottom" etc. methods in TilePartitioner.

fv3core/utils/grid.py
mcgibbon: We currently have a hard-codable compute origin and compute domain, but it's very possible we'd want to change that in the future, e.g. to use 4 halos for certain variables. The Sizer class currently does the conversion from cXX to an origin and domain, so we already have that initialization responsibility covered.

fv3core/utils/grid.py
mcgibbon: I like it. I'll move the new version to be a method on that class.

fv3core/stencils/d_sw.py
rheacangeo: You can leave it, I just laughed when I read it :)

fv3core/stencils/fvtp2d.py
rheacangeo: I think a future refactor could use different shapes, but is separate from the purposes of this PR. Making it clear we are using the same shape everywhere currently with a single named variable would provide clarity. It wasn't just simpler for us porting code to use the same shape, at the time it was necessary in gt4py. That is no longer true, but we don't know if there are bugs that only pop up when you use different shapes, especially when you get regions involved. Using the same shape everywhere is also easier to understand for the user, in my opinion. I doubt there is much performance implication with doing extra computation on some buffer points. We are indeed doing some calculation on the buffer z point where we don't need to, but it does not harm as those points are never read. 
Currently this variable would be used in every single class, every 3d variable is this shape currently and is currently saying shape = grid.domain_shape_full(add=(1, 1, 1)). I think refactoring to remove global state should be a separate and distinct set of PRs from one using different sized array.s

fv3core/utils/grid.py
rheacangeo: I don't know this use case of wanting to use 4 halos for certain variables -- what modules/experiments are you thinking of that would use this? The GridIndexing class then could use the Sizer class. This new class implies more flexibility than is possible to work(the old class does too, but while we are rewriting it, good to consider).

fv3core/stencils/xppm.py
mcgibbon: Done.

fv3core/stencils/xppm.py
mcgibbon: Noting here that we discussed adding an interface like:
```python
origin, domain = grid_indexing.get_origin_domain([X_INTERFACE_DIM, Y_DIM, Z_DIM], halos=[0, grid.n_halo])
```
which I'll do in a follow-on PR.

fv3core/utils/grid.py
mcgibbon: I removed the TODO for now since we added `axis_offsets` to this class. This might make it big enough that we want to split it up later, but I'm not confident enough in that to keep a TODO.

fv3core/utils/grid.py
mcgibbon: Removed.

fv3core/utils/grid.py
jdahm: Yes, it will be called `gtscript.AxisIndex` soon-ish.

fv3core/utils/grid.py
mcgibbon: > This new class implies more flexibility than is possible to work

Is that true? If we use the GridIndexing everywhere to handle horizontal dimensions, it should "just work" to provide arrays with 4 halo points to fv3core (alongside the appropriate GridIndexing instance). It's also hard to be certain we don't need to support 4 halo points, because fv3core isn't the top level of the model. If we hard-code 3 halo points, an ML team can't make physics schemes or integrate analysis code that needs more halo points.

Probably more importantly, it's a fairly complicated responsibility to go from a configuration to the shape of your arrays on a given rank. You need not just the cXX value, but also the layout, the rank, and how data is distributed on a tile across ranks. All of this is encapsulated right now [in Sizer](https://github.com/VulcanClimateModeling/fv3gfs-util/blob/master/fv3gfs/util/initialization/sizer.py#L32) (which in turn uses TilePartitioner for most of the work).

fv3core/utils/grid.py
```diff
@@ -366,14 +367,202 @@ def horizontal_starts_from_shape(self, shape):
...
+        return self.origin[1] + self.domain[1]
+
+    @property
+    def ks(self):
+        """start of the data domain along the z-axis"""
+        return self.origin[2]
+
+    @property
+    def ke(self):
+        """end of the data domain along the z-axis"""
```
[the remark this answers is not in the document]
mcgibbon: It may be OK to "duplicate" ks/ke in that way, in the same way that there were modules that took in "istart" or "iend". In both cases they have the same meaning (the start and end indices in the vertical), just applied to the appropriate context for the object (GridIndexing is about the full grid, RayleighFriction is about where you're applying Rayleigh friction).

> This may result in someone initializing a new grid indexer when they are using different values there, which we probably don't want

Yes, definitely. We talked about ways to handle vertical indexing separately and use GridIndexing for just the horizontal. Hopefully this doesn't come up often. If we find there are a lot of cases that need that then we can think about how to integrate those special cases into the GridIndexing object.

oelbert (approved):
I don't have any issues or concerns to add; good work!

fv3core/utils/grid.py
rheacangeo: What I mean is that the compute origin and domain should be fixed for all of the current model, so there is flexibility that is not currently needed. If there is indeed a known use case, sure, we can provide that.  Do some of the physics schemes or ML code actually need more halo points? I haven't heard of anyone wanting this use case in this model, but of course doesn't mean it won't be a possible request this would be ready for. The local index values set and used here should not depend on the rank and layout (though the 'edge' variables currently in the class will for now). The job of this class is relatively simple, I just hope it doesn't end up getting used in weird ways (like making a different grid indexing object of staggered variables), such that the meaning of the compute start and end indices changes dramatically. This is probably fine for now.

mcgibbon: > The local index values set and used here should not depend on the rank and layout (though the 'edge' variables currently in the class will for now).

We're thinking pretty seriously about reducing the number of points given to ranks at tile edges, to even out the work since they have extra edge handling logic to perform. This would make these local index values depend on rank.

The local index values already depend on layout. End indices are half as big for a (2, 2) layout as for a (1, 1) layout.

> The job of this class is relatively simple, I just hope it doesn't end up getting used in weird ways (like making a different grid indexing object of staggered variables)

Yes, let's try to avoid that. The new `origin_domain([X_DIM, Y_INTERFACE_DIM])` should hopefully placate anyone trying to handle interface variables.

fv3core/stencils/fvtp2d.py
mcgibbon: I'll add a `grid_indexing.max_shape` property, with a docstring explaining why it should be rarely used, to encourage us to do these refactors later.

rheacangeo (approved):

PR #522, opened 2021-06-03
Separate non-translate tests into their own make target

mcgibbon:
## Purpose

Currently, all non-translate tests run every time the translate tests are run, even when only a specific translate class is targeted by command-line options. The non-translate tests also require serialbox and can only be run in an environment with test data, after spending time loading that data to process savepoints.

This PR separates "savepoint" tests into their own make target and test directory, and python-only tests into a separate make target and test directory, allowing the tests to be run independently.

To keep the makefile simple when doubling the number of test types, this PR also unifies "dev" and "venv" make targets into a single set of make targets. "dev" is accomplished by setting `DEV=y` from the command line, while "venv" is equivalent to `CONTAINER_CMD=""`.

## Infrastructure changes:

- Previous make targets "tests" and "tests_mpi" are now "savepoint_tests" and "savepoint_tests_mpi"
- Make targets "tests" and "test" now run python-only tests, under `tests/main`
- "dev" targets are now run using `DEV=y` on the corresponding test make target
- "venv" targets are now run using `CONTAINER_CMD=""` on the corresponding test make target
- jenkins output file mounts have been moved to the bash script which specifies the pytest xml output argument that causes files to write to the mounted directory
- BASH_PREFIX make variable is removed, instead using CONTAINER_CMD in the one place it was previously used
- tests_base and tests_base_parallel are now unified into one make target, differing by the `PYTEST_CMD` value used

## Checklist


tests/conftest.py
```diff
@@ -1,42 +1,6 @@
-import collections
```
mcgibbon, line 1: This file was moved without edits.

tests/savepoint/conftest.py
```diff
@@ -0,0 +1,519 @@
+import collections
```
mcgibbon, line 1: This file was moved without edits.

Makefile
```diff
@@ -126,17 +131,20 @@ cleanup_remote:
...
+savepoint_tests:
 	$(MAKE) get_test_data
-	$(MAKE) run_tests_parallel
+	VOLUMES='-v $(TEST_DATA_HOST):$(TEST_DATA_RUN_LOC)' \
+	PYTEST_CMD="$(PYTEST_SEQUENTIAL)" $(MAKE) test_base
 
-test_mpi: tests_mpi
+savepoint_tests_mpi:
+	$(MAKE) get_test_data
+	VOLUMES='-v $(TEST_DATA_HOST):$(TEST_DATA_RUN_LOC)' \
```
rheacangeo, line 146: shouldn't these VOLUMES be a += rather than overwriting? What about the VOLUME you set in the jenkins action or the DEV above?

docker/Dockerfile
```diff
@@ -54,7 +54,6 @@ RUN chmod +x /entrypoint.sh && \
 ARG ENV_CUDA_PATH=""
 ENV CUDA_PATH=${ENV_CUDA_PATH}
 ENV IN_DOCKER=True
-ENTRYPOINT [ "/entrypoint.sh" ]
```
rheacangeo, line 57: this will help development, hopefully the reasons it was added in the first place don't apply

rheacangeo (commented):
Looks good, I am glad to have the translate tests separated from the straight python tests. Thank you for simplifying the Makefile instead of making it larger. Some of the reasons for the separate endpoints no longer apply and this looks like it works. With most infrastructure refactors we encounter downstream impacts we didn't think about during the review or testing phase, so I'll ready myself for those. I don't anticipate anything major though, this seems like a good improvement. One small question and then I'll approve.

Makefile
```diff
@@ -148,38 +156,11 @@ dev:
...
-dev_tests:
-	VOLUMES=$(DEV_MOUNTS) $(MAKE) test_base
-
-dev_tests_mpi:
-	VOLUMES=$(DEV_MOUNTS) $(MAKE) test_base_parallel
-
-dev_test_mpi: dev_tests_mpi
-
-dev_tests_mpi_host:
-	MOUNTS=$(DEV_MOUNTS) $(MAKE) run_tests_parallel_host
```
rheacangeo, line 160: This leftover endpoint from a a previous era of things was where MOUNTS was still set. You can remove the MOUNTS variable.

Makefile
mcgibbon: Yes, good catch! It looks like this was a pre-existing bug, which will now affect us because I'm using VOLUMES= from the command-line. Will fix.

rheacangeo: I don't think it is a bug in master as VOLUMES isn't used outside of the Makefile and is set to different values for the different endpoints, but with them combined is great o have them be additive

mcgibbon: Yes the "bug" in master is that it's settable outside the Makefile (uses `?=` and not `=`) but if you set it then in theory it will be ignored/overridden. In practice we've never set it outside the Makefile so it's never affected us.

rheacangeo: The use case would have been running VOLUMEs=<> make test_base separately with some custom volumes, while the other endpoints hardcoded it and indeed couldn't add to it or override it. It was intended behavior.

rheacangeo: (that is no longer needed)

rheacangeo (approved):

PR #523, opened 2021-06-04
manually run jenkins tests on daint without workarounds

rheacangeo:
## Purpose

To enable easily running the jenkins tests as a dev on Piz Daint with just
`slave=daint .jenkins/jenkins.sh run_regression_tests numpy c12_6ranks_standard`

## Code changes:

- .jenkins/jenkins.sh
    - changed default VIRTUALENV if not set to a local directory path
   - installs the VIRTUALENV if it doesn't exist yet and running in virtualenv mode
   -  changed default JENKINS_TAG
   - using the constant DATA_VERSION rather than FORTRAN_VERSION to match the standalone convention
   - default TEST_DATA_DIR is the /project location so you shouldn't have to wait to download data when you decide to run manually
   - the module load of gcloud is updated to the one that works now, in case you want to download a different dataset. 
- removed setup jenkins action, instead the plan calls the install script directly. 
- Makefile has separate options for updating submodules for the venv vs docker
- pip installs of fv3core both with jenkins and in the standalone are editable so it's easier to discover the impact of changes to code. 
- standalone cleans up the cache files and virtualenv on job failures only if running through jenkins (e.g. the BUILD_TAG environment variable is set). 

## Checklist


.jenkins/jenkins.sh
```diff
@@ -122,11 +122,9 @@ if grep -q "fv_dynamics" <<< "${script}"; then
 fi
 
 module load daint-gpu
-module add "${installdir}/modulefiles/"
-module load gcloud
-
+module load ${installdir}/modulefiles/gcloud/303.0.0
```
twicki, line 125: is there a reason to switch this? I think the more general approach of adding the module directory as it allows us to load all the modules we custom made.

In the case where we only need the one it is also fine this way. Just curious

Makefile
```diff
@@ -47,9 +47,14 @@ clean:
 
 update_submodules:
 	if [ ! -f $(FV3UTIL_DIR)/requirements.txt  ]; then \
-		git submodule update --init external/fv3gfs-util external/daint_venv; \
+		git submodule update --init external/fv3gfs-util ; \
 	fi
 
+update_submodules_venv: update_submodules
```
twicki, line 53: Can we improve the naming here that `update_submodules` still does everything and have one that just does venv and one that just does utils if we need that functionality?

twicki (changes requested):
two minor questions only. Looks really good

.jenkins/jenkins.sh
rheacangeo: I thought it didn't work anymore to module load the directory. I can check again

Makefile
rheacangeo: The utils one is always needed, but the venv one is only needed when using a virtualenv. It has caused problems for users trying to use fv3core out of the box with docker, expanding a submodule that's not even used.  Almost everything in the Makefile is geared toward the container version of things. I could rename update_submodules to something like update_submodules_utils, but if there is another base submodule added, could instead call it update_submodules_base?

.jenkins/jenkins.sh
rheacangeo: Yes that still doesn't work -- please let me know the command if it works for you

twicki (approved):
lgtm

PR #528, opened 2021-06-07
update boost location

rheacangeo:
## Purpose

Boost has moved! docker dependencies file should get it from the right spot. Luckily we have been shielded from this problem by not rebuilding all the dependencies from scratch. 

## Code changes:

- dependencies.Dockerfile points to updated boost location


twicki (approved):
lgtm

PR #537, opened 2021-06-14
Changes needed for using gtc backends

gronerl:
Changes to stencils are where current code relies on undefined behavior:
(1)
```
- with computation(PARALLEL), interval(...):
+ with computation(PARALLEL), interval(0, -1):
```
This was changed in places where the respective interval reads at `k+1`, which would result in an OOB read. The front end seems to not check this currently, but the dace subsets run out of bounds. Not sure why this currently works, at least numpy should complain. In any case, the behavior is not defined with `interval(...)` in this case.
(2)
an additional `with computation` was introduced since the code was not compliant with the parallel model implemented in GTC. 

Further, the `get_state_from_input` returned a dict with a single key to a SimpleNamespace. On the dict,  `__dict__` was accessed, which is not an attribute of dicts. instead the SimpleNamespace is returned directly, which seems to be the intended behavior.


ofuhrer:
Can one of the admins verify this patch?

ofuhrer:
launch jenkins

gronerl:
Note that this is a PR into the `for-dace` branch.  Sorry if this just triggered everyone's notifications, I just wanted to stick to a single branch for everyone working on the acoustic/dace project.

fv3core/stencils/nh_p_grad.py
```diff
@@ -27,7 +27,7 @@ def calc_u(
     rdx: FloatFieldIJ,
     dt: float,
 ):
-    with computation(PARALLEL), interval(...):
+    with computation(PARALLEL), interval(0, -1):
```
rheacangeo, line 30: If you are going to make this change, you should also change the k range that is called, otherwise this breaks validation

rheacangeo:
Thanks! We appreciate the PR! It fixes some things we were confused on. With regard to the interval changes, the data arrays all have an extra datapoint in the k dimension, and gt4py is able to read outside of the origin and domain specified, just not write to it.

fv3core/stencils/nh_p_grad.py
rheacangeo: replace
`self.domain_full_k = grid.domain_shape_compute(add=(1, 1, 0)) 
self.u_domain = grid.domain_shape_compute(add=(0, 1, 0))      
  self.v_domain = grid.domain_shape_compute(add=(1, 0, 0))`
with
`self.domain_full_k = grid.domain_shape_compute(add=(1, 1, 1)) 
self.u_domain = grid.domain_shape_compute(add=(0, 1, 1))      
  self.v_domain = grid.domain_shape_compute(add=(1, 0, 1))`

fv3core/stencils/dyn_core.py
```diff
@@ -116,7 +116,7 @@ def p_grad_c_stencil(
     """
     from __externals__ import hydrostatic
 
-    with computation(PARALLEL), interval(...):
+    with computation(PARALLEL), interval(0, -1):
```
rheacangeo, line 119: To make this change please also change in this file:
-        pgradc_domain = self.grid.domain_shape_compute(add=(1, 1, 0))
+        pgradc_domain = self.grid.domain_shape_compute(add=(1, 1, 1))

gronerl:
Huh, this is interesting, there are no checks or analysis in place to make sure that the shapes are correct in K, but if you get it right, it works for most backends, thanks to still putting origins and offsets in K everywhere. I will consider updating the dace backend to permit it as well in that case :thinking:

gronerl:
launch jenkins

rheacangeo:
All the fields do have the same shape, but in fv3core we always specify the origin and domain that indicates the domain being changed typically. So setting interval(...), but passing a domain that is sized one smaller than the data shape allows us to still access the datapoint outside of the specified range for most backends. If we weren't setting origin and domain and relying on gt4py to determine what those should be, it'd be a different story for sure.

PR #539, opened 2021-06-15
add --print_domains option to savepoint tests

mcgibbon:
## Purpose

It can be difficult to track regressions that are due to inadvertently changing the origin and domain of stencil calls. This PR makes that process much easier by adding a `--print_domains` argument to the savepoint tests which causes the function name, origin, and domain to be printed whenever a FrozenStencil is created, allowing these values to be compared between the main branch and a PR branch.

## Code changes:

- `fv3core.decorators` is now imported automatically when `fv3core` is imported

## Infrastructure changes:

- Added --print-domains argument for savepoint tests which causes the function name, origin, and domain to be printed whenever a FrozenStencil is created

## Checklist

### DIFF
--- a/tests/savepoint/conftest.py
+++ b/tests/savepoint/conftest.py
@@ -492,5 +492,16 @@ def failure_stride(pytestconfig):
 
 
 @pytest.fixture()
+def print_domains(pytestconfig):
+    value = bool(pytestconfig.getoption("print_domains"))
+    original_value = fv3core.decorators.FrozenStencil.DEBUG
+    try:
+        fv3core.decorators.FrozenStencil.DEBUG = value
+        yield value
+    finally:
+        fv3core.decorators.FrozenStencil.DEBUG = original_value
+
+
+@pytest.fixture()
 def python_regression(pytestconfig):
     return pytestconfig.getoption("python_regression")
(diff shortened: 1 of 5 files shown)


tests/savepoint/conftest.py
```diff
@@ -491,6 +491,17 @@ def failure_stride(pytestconfig):
     return int(pytestconfig.getoption("failure_stride"))
 
 
+@pytest.fixture()
+def print_domains(pytestconfig):
+    value = bool(pytestconfig.getoption("print_domains"))
+    original_value = fv3core.decorators.FrozenStencil.DEBUG
+    try:
+        fv3core.decorators.FrozenStencil.DEBUG = value
```
FlorianDeconinck, line 499: If you are doing a FrozenStencil wide debug you might want to do and or on assignement to not override another future usage

FlorianDeconinck (approved):

tests/savepoint/conftest.py
mcgibbon: I think for now I’ll leave it as is, I had intended for it to override whatever value is there for the duration of the tests. Good to keep in mind if there’s ever a reason to enable this by default or through a different mechanism.

mcgibbon: I refactored the code to put this concern in the test code, which should avoid the issue you mention.

PR #546, opened 2021-06-24
DGrid2AGrid2CGridVectors has minimal init, no global access

mcgibbon:
## Purpose

This PR refactors DGrid2AGrid2CGridVectors to use no global state, expanding its init routine to take in all required variables explicitly.

## Checklist

### DIFF
--- a/fv3core/stencils/d2a2c_vect.py
+++ b/fv3core/stencils/d2a2c_vect.py
@@ -1,5 +1,3 @@
-from typing import Tuple
-
 import gt4py.gtscript as gtscript
 from gt4py.gtscript import PARALLEL, computation, horizontal, interval, region
 
@@ -396,10 +394,6 @@ class DGrid2AGrid2CGridVectors:
         sin_sg4,
         nested: bool,
         grid_type: int,
-        # TODO: remove npx and layout, and in doing so
-        # remove MPI-grid dependence of outputs
-        npx: int,
-        layout: Tuple[int, int],
         dord4: bool,
     ):
         self._cosa_s = cosa_s
@@ -482,7 +476,7 @@ class DGrid2AGrid2CGridVectors:
         origin = grid_indexing.origin_full()
         domain = grid_indexing.domain_full()
         ax_offsets = axis_offsets(grid_indexing, origin, domain)
-        if npx <= 13 and layout[0] > 1:
+        if npt == 0:
             d2a2c_avg_offset = -1
         else:
             d2a2c_avg_offset = 3
--- a/fv3core/stencils/d2a2c_vect.py
+++ b/fv3core/stencils/d2a2c_vect.py
@@ -1,12 +1,13 @@
+from typing import Tuple
+
 import gt4py.gtscript as gtscript
 from gt4py.gtscript import PARALLEL, computation, horizontal, interval, region
 
-import fv3core._config as spec
 import fv3core.utils.gt4py_utils as utils
 from fv3core.decorators import FrozenStencil
 from fv3core.stencils.a2b_ord4 import a1, a2, lagrange_x_func, lagrange_y_func
 from fv3core.utils import corners
-from fv3core.utils.grid import axis_offsets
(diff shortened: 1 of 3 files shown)


fv3core/stencils/d2a2c_vect.py
```diff
@@ -382,72 +379,110 @@ class DGrid2AGrid2CGridVectors:
...
         )
 
-        js1 = npt + OFFSET if self.grid.south_edge else self.grid.js - 1
-        je1 = ny - npt if self.grid.north_edge else self.grid.je + 1
-        is1 = npt + OFFSET if self.grid.west_edge else self.grid.isd
-        ie1 = nx - npt if self.grid.east_edge else self.grid.ied
+        js1 = npt + OFFSET if grid_indexing.south_edge else grid_indexing.jsc - 1
+        je1 = ny - npt if grid_indexing.north_edge else grid_indexing.jec + 1
+        is1 = npt + OFFSET if grid_indexing.west_edge else grid_indexing.isd
+        ie1 = nx - npt if grid_indexing.east_edge else grid_indexing.ied
```
rheacangeo, line 441: so fun. luckily all of this is removed in the single stencil version of this code

fv3core/stencils/d2a2c_vect.py
```diff
@@ -382,72 +379,110 @@ class DGrid2AGrid2CGridVectors:
...
+        sin_sg1,
+        sin_sg2,
+        sin_sg3,
+        sin_sg4,
+        nested: bool,
+        grid_type: int,
+        # TODO: remove npx and layout, and in doing so
+        # remove MPI-grid dependence of outputs
+        npx: int,
+        layout: Tuple[int, int],
```
rheacangeo, line 402: In the long run, it's true, we don't want to have a special case of this stencil for c12_54ranks. I think it is not unreasonable for modules to have awareness of the model sizing/rank, but in this case since it is only used for a workaround for small domains, I am ok with us eventually removing it, either by testing with c48 (or c20 or whatever) 54 ranks and raising an error early on in the model, or adapting gt4py or the validation to work in the case of c12_54ranks. I appreciate it not being changed in the scope of this PR>

rheacangeo (approved):
Looks good to me. A lot of this code goes away with regions/single stencil, and this change looks easy to adapt from. I still think it would be better for grid variables to be identified as such in their naming to make it extra easy on the user to immediately know where a variable is coming from, but that can be discussed more in a later refactor.

fv3core/stencils/d2a2c_vect.py
rheacangeo: Actually on second thought you could remove these now

fv3core/stencils/d2a2c_vect.py
```diff
@@ -382,72 +379,110 @@ class DGrid2AGrid2CGridVectors:
...
         )
 
-        origin = self.grid.full_origin()
-        domain = self.grid.domain_shape_full()
-        ax_offsets = axis_offsets(self.grid, origin, domain)
-        if namelist.npx <= 13 and namelist.layout[0] > 1:
+        origin = grid_indexing.origin_full()
+        domain = grid_indexing.domain_full()
+        ax_offsets = axis_offsets(grid_indexing, origin, domain)
+        if npx <= 13 and layout[0] > 1:
```
rheacangeo, line 485: This was written to represent the use case we wanted to avoid. But actually for now, you could just do exactly what the fortran is doing with GT4PY_DEV (and include a TODO to revisit still). 
npt = 0 if npt > (je - js) .or. npt > (ie - is)
 That is already set in this code above. So the check here could just be
`if npt == 0` set d2a2c_avg_offset = -1 otherwise   d2a2c_avg_offset = 3
This may not be the right thing for nested, but a lot of other code also would need to be added/modified to support nested. Then you can ignore the npx and layout check. I think this ended up this way due to multiple refactors and rollbacks by multiple people.

mcgibbon:
launch jenkins

PR #551, opened 2021-07-03
Allow ignore_near_zero_errors to have variable specific value

elynnwu:
## Purpose

This PR introduces the option to specify `ignore_near_zero_errors` as a dictionary with variable specific near_zero value. As a result of this PR, `c12_6ranks_baroclinic` passes tests using `baroclinic.yaml`.

## Code changes:

- When reading `ignore_near_zero_errors` with `--threshold_overrides_file` option, if it's a dictionary, the key is the variable and the value is its near_zero value
- `ignore_near_zero_errors` functionality is unchanged if the previous format is passed (a list of values with near_zero specified separately)
- If `near_zero` is specified and `ignore_near_zero_errors` is a dictionary, the value in `ignore_near_zero_errors` is used.

## Checklist

### DIFF
--- a/tests/savepoint/test_translate.py
+++ b/tests/savepoint/test_translate.py
@@ -190,7 +190,9 @@ def process_override(threshold_overrides, testobj, test_name, backend):
                                 ] = float(match["all_other_near_zero"])
 
                 else:
-                    raise TypeError("ignore_near_zero_errors is either a list or a dict")
+                    raise TypeError(
+                        "ignore_near_zero_errors is either a list or a dict"
+                    )
         elif len(matches) > 1:
             raise Exception(
                 "misconfigured threshold overrides file, more than 1 specification for "
--- a/tests/savepoint/test_translate.py
+++ b/tests/savepoint/test_translate.py
@@ -34,13 +34,24 @@ def success_array(computed_data, ref_data, eps, ignore_near_zero_errors, near_ze
         np.logical_and(np.isnan(computed_data), np.isnan(ref_data)),
         compare_arr(computed_data, ref_data) < eps,
     )
-    if ignore_near_zero_errors:
+    if isinstance(ignore_near_zero_errors, dict):
+        if ignore_near_zero_errors.keys():
+            near_zero = ignore_near_zero_errors["near_zero"]
+            success = np.logical_or(
+                success,
+                np.logical_and(
+                    np.abs(computed_data) < near_zero,
+                    np.abs(ref_data) < near_zero,
+                ),
+            )
+    elif ignore_near_zero_errors:
         success = np.logical_or(
             success,
             np.logical_and(
                 np.abs(computed_data) < near_zero, np.abs(ref_data) < near_zero
             ),
         )
+
     return success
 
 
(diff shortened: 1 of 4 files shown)


tests/savepoint/test_translate.py
```diff
@@ -34,13 +34,24 @@ def success_array(computed_data, ref_data, eps, ignore_near_zero_errors, near_ze
...
+        if ignore_near_zero_errors.keys():
+            near_zero = ignore_near_zero_errors["near_zero"]
+            success = np.logical_or(
+                success,
+                np.logical_and(
+                    np.abs(computed_data) < near_zero,
+                    np.abs(ref_data) < near_zero,
+                ),
+            )
+    elif ignore_near_zero_errors:
```
mcgibbon, line 47: I don't see any logic here involving comparing a variable name, which seems like it would be necessary?

Is our test code complicated enough that we need some unit tests for these helper functions?

mcgibbon (commented):
Great work! Thanks for the doc update.

Is it meant to be that you can specify ignore_near_zero_errors as a scalar for all variables or a value per-variable? If both aren't needed, I'd suggest removing support for scalar-only. If scalar-only is needed, I would suggest adding a way to specify a "for all other variables" option alongside the specific-variable options. It feels strange to not be able to specify a global value if you're specifying a specific value.

Not strictly needed, but if you can could you add type hints to some of the routines you modified? `ignore_near_zero_errors` in particular sounds like it's a float, having a type hint would make me a little less confused.

tests/savepoint/test_translate.py
elynnwu: So the way it checks the variable is by first setting all fields to have ignore_near_zero_errors to False, then with the override, it reads the yaml variable and change it to True. I think we're probably okay here?

elynnwu: Oh but you're right that we probably want to make sure that the specified variable is within the list of variables to be compared.

tests/savepoint/test_translate.py
```diff
@@ -148,9 +166,31 @@ def process_override(threshold_overrides, testobj, test_name, backend):
...
+                        for key in testobj.out_vars.keys():
+                            if key not in testobj.ignore_near_zero_errors:
+                                testobj.ignore_near_zero_errors[key] = True
+                                testobj.ignore_near_zero_errors[key] = {}
+                                testobj.ignore_near_zero_errors[key][
+                                    "near_zero"
+                                ] = float(match["all_other_near_zero"])
+
+                else:
+                    raise TypeError("ignore_near_zero_erros is either a list or a dict")
```
mcgibbon, line 193:
```suggestion
                    raise TypeError("ignore_near_zero_errors is either a list or a dict")
```

mcgibbon (approved):
Great job on this PR! Looks good to me, just one error string typo.

The testing framework itself has gotten fairly complicated, and I'm a little worried about potential bugs that could go untested/unnoticed (if they give false negatives). We talked about a follow-on task being to write unit tests for the "error checking and specification of error constraints" part of the test framework, which may involve a partial refactor. This could involve making a new error checking class (even just a function with configuration) to handle loading yaml specifications and checking for errors (maybe returning the "success" array).

elynnwu:
launch jenkins

tests/savepoint/test_translate.py
rheacangeo: the variable name comparison is already in the code, it was just set to Tru/False rather than a value

PR #552, opened 2021-07-07
Fix "easy" race conditions

jdahm:
## Purpose

Tackle the "easy" race condition fixes in `circulation_cgrid` and `divergence_corner`.


fv3core/stencils/c_sw.py
```diff
@@ -251,10 +251,10 @@ def circulation_cgrid(
         vort_c = fx[0, -1, 0] - fx - fy[-1, 0, 0] + fy
 
         with horizontal(region[i_start, j_start], region[i_start, j_end + 1]):
-            vort_c += fy[-1, 0, 0]
+            vort_c += dyc[-1, 0] * vc[-1, 0, 0]
 
         with horizontal(region[i_end + 1, j_start], region[i_end + 1, j_end + 1]):
-            vort_c -= fy[0, 0, 0]
+            vort_c -= dyc * vc
```
rheacangeo, line 258: This works. You could consider making the regions independent of the main domain computation, though is not necessary. e.g.
`vort_c = fx[0, -1, 0] - fx + fy` for the first region and `vort_c = fx[0, -1, 0] - fx - fy[-1, 0, 0]` for the second. This may cause validation to fail, but in this case it is justifiable to loosen the empirical error thresholds since we can see exactly the math change that would cause the answer to be slightly different. If you prefer not to mess with that, I am fine with us leaving this as-is.

fv3core/stencils/c_sw.py
```diff
@@ -218,12 +218,12 @@ def divergence_corner(
...
 
-        divg_d = vf[0, -1, 0] - vf + uf[-1, 0, 0] - uf
+        divg_d = (vf[0, -1, 0] - vf + uf[-1, 0, 0] - uf) * rarea_c
         with horizontal(region[i_start, j_start], region[i_end + 1, j_start]):
-            divg_d -= vf[0, -1, 0]
+            divg_d = (-vf + uf[-1, 0, 0] - uf) * rarea_c
+
         with horizontal(region[i_end + 1, j_end + 1], region[i_start, j_end + 1]):
-            divg_d += vf
-        divg_d *= rarea_c
```
rheacangeo, line 226: This is causing validation to fail because the math is changed. Excellent. We can justify increasing the max_error slightly in the DivergenceCorner error -- please verify all the errors are tiny (e.g. run the test locally with --print_failures). The one jenkins reports is 3.828570515762384e-07, reference: 3.828570515762432e-07, diff: 4.817499887508833e-21, so actually very close to the default relative error metric of 1e-14.

rheacangeo (commented):
Great job. Looks like you chose 2 different approaches for a similar pattern. I think it makes sense to have them both compute the regions independently of the main domain computation, and adjust the error thresholds, but if you prefer to have both pass and keep the same math, that works too.

fv3core/stencils/c_sw.py
jdahm: Oh, I see the problem: `rarea_c` needs to be shifted.

fv3core/stencils/c_sw.py
jdahm: Oh that's a good idea. I'll switch it to that.

rheacangeo (approved):
Looks great!

PR #554, opened 2021-07-07
GTC safety/fxadv pgradc

rheacangeo:
## Purpose

Removing race conditions from stencils with regions. FxAdv corner stencils have self reference with an offset - although in theory there isn't a collision of the actual points, we know from our experience with GTC that those backends will not work with this pattern (in the without regions version, though is likely to also not be ok with regions). The solution to pass the save variable twice works here because the corner stencils aren't writing into the points being read. But if stencils get to be merged such that this is happening, we will likely need full copies of ut and vt variables. 
The regions are removed from p_grad_c, which overcomputes uc and vc both in 1 grid cell (the point that they aren't staggered for, which is undefined in the Fortran code). The Translate test for PGradC is removed. Using selective validation had some issues with DynCore validation and PGradC is a very small stencil test not covering any feature uniquely.  

## Code changes:

- fxadv ut and vt corner computations to not write and read with an offset the same variable. 
- dyn_core removes regions from p_grad_c
- test code for PGradC removed

## Checklist

### DIFF
--- a/fv3core/stencils/dyn_core.py
+++ b/fv3core/stencils/dyn_core.py
@@ -120,25 +120,22 @@ def p_grad_c_stencil(
     Grid variable inputs:
         rdxc, rdyc
     """
-    from __externals__ import hydrostatic, local_ie, local_is, local_je, local_js
+    from __externals__ import hydrostatic
 
     with computation(PARALLEL), interval(...):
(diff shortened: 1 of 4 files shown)


jdahm (approved):
Looks good to me!

PR #558, opened 2021-07-10
Fix regions-related race conditions in `divergence_corner` and `circulation_cgrid`

jdahm:
Also reduces tolerance on `circulation_cgrid` to `5e-13`.

### DIFF
--- a/fv3core/stencils/c_sw.py
+++ b/fv3core/stencils/c_sw.py
@@ -206,8 +206,6 @@ def divergence_corner(
             * 0.5
             * (sin_sg4[0, -1] + sin_sg2)
         )
-        with horizontal(region[:, j_start], region[:, j_end + 1]):
-            uf = u * dyc * 0.5 * (sin_sg4[0, -1] + sin_sg2)
 
         vf = (
             (v - 0.25 * (ua[-1, 0, 0] + ua) * (cos_sg3[-1, 0] + cos_sg1))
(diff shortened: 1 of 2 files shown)


fv3core/stencils/c_sw.py
```diff
@@ -206,24 +206,58 @@ def divergence_corner(
...
             * dxc
             * 0.5
             * (sin_sg3[-1, 0] + sin_sg1)
         )
+
+        divg_d = (vf[0, -1, 0] - vf + uf[-1, 0, 0] - uf) * rarea_c
         with horizontal(region[i_start, :], region[i_end + 1, :]):
             vf = v * dxc * 0.5 * (sin_sg3[-1, 0] + sin_sg1)
+            vf1 = v[0, -1, 0] * dxc[0, -1] * 0.5 * (sin_sg3[-1, -1] + sin_sg1[0, -1])
+            uf1 = (
```
rheacangeo, line 221: It might be nice to in include a TODO of using a function with an argument with an offset, if you think that's a potential future feature. That would help reduce confusion later if someone tries to refactor this and gets overwhelmed with the long math statements.

fv3core/stencils/c_sw.py
```diff
@@ -206,24 +206,58 @@ def divergence_corner(
...
+                * dyc[-1, 0]
+                * 0.5
+                * (sin_sg4[-1, -1] + sin_sg2[-1, 0])
+            )
+            divg_d = (vf1 - vf + uf1 - uf) * rarea_c
+
+        with horizontal(region[:, j_start], region[:, j_end + 1]):
+            uf = u * dyc * 0.5 * (sin_sg4[0, -1] + sin_sg2)
+            uf1 = u[-1, 0, 0] * dyc[-1, 0] * 0.5 * (sin_sg4[-1, -1] + sin_sg2[-1, 0])
+            vf1 = (
```
rheacangeo, line 237: same comment as above

fv3core/stencils/c_sw.py
```diff
@@ -206,24 +206,58 @@ def divergence_corner(
...
+                * 0.5
+                * (sin_sg3[-1, -1] + sin_sg1[0, -1])
+            )
+            divg_d = (vf1 - vf + uf1 - uf) * rarea_c
 
-        divg_d = (vf[0, -1, 0] - vf + uf[-1, 0, 0] - uf) * rarea_c
         with horizontal(region[i_start, j_start], region[i_end + 1, j_start]):
-            divg_d = (-vf + uf[-1, 0, 0] - uf) * rarea_c
+            vf = v * dxc * 0.5 * (sin_sg3[-1, 0] + sin_sg1)
+            uf = u * dyc * 0.5 * (sin_sg4[0, -1] + sin_sg2)
```
rheacangeo, line 252: maybe we should add functions for these?

fv3core/stencils/c_sw.py
```diff
@@ -246,15 +280,18 @@ def circulation_cgrid(
 
     with computation(PARALLEL), interval(...):
         fx = dxc * uc
+        fx1 = dxc[0, -1] * uc[0, -1, 0]
+
         fy = dyc * vc
+        fy1 = dyc[-1, 0] * vc[-1, 0, 0]
 
-        vort_c = fx[0, -1, 0] - fx - fy[-1, 0, 0] + fy
+        vort_c = fx1 - fx - fy1 + fy
```
rheacangeo, line 288: ah this is necessary, for below, accessing a temporary with an offset. this workaround works? awesome.

rheacangeo (approved):
Looks great! As mentioned, I think it would be helpful to add some TODO comments for the repeated math. Great work!

fv3core/stencils/c_sw.py
```diff
@@ -246,15 +280,18 @@ def circulation_cgrid(
 
     with computation(PARALLEL), interval(...):
         fx = dxc * uc
+        fx1 = dxc[0, -1] * uc[0, -1, 0]
+
         fy = dyc * vc
+        fy1 = dyc[-1, 0] * vc[-1, 0, 0]
```
rheacangeo, line 286: Might want to add a comment to indicate this is done this way to prevent a race condition with the regions access below

fv3core/stencils/c_sw.py
jdahm: Yeah, it's that temporaries cannot be referenced with any offsets, but the temporary itself can represent an offset'd calculation :)

PR #559, opened 2021-07-13
Add GridData, DampingCoefficients classes

mcgibbon:
Classes are used in currently-refactored inits. d_sw is partially refactored, there's a bit more work to remove the old Grid class from it entirely but I'd like to push the changes so far through and complete d_sw in a follow-on PR.

## Purpose

We looked at what grid variables are generated by the grid savepoints and where those grid variables are used to determine some container classes. Variables that are used in only one place will later be calculated in that init routine, for now I've converted those to access global grid so they stay out of the init signatures. These should be cleaned up when the grid generation code is available.

Some grid variables not used by the dynamical core but used by the physics are included as properties that raise NotImplementedError. This includes `lat`, `lon`, `ptop`, `ak`, and `bk`.

## Code changes:

- Created `GridData` class composed of `ContravariantGridData`, `AngleGridData`, `HorizontalGridData`, and `VerticalGridData`
- Created `DampingCoefficients` class
- Defined properties on `Grid` to retrieve these new classes
- Refactored DivergenceDamping, AGrid2BGridFourthOrder, DGrid2AGrid2CGridVectors, DelnFlux, DelnFluxNoSG, FiniteVolumeFluxPrep, XTP_U, YTP_V, and partially DGridShallowWaterLagrangianDynamics to use new data container classes

## Checklist


FlorianDeconinck (approved):
Maybe tag with a TODO the docstring we expect to fill on the remaining properties

fv3core/stencils/a2b_ord4.py
```diff
@@ -458,16 +459,7 @@ class AGrid2BGridFourthOrder:
...
-        agrid2,
-        bgrid1,
-        bgrid2,
-        dxa,
-        dya,
-        edge_n,
-        edge_s,
-        edge_e,
-        edge_w,
+        grid_data: GridData,
```
rheacangeo, line 462: yay!!

fv3core/stencils/a2b_ord4.py
```diff
@@ -11,11 +11,12 @@
     sqrt,
 )
 
+import fv3core._config as spec
```
rheacangeo, line 14: When we talked I thought you had wanted to pass the legacy grid object as an argument to remove this? I don't object-- I see here they are used for getting the grid variables we actually want to compute in this module. So it could be a handy way to search and find those. Just want to check I am understanding the intention correctly here?

fv3core/stencils/fxadv.py
```diff
@@ -376,31 +376,20 @@ class FiniteVolumeFluxPrep:
...
-        self._sin_sg3 = sin_sg3
-        self._sin_sg4 = sin_sg4
+        self._dx = grid_data.dx
+        self._dy = grid_data.dy
+        self._rdxa = grid_data.rdxa
+        self._rdya = grid_data.rdya
+        self._cosa_u = grid_data.cosa_u
+        self._cosa_v = grid_data.cosa_v
+        self._rsin_u = grid_data.rsin_u
+        self._rsin_v = grid_data.rsin_v
```
rheacangeo, line 388: these are part of the Contravariant piece, but are directly accessible, that's useful

fv3core/utils/grid.py
```diff
@@ -345,15 +346,6 @@ def full_origin(self, add: Tuple[int, int, int] = (0, 0, 0)):
...
 
-    def default_origin(self):
-        # This only exists as a reminder because devs might
-        # be used to writing "default origin"
-        # if it's no longer useful please delete this method
-        raise NotImplementedError(
-            "This has been renamed to `full_origin`, update your code!"
-        )
-
-    # TODO, expand to more cases
```
rheacangeo, line 356: haha this does seem like a good time for this!

fv3core/utils/grid.py
```diff
@@ -371,6 +363,368 @@ def horizontal_starts_from_shape(self, shape):
...
+    Terms defining the vertical grid.
+
+    Eulerian vertical grid is defined by p = ak + bk * p_ref
+    """
+
+    # TODO: refactor so we can init with this,
+    # instead of taking it as an argument to DynamicalCore
+    # we'll need to initialize this class for the physics
+    @property
+    def ptop(self) -> float:
```
rheacangeo, line 468: I always thought this was a runtime constant, but makes sense to be defined in init when we can

rheacangeo (approved):
It's exciting to see these variable lists compressing and see the structure start to build around grid related variables. We'll want to continue to think about/plan for making it very clear what is a constant and what is not, but that's outside the scope of this pr. Looks great.

fv3core/stencils/a2b_ord4.py
mcgibbon: Yes, that was the original plan. This will make it easier to search for those, and also let me rely on "grid is not an argument to the object" as a marker that the object has at least undergone the initial round of init refactors.

fv3core/utils/grid.py
mcgibbon: Yeah, I could imagine someone e.g. wanting to run a 2D shallow water model alongside a fully 3D model (e.g. to train a 2D ML model online), which would necessitate different ptop's. Even the physical constants might make sense to convert later for this kind of use case.

PR #561, opened 2021-07-13
remove race condition of ut and vt from ke_horizontal_vorticity 

rheacangeo:
## Purpose
There was a race condition in ke_horizontal_vorticity because at the corners, ut and vt are read with an offset, then ut/vt are written to.  Split the computation of ke from that of vorticity to prevent this race condition. 
Also removed the need for ke to be read into the corner_ke function

## Code changes:

- ke_horizontal_vorticity in d_sw split into 2 stencils
- corners.corner_ke does not take ke as an argument

### DIFF
--- a/fv3core/stencils/d_sw.py
+++ b/fv3core/stencils/d_sw.py
@@ -89,13 +89,13 @@ def all_corners_ke(ke, u, v, ut, vt, dt):
 
     # Assumption: not __INLINED(grid.nested)
     with horizontal(region[i_start, j_start]):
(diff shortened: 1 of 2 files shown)


jdahm (approved):
Great work! This makes it much easier to read too.

rheacangeo:
launch jenkins

rheacangeo:
This actually passed, but ran into an error before completing the slurm job
MPID_nem_gni_check_localCQ(): GNI_CQ_EVENT_TYPE_POST had error (SOURCE_SSID:AT_MDD_INV:CPLTN_DREQ)
hopefully is not repeatable

PR #562, opened 2021-07-13
remove race condition from C_SW transportdelp updatevorticity ke

rheacangeo:
## Purpose

Using fill corners and the nonhydrofluxes accesses delp, pt and w at and offset and then writes to them, a race condition. This PR splits apart the transportdelp stencil into 4 parts to avoid this and enable the transition to GTC. 

## Code changes:

- Split the transportdelp_updatevorticity_key into multiple stencils in the transport part

### DIFF
--- a/fv3core/stencils/c_sw.py
+++ b/fv3core/stencils/c_sw.py
@@ -46,28 +46,17 @@ def absolute_vorticity(vort: FloatField, fC: FloatFieldIJ, rarea_c: FloatFieldIJ
         vort[0, 0, 0] = fC + rarea_c * vort
 
 
-def fill_corners_x_delp_pt_w(
+def fill_corners_delp_pt_w(
     delp: FloatField,
     pt: FloatField,
     w: FloatField,
 ):
+    from __externals__ import fill_corners_func
 
     with computation(PARALLEL), interval(...):
-        delp = corners.fill_corners_2cells_x(delp)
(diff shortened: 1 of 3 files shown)


fv3core/stencils/c_sw.py
```diff
@@ -47,26 +46,46 @@ def absolute_vorticity(vort: FloatField, fC: FloatFieldIJ, rarea_c: FloatFieldIJ
...
-    fx2 = fx1 * fx2
-    return fx, fx1, fx2
+def fill_corners_x_delp_pt_w(
+    delp: FloatField,
+    pt: FloatField,
+    w: FloatField,
+):
+
+    with computation(PARALLEL), interval(...):
+        delp = corners.fill_corners_2cells_x(delp)
```
jdahm, line 56: If you want to reduce code duplication slightly, `corners.fill_corners_2cells_x` can instead be an external `corners_func` and you can compile the same stencil twice: once for the `_x` variant and once for the `_y`.

rheacangeo: We had that in master previously, I thought there was a reason we removed it.

rheacangeo: Oh, that was because it was happening at runtime and was causing slowness, but it was not due to using the x and y as functions but rather the runtime recompilation, which this should not face. updated to use your suggestion.

jdahm (approved):
Looks great!

PR #572, opened 2021-07-27
replace SimpleNamespace namelist with dataclass

mcgibbon:
## Purpose

Dataclasses have worked out very well for storing and organizing grid variables. This PR turns the existing flat SimpleNamespace namelist into an identical object with explicit typing for its attributes, using a dataclass. This initial refactor will make it easier to organize groups of namelist attributes into smaller objects.

## Code changes:

- Replaced `namelist` SimpleNamespace with an instance of a new `Namelist` dataclass

## Checklist

### DIFF
--- a/fv3core/_config.py
+++ b/fv3core/_config.py
@@ -1,5 +1,6 @@
+import dataclasses
 import os
-from types import SimpleNamespace
+from typing import Tuple
 
 import f90nml
 
@@ -8,7 +9,6 @@ from fv3core.utils.grid import Grid
 
 
 grid = None
-namelist = SimpleNamespace()
 
 # Global set of namelist defaults
 namelist_defaults = {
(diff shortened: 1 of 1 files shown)


fv3core/_config.py
```diff
@@ -65,6 +65,314 @@
...
+    # fvmxl: Any
+    # ldebug: Any
+    grid_type: int = 0
+    do_f3d: bool = False
+    inline_q: bool = False
+    do_skeb: bool = False  # save dissipation estimate
+    use_logp: bool = False
+    moist_phys: bool = True
+    check_negative: bool = False
+    # gfdl_cloud_mucrophys.F90
```
elynnwu, line 325:
```suggestion
    # gfdl_cloud_microphys.F90
```

elynnwu (approved):
Looks good! I like that namelist is flat, but I wonder if we should have additional information as in the yml file (e.g., when one goes to call `namelist.hord_dp`, there's a way to see that this is a namelist option inside `fv_core_nml`, and it stands for horizontal advection scheme for mass). It's not really an issue now, but as we grow to have more options and include physics, I think it will be nice to be able to tell whether this namelist option is for dycore or a specific physics scheme easily.

mcgibbon:
@elynnwu yes, my next task will be to take this flat class and decompose it into individual parts that can be broken off and used to initialize parts of the model. Will be similar to what we described for calling physics routines with containers. Having this class will also enable us to rename some of these variables, with central documentation for their previous name.

PR #573, opened 2021-07-28
Create config classes for d_sw, riemann solver, and AcousticDynamics

mcgibbon:
## Purpose

The namelist currently exists only as a flat namespace, with individual parameters being passed to many internal routines to handle data containerization. This can result in many arguments (when data is not containerized), or being unable to determine what namelist parameters are being used (when it is containerized).

This PR creates configuration container classes for AcousticDynamics, DGridShallowWaterLagrangianDynamics, and RiemannSolver3 which contain only the data they require. These can be retrieved as attributes from the top-level namelist.

## Code changes:

- Added classes AcousticDynamicsConfig, DGridShallowWaterLagrangianDynamicsConfig, and RiemannConfig
- Added attributes `riemann`, `d_grid_shallow_water`, and `acoustic_dynamics` to `Namelist`

## Checklist


fv3core/_config.py
```diff
@@ -73,6 +73,93 @@
 DEFAULT_BOOL = False
 
 
+@dataclasses.dataclass(frozen=True)
+class RiemannConfig:
```
rheacangeo, line 77: Config is pretty generic. It might be helpful for users to explicitly call it NamelistConfig, or just Namelist (e.g. RiemannNamelist), so it is clear what kind of a configuration it is and where it typically comes from.

rheacangeo (commented):
Nice organization of the namelist! I like that you added properties to the parent object such that deep access is not required. My main comment is that I think we should call this thing Namelist rather than Config throughout. There are other types of config (e.g. gt4py config) that aren't set by the same mechanism and shouldn't be part of this object. Model developers are familiar with the idea of a namelist, and it's really nice to know right away a config option is controlled by a specific file on disk.

fv3core/_config.py
mcgibbon: Maybe, I wanted to separate the word Namelist from Config though to create a separation of "this class corresponds directly to the Fortran namelist, uses the same variable names, and has a similar structure" from "these are internal new configuration classes where we might use different names, consolidate multiple namelist settings, or change types of settings (e.g. integers to descriptive strings)" with functions that map between the two. We can chat about it.

mcgibbon: Also consider that an analysis user might want to use the Riemann object directly outside of a model, and they'd be making a RiemannConfig in this case. The arguments could be hard-coded then, as opposed to coming from any sort of file. It being "config" is really all we know for sure about this data.

rheacangeo: I can see the intent here, I worry it makes it unnecessarily confusing for a new user. Something being a Namelist object doesn't necessarily mean it has to come from a file on disk, but indicates it is part of the model algorithm options configuration. The Riemann is one of the few modules that could have the potential to get used in other contexts, most are highly integrated and rely on other parts of this model.  If someone is in the riemann solver code, it should be obvious to them where the configuration settings came from and how to change them. If a configuration object combines settings set in a variety of sources, that is not super user-friendly for them understanding the model. We already have quite a handful of varieties of 'configurations' defined and set in different places in the model, and I think 'Config' is not specific enough. 
I also think we shouldn't be internally combining and changing these options, they are already a lot to understand, and adding additional mappings and interpretation layers will make it difficult for anyone new to work on and improve this code. It will also make it easier to make configuration mistakes. Especially since we haven't ported all existing options, changing the way objects are configured will make doing that work harder than it needs to be. We should use the settings as provided by the Fortran model and keep it as straightforward as possible at this stage, in my opinion. 
I'm a fan of using configuration classes as you have laid out though, with classes specific to a particular part of the model. That seems useful.

rheacangeo (approved):
I don't want to drag out this PR unnecessarily just over a naming convention. I think we disagree somewhat on the vision of who will use this and how and what is easiest for them to deal with, but calling something a 'Config' rather than something more specific is not a big deal in this phase of development. If we ever get users and get feedback on this, and they would like a change, we can do that then.

mcgibbon:
Yes, I think this should be a point we specifically ask about when getting feedback in the future.

PR #576, opened 2021-07-29
remove halo update of phis at start of dynamics step

mcgibbon:
## Purpose

The dynamics step currently halo updates surface geopotential (phis) each timestep, even though it is a constant. This PR removes the halo update entirely, since phis is already halo updated by its initialization code.

## Code changes:

- Removed halo update of phis at the start of DynamicalCore.step_dynamics

## Checklist

### DIFF
--- a/fv3core/stencils/fv_dynamics.py
+++ b/fv3core/stencils/fv_dynamics.py
@@ -352,11 +352,6 @@ class DynamicalCore:
             self.grid, namelist, DynamicalCore.NQ, self._pfull
         )
 
-        phis_spec = self.grid.get_halo_update_spec(
-            phis.data.shape, phis.origin, utils.halo, phis.dims
-        )
-        self._phis_halo_updater = self.comm.get_scalar_halo_updater([phis_spec])
-
         full_xyz_spec = self.grid.get_halo_update_spec(
             self.grid.domain_shape_full(add=(1, 1, 1)),
             self.grid.compute_origin(),
@@ -419,8 +414,6 @@ class DynamicalCore:
         state.ak = self._ak
         state.bk = self._bk
         last_step = False
-        if self.do_halo_exchange:
-            self._phis_halo_updater.update([state.phis_quantity])
         compute_preamble(
             state,
             self.grid,


rheacangeo:
We should be careful here -- I spot at least one spot in remapping where phis actually does change with some namelist options (consv > consv_min). We don't support that at the moment, but it will not be obvious to someone adding that option that they also need to add the halo update if they are modifying phis. There may be other places in the code as well. How much of a speedup is there?

rheacangeo:
Ah the case I saw, there's a name change inside of remapping. We should be extra sure with the fortran code that phis is not changing.

rheacangeo:
Looks like phis is an inout for dyn_core for specific test cases:  case9_forcing_1 and case9_forcing2, triggered with test_case == 9 and USE_OLD compiler flag. So it could potentially be done if we aren't going to replicate the test cases. But we should try to not do this kind of optimization without a full search and reporting of the use cases in the Fortran code.

mcgibbon:
> Ah the case I saw, there's a name change inside of remapping. We should be extra sure with the fortran code that phis is not changing with some option though.

I haven't heard of an option in any weather/climate model to change surface height with time, but to be sure I grepp'd through the Fortran repo with `git grep phis . | grep intent`. The only in-timestep routines that have `inout` intents are in fv_dynamics.F90 and dyn_core.F90 and in neither place is there an assignment (so they could be refactored to `intent(in)`). It is also written during initialization, e.g. in fv_surf_map.F90 and test_cases.F90. In fv_nesting.F90 it is also not modified (not sure when this is called).

rheacangeo (commented):
Ok, with the fuller evidence that it is unlikely to bite someone, we can do this -- with the awareness if someone adds 'test case 9' they will likely be missing a piece that makes it work and might be very confused as to why. It could be helpful to add a comment at line 422 about that. I think generally trying to show we are 'faster than fortran', we should aim to run the same things to have a comparison that is equivalent, rather than removing code to get speed.  It is also very easy to miss a use case that our particular tests don't exercise, so we should tread lightly with this kind of optimization. But in this case I agree, appears unnecessary to do every timestep.

mcgibbon:
If we do implement `test_cases.F90`, that code is only called on init, so an implementer is unlikely to come across a comment at the start of the dynamics loop. And in that case we'd want the halo update to occur once, on init, and not every step. Halo updates of phis do occur in test_cases.F90, e.g. on lines 1565 and 1624.

Could I handle this instead by removing phis from call time and instead using the value passed on init, and then halo updating phis only during initialization? This should work for any initialization case, implemented or not.

mcgibbon:
Could I handle this instead by removing phis from call time and instead using the value passed on init, and then halo updating phis only during initialization? This should work for any initialization case, implemented or not, and doesn't depend on the writer noticing our comment.

If we do implement `test_cases.F90`, that code is only called on init, so an implementer is unlikely to come across a comment at the start of the dynamics loop. And in that case we'd want the halo update to occur once, on init, and not every step. Halo updates of phis do occur in test_cases.F90, e.g. on lines 1565 and 1624.

The case9_forcing(1/2) code called during run time initializes phis in the halos as well, so it shouldn't require a halo update.

rheacangeo:
The test case code is called in the acoustic timestep, in dyn_core.F90,  `if (test_case==9) call case9_forcing1(phis, time_total)`, not just the init. This seems like it must be the reason  halo update on phis exists in the code every timestep. Of course it is always possible it is leftover from some code that has since been removed. 
Putting a halo update in the init sounds fine, as does, running the init halo update wherever phis is initially defined.

rheacangeo (approved):
wait a minute, I don't actually see a phis halo update happening in the Fortran code, lol. This must have been a mistake.

rheacangeo:
Turns out this was added to play nice with the wrapped fortran, which doesn't provide halo data. But Jeremy has now made a new jira issue to make sure the halo updating of phis happens in the runfile, rather than in step_dynamics

PR #578, opened 2021-07-31
divergence corner thresholds

rheacangeo:
## Purpose
The changes to DivergenceCorner that just reordered math statements needs even looser error bounds for other datasets (c48_6ranks_standard, c12_54ranks_standard


jdahm:
Ugh. We can just set back the math if that makes this easier. I don't believe it was critical to removing the regions, rather it was just a purely mathematical cancellation.

rheacangeo:
The error doesn't seem to propagate upward badly in this case at least. We should be able to reorder math as long as we don't end up with huge errors on the top level as a result I think. But if you prefer we change the math back to what it was, I'm open.

rheacangeo:
launch jenkins

FlorianDeconinck (approved):
Original commit https://github.com/VulcanClimateModeling/fv3core/pull/558 removes a `y = x-x` type of calculation that could lead to those difference.

Calculation re-ordering is correct, I agree with changing the threshold.

PR #579, opened 2021-08-03
use NullComm to disable halo updates instead of config option

mcgibbon:
## Purpose

Currently we disable halo updates during compilation by passing a global configuration option. This PR refactors the code to disable these instead  by making use of a mock of the mpi4py comm object which does no communication.

## Code changes:

- Removed `do_halo_exchange` options from several classes
- Added NullComm class in `fv3core.util.null_comm`. This should later be moved to `fv3gfs.util`.

## Checklist

### DIFF
--- a/examples/standalone/runfile/dynamics.py
+++ b/examples/standalone/runfile/dynamics.py
@@ -2,13 +2,13 @@
 
 import copy
 import json
+import os
 from argparse import ArgumentParser, Namespace
 from datetime import datetime
 from typing import Any, Dict, List
 
 import numpy as np
 import serialbox
-import yaml
 from mpi4py import MPI
 
 import fv3core
@@ -37,7 +37,7 @@ def parse_args() -> Namespace:
         "backend",
         type=str,
         action="store",
-        help="path to the namelist",
+        help="gt4py backend to use",
     )
     parser.add_argument(
         "hash",
@@ -182,12 +182,7 @@ if __name__ == "__main__":
 
         spec.set_namelist(args.data_dir + "/input.nml")
 
-        experiment_name = yaml.safe_load(
-            open(
-                args.data_dir + "/input.yml",
-                "r",
-            )
-        )["experiment_name"]
+        experiment_name = os.path.basename(os.path.normpath(args.data_dir))
 
         # set up of helper structures
         serializer = serialbox.Serializer(
--- a/examples/standalone/runfile/dynamics.py
+++ b/examples/standalone/runfile/dynamics.py
@@ -14,15 +14,12 @@ from mpi4py import MPI
 import fv3core
 import fv3core._config as spec
 import fv3core.testing
-import fv3core.utils.global_config as global_config
 import fv3gfs.util as util
+from fv3core.utils.null_comm import NullComm
 
 
 def parse_args() -> Namespace:
-    usage = (
-        "usage: python %(prog)s <data_dir> <timesteps> <backend> <hash> <halo_exchange>"
-    )
-    parser = ArgumentParser(usage=usage)
+    parser = ArgumentParser()
 
     parser.add_argument(
         "data_dir",
@@ -182,7 +179,6 @@ if __name__ == "__main__":
         fv3core.set_backend(args.backend)
         fv3core.set_rebuild(False)
         fv3core.set_validate_args(False)
-        global_config.set_do_halo_exchange(not args.disable_halo_exchange)
 
         spec.set_namelist(args.data_dir + "/input.nml")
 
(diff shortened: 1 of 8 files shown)


FlorianDeconinck (approved):
LGTM

fv3core/utils/null_comm.py
```diff
@@ -0,0 +1,53 @@
...
+
+class NullComm:
+    """
+    A class with a subset of the mpi4py Comm API, but which
+    'receives' zeros instead of using MPI.
+    """
+
+    def __init__(self, rank, total_ranks):
+        self.rank = rank
+        self.total_ranks = total_ranks
```
twicki, line 18: is there a reason not to keep with `size` as the default communication syntax for this?

fv3core/utils/null_comm.py
```diff
@@ -0,0 +1,53 @@
...
+
+    def wait(self):
+        if self._recvbuf is not None:
+            self._recvbuf[:] = 0.0
+
+
+class NullComm:
+    """
+    A class with a subset of the mpi4py Comm API, but which
+    'receives' zeros instead of using MPI.
```
twicki, line 13: what is the reason to write `0` instead of `NaN`? Is this just to make sure that the computation does something realistic (even though not valid)?

examples/standalone/runfile/dynamics.py
```diff
@@ -182,27 +179,21 @@ def collect_data_and_write_to_file(
...
 
         spec.set_namelist(args.data_dir + "/input.nml")
 
-        experiment_name = yaml.safe_load(
-            open(
-                args.data_dir + "/input.yml",
-                "r",
-            )
-        )["experiment_name"]
+        experiment_name = os.path.basename(os.path.normpath(args.data_dir))
```
twicki, line 185: I see this being shorter, but is it not significantly less safe to rely on the folder name to be something specific compared to reading the `yml` file?

fv3core/utils/null_comm.py
mcgibbon: I used `total_ranks` to mirror `Partitioner.total_ranks`, which this is acting in place of. Normally the CubedSphereCommunicator doesn't have size, you get it from `cube.comm` instead.

fv3core/utils/null_comm.py
mcgibbon: It's so it doesn't crash when we're compiling. If NaNs also wouldn't crash, that would be better. I don't know whether they do or don't but it's hard to test.

examples/standalone/runfile/dynamics.py
mcgibbon: The point wasn't that it's shorter, it's that we actually do not have this yaml file in the data download. I have no idea where this yaml file comes from. This name is also the only thing we use from that yaml file, so we should probably just be passing the experiment name as an argument instead of a path to a yaml file with an experiment name. I     
 can change it to do that instead, but this involves changing a jenkins plan that is not tested on PRs so it's more risky, and debugging is a many-hour turnaround.

fv3core/utils/null_comm.py
mcgibbon: I'll add a `fill_value=0.` argument.

twicki (approved):

examples/standalone/runfile/dynamics.py
mcgibbon: I will restore the use of `input.yml` since we'll probably use it in the long term. This will happen by moving the "move experiment.yml to input.yml" command into the make target instead of in the jenkins scripts. In the long term, we want to rename the experiment yml's to `input.yml`. I'll make a JIRA task for this.

fv3core/utils/null_comm.py
mcgibbon: Added.

examples/standalone/runfile/dynamics.py
mcgibbon: Since you have the latter part of this plan implemented and in a PR, instead I will revert the code here and we can use your full solution. This was changed only to fix the make target working locally, so it should be OK to leave that not-working (requiring a manual `mv`) for another few days/week.

mcgibbon:
As Rhea suggested, I disabled the halo exchange in `run_on_daint.sh` and ran the performance tests to test this - they completed successfully. https://jenkins.ginko.ch/job/fv3core-performance-profile/307/backend=gtx86,experiment=c128_6ranks_baroclinic,slave=daint_submit/console

fv3core/utils/null_comm.py
```diff
@@ -0,0 +1,61 @@
...
```
mcgibbon, line 13:
```suggestion
    'receives' a fill value (default zero) instead of using MPI.
```

PR #580, opened 2021-08-06
Feature/rename xtp ytp

mcgibbon:
## Purpose

Describe the purpose of this PR and the motivation if relevant to understanding. Include links to related issues, bugs or features.

Remove the sections below which do not apply.

## Code changes:

- Provide a list of relevant code changes and the side effects they have on other code.

## Requirements changes:

- Provide a list of any changes made to requirements, e.g. changes to files requirements*.txt, constraints.txt, setup.py, pyproject.toml, pre-commit-config.yaml and a reason if not included in the Purpose section (e.g. incompatibility, updates, etc)

## Infrastructure changes:

- Provide a list of changes that impact the infrastructure around running the code -- that is, changes to Makefiles, docker files, git submodules, or .jenkins (testing infrastructure changes). If Jenkins plans are also being manually changed, indicate that as well.

## Checklist

### DIFF
--- a/fv3core/stencils/d_sw.py
+++ b/fv3core/stencils/d_sw.py
@@ -19,8 +19,8 @@ from fv3core.stencils.delnflux import DelnFluxNoSG
 from fv3core.stencils.divergence_damping import DivergenceDamping
 from fv3core.stencils.fvtp2d import FiniteVolumeTransport
 from fv3core.stencils.fxadv import FiniteVolumeFluxPrep
-from fv3core.stencils.xtp_u import XTP_U
-from fv3core.stencils.ytp_v import YTP_V
+from fv3core.stencils.xtp_u import XTP_U, xtp_u
+from fv3core.stencils.ytp_v import YTP_V, ytp_v
 from fv3core.utils.grid import DampingCoefficients, GridData, GridIndexing, axis_offsets
 from fv3core.utils.typing import FloatField, FloatFieldIJ, FloatFieldK
 from fv3gfs.util import X_DIM, X_INTERFACE_DIM, Y_DIM, Y_INTERFACE_DIM, Z_DIM
@@ -118,7 +118,30 @@ def all_corners_ke(ke, u, v, ut, vt, dt):
     return ke
 
 
-def pressure_and_vbke(
+@gtscript.function
+def pressure_update(
+    gx: FloatField,
+    gy: FloatField,
+    rarea: FloatFieldIJ,
+    fx: FloatField,
+    fy: FloatField,
+    pt: FloatField,
+    delp: FloatField,
+):
+    from __externals__ import inline_q, local_ie, local_is, local_je, local_js
+
+    # TODO: local region only needed for d_sw halo validation
+    # use selective validation instead
+    if __INLINED(inline_q == 0):
+        with horizontal(region[local_is : local_ie + 1, local_js : local_je + 1]):
+            # pt = flux_integral(pt, delp, gx, gy, rarea)
+            pt = pt * delp + flux_increment(gx, gy, rarea)
+            delp = delp + flux_increment(fx, fy, rarea)
+            pt = pt / delp
+    return pt, delp
+
+
+def kinetic_energy_update(
     gx: FloatField,
     gy: FloatField,
     rarea: FloatFieldIJ,
@@ -130,27 +153,27 @@ def pressure_and_vbke(
     uc: FloatField,
     cosa: FloatFieldIJ,
     rsina: FloatFieldIJ,
+    v: FloatField,
     vt: FloatField,
-    vb: FloatField,
-    dt4: float,
-    dt5: float,
+    u: FloatField,
+    ut: FloatField,
+    ke: FloatField,
+    dx: FloatFieldIJ,
+    dxa: FloatFieldIJ,
+    rdx: FloatFieldIJ,
+    dy: FloatFieldIJ,
+    dya: FloatFieldIJ,
+    rdy: FloatFieldIJ,
+    dt: float,
 ):
-    from __externals__ import inline_q, local_ie, local_is, local_je, local_js
-
     with computation(PARALLEL), interval(...):
-        # TODO: only needed for d_sw validation
-        if __INLINED(inline_q == 0):
-            with horizontal(region[local_is : local_ie + 1, local_js : local_je + 1]):
-                # pt = flux_integral(pt, delp, gx, gy, rarea)
-                pt = pt * delp + flux_increment(gx, gy, rarea)
-                delp = delp + flux_increment(fx, fy, rarea)
-                pt = pt / delp
-        vb = vbke(vc, uc, cosa, rsina, vt, dt4, dt5)
-
-
-@gtscript.function
-def ke_from_bwind(ke, ub, vb):
-    return 0.5 * (ke + ub * vb)
+        pt, delp = pressure_update(gx, gy, rarea, fx, fy, pt, delp)
+        v_on_cell_corners = interpolate_v_to_cell_corners(vc, uc, cosa, rsina, vt, dt)
+        advected_v = ytp_v(v_on_cell_corners, v, dy, dya, rdy)
+        u_on_cell_corners = interpolate_u_to_cell_corners(uc, vc, cosa, rsina, ut, dt)
+        advected_u = xtp_u(u_on_cell_corners, u, dx, dxa, rdx)
+        ke = 0.5 * (u_on_cell_corners * advected_u + v_on_cell_corners * advected_v)
+        ke = all_corners_ke(ke, u, v, ut, vt, dt)
 
 
 def ub_vb_from_vort(
@@ -352,21 +375,6 @@ def heat_source_from_vorticity_damping(
                 v = v - ut
 
 
-def update_ke(
-    ke: FloatField,
-    u: FloatField,
-    v: FloatField,
-    ub: FloatField,
-    vb: FloatField,
-    ut: FloatField,
-    vt: FloatField,
-    dt: float,
-):
-    with computation(PARALLEL), interval(...):
-        ke = ke_from_bwind(ke, ub, vb)
-        ke = all_corners_ke(ke, u, v, ut, vt, dt)
-
-
 def update_horizontal_vorticity(
     u: FloatField,
     v: FloatField,
@@ -466,46 +474,29 @@ def get_column_namelist(config: DGridShallowWaterLagrangianDynamicsConfig, npz):
 
 
 @gtscript.function
-def ubke(uc, vc, cosa, rsina, ut, dt4, dt5):
+def interpolate_u_to_cell_corners(uc, vc, cosa, rsina, ut, dt):
     from __externals__ import i_end, i_start, j_end, j_start
 
-    ub = dt5 * (uc[0, -1, 0] + uc - (vc[-1, 0, 0] + vc) * cosa) * rsina
+    ub = 0.5 * dt * (uc[0, -1, 0] + uc - (vc[-1, 0, 0] + vc) * cosa) * rsina
     # if __INLINED(spec.namelist.grid_type < 3):
     with horizontal(region[:, j_start], region[:, j_end + 1]):
-        ub = dt4 * (-ut[0, -2, 0] + 3.0 * (ut[0, -1, 0] + ut) - ut[0, 1, 0])
+        ub = 0.25 * dt * (-ut[0, -2, 0] + 3.0 * (ut[0, -1, 0] + ut) - ut[0, 1, 0])
     with horizontal(region[i_start, :], region[i_end + 1, :]):
-        ub = dt5 * (ut[0, -1, 0] + ut)
+        ub = 0.5 * dt * (ut[0, -1, 0] + ut)
 
     return ub
 
 
-def mult_ubke(
-    vb: FloatField,
-    ke: FloatField,
-    uc: FloatField,
-    vc: FloatField,
-    cosa: FloatFieldIJ,
-    rsina: FloatFieldIJ,
-    ut: FloatField,
-    ub: FloatField,
-    dt4: float,
-    dt5: float,
-):
-    with computation(PARALLEL), interval(...):
-        ke = vb * ub
-        ub = ubke(uc, vc, cosa, rsina, ut, dt4, dt5)
-
-
 @gtscript.function
-def vbke(vc, uc, cosa, rsina, vt, dt4, dt5):
+def interpolate_v_to_cell_corners(vc, uc, cosa, rsina, vt, dt):
     from __externals__ import i_end, i_start, j_end, j_start
 
-    vb = dt5 * (vc[-1, 0, 0] + vc - (uc[0, -1, 0] + uc) * cosa) * rsina
+    vb = 0.5 * dt * (vc[-1, 0, 0] + vc - (uc[0, -1, 0] + uc) * cosa) * rsina
     # ASSUME : if __INLINED(namelist.grid_type < 3):
     with horizontal(region[i_start, :], region[i_end + 1, :]):
-        vb = dt4 * (-vt[-2, 0, 0] + 3.0 * (vt[-1, 0, 0] + vt) - vt[1, 0, 0])
+        vb = 0.25 * dt * (-vt[-2, 0, 0] + 3.0 * (vt[-1, 0, 0] + vt) - vt[1, 0, 0])
     with horizontal(region[:, j_start], region[:, j_end + 1]):
-        vb = dt5 * (vt[-1, 0, 0] + vt)
+        vb = 0.5 * dt * (vt[-1, 0, 0] + vt)
 
     return vb
 
@@ -657,9 +648,16 @@ class DGridShallowWaterLagrangianDynamics:
             [X_INTERFACE_DIM, Y_INTERFACE_DIM, Z_DIM]
         )
         ax_offsets_b = axis_offsets(grid_indexing, b_origin, b_domain)
-        self._pressure_and_vbke_stencil = FrozenStencil(
-            pressure_and_vbke,
-            externals={"inline_q": config.inline_q, **ax_offsets_b},
+        self._kinetic_energy_update = FrozenStencil(
+            kinetic_energy_update,
+            externals={
+                "inline_q": config.inline_q,
+                "iord": config.hord_mt,
+                "jord": config.hord_mt,
+                "mord": config.hord_mt,
+                "xt_minmax": False,
+                **ax_offsets_b,
+            },
             origin=b_origin,
             domain=b_domain,
         )
@@ -710,19 +708,10 @@ class DGridShallowWaterLagrangianDynamics:
             origin=b_origin,
             domain=b_domain,
         )
-        self._update_ke_stencil = FrozenStencil(
-            update_ke,
-            externals=ax_offsets_full,
-            origin=full_origin,
-            domain=full_domain,
-        )
         self._update_horizontal_vorticity_stencil = FrozenStencil(
             update_horizontal_vorticity,
             origin=full_origin,
             domain=full_domain,
-        )
-        self._mult_ubke_stencil = FrozenStencil(
-            mult_ubke, externals=ax_offsets_b, origin=b_origin, domain=b_domain
         )
         self._damping_factor_calculation_stencil = FrozenStencil(
             delnflux.calc_damp, origin=(0, 0, 0), domain=(1, 1, grid_indexing.domain[2])
@@ -901,10 +890,7 @@ class DGridShallowWaterLagrangianDynamics:
             mfy=self._tmp_fy,
         )
 
-        dt5 = 0.5 * dt
-        dt4 = 0.25 * dt
-
-        self._pressure_and_vbke_stencil(
+        self._kinetic_energy_update(
             self._tmp_gx,
             self._tmp_gy,
             self.grid.rarea,
@@ -916,43 +902,17 @@ class DGridShallowWaterLagrangianDynamics:
             uc,
             self.grid.cosa,
             self.grid.rsina,
-            self._tmp_vt,
-            self._tmp_vb,
-            dt4,
-            dt5,
-        )
-
-        advected_v = self._tmp_ub
-
-        self.ytp_v(self._tmp_vb, v, advected_v)
-
-        # these should become separate variables when inside gt4py stencils,
-        # just aliasing memory here for "optimization" (that the Fortran does)
-        u_on_cell_corners = self._tmp_ub
-
-        self._mult_ubke_stencil(
-            self._tmp_vb,
-            self._tmp_ke,
-            uc,
-            vc,
-            self.grid.cosa,
-            self.grid.rsina,
-            self._tmp_ut,
-            advected_v,
-            dt4,
-            dt5,
-        )
-
-        self.xtp_u(self._tmp_ub, u, self._tmp_vb)
-
-        self._update_ke_stencil(
-            self._tmp_ke,
-            u,
             v,
-            self._tmp_ub,
-            self._tmp_vb,
-            self._tmp_ut,
             self._tmp_vt,
+            u,
+            self._tmp_ut,
+            self._tmp_ke,
+            self.grid.dx,
+            self.grid.dxa,
+            self.grid.rdx,
+            self.grid.dy,
+            self.grid.dya,
+            self.grid.rdy,
             dt,
         )
 
--- a/fv3core/stencils/d_sw.py
+++ b/fv3core/stencils/d_sw.py
@@ -39,20 +39,36 @@ def k_bounds():
 
 
 @gtscript.function
-def flux_component(gx, gy, rarea):
+def flux_increment(gx, gy, rarea):
+    """
+    Args:
+        gx: x-direction flux of some scalar q in units of q * area
+            defined on cell interfaces
+        gy: y-direction flux of some scalar q in units of q * area
+            defined on cell interfaces
+        rarea: 1 / area
+
+    Returns:
+        tendency increment in units of q defined on cell centers
+    """
     return (gx - gx[1, 0, 0] + gy - gy[0, 1, 0]) * rarea
 
 
-@gtscript.function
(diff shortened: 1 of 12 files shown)


fv3core/stencils/divergence_damping.py
```diff
@@ -19,60 +20,56 @@ def damp_tmp(q, da_min_c, d2_bg, dddmp):
...
+
+        delpc = (
+            v_contra_dxc[0, -1, 0]
+            - v_contra_dxc
+            + u_contra_dyc[-1, 0, 0]
+            - u_contra_dyc
+        )
         with horizontal(region[i_start, j_start], region[i_end + 1, j_start]):
-            delpc = ptc[-1, 0, 0] - ptc - vort
+            delpc = u_contra_dyc[-1, 0, 0] - u_contra_dyc - v_contra_dxc
```
jdahm, line 70: `u_contra_dyc` is written to before getting read with offset in the horizontal region inside the stencil, which is currently not allowed by GTC, sorry!

fv3core/stencils/divergence_damping.py
```diff
@@ -19,60 +20,56 @@ def damp_tmp(q, da_min_c, d2_bg, dddmp):
...
+        with horizontal(region[:, j_start], region[:, j_end + 1]):
+            u_contra = u * sin_sg4[0, -1] if vc > 0 else u * sin_sg2
+        u_contra_dyc = u_contra * dyc
+
+        delpc = (
+            v_contra_dxc[0, -1, 0]
+            - v_contra_dxc
+            + u_contra_dyc[-1, 0, 0]
+            - u_contra_dyc
+        )
```
mcgibbon, line 68: @jdahm pointed out we can't have a write before read with offset here, will add a `u_contra_dyc_n1_0_0 = u_contra[-1, 0, 0] * dyc[-1, 0]` and use that instead of reading with offset.

fv3core/stencils/divergence_damping.py
```diff
@@ -19,60 +20,56 @@ def damp_tmp(q, da_min_c, d2_bg, dddmp):
...
+        uc_from_ua = 0.5 * (ua[-1, 0, 0] + ua)
+        # TODO: why do we use uc_from_ua and not just uc?
+        v_contra = contravariant(v, uc_from_ua, cosa_u, sina_u)
+        with horizontal(region[i_start, :], region[i_end + 1, :]):
+            v_contra = v * sin_sg3[-1, 0] if uc > 0 else v * sin_sg1
+        v_contra_dxc = v_contra * dxc
+
+        vc_from_va = 0.5 * (va[0, -1, 0] + va)
+        u_contra = contravariant(u, vc_from_va, cosa_v, sina_v)
+        u_contra = (u - 0.5 * (va[0, -1, 0] + va) * cosa_v) * sina_v
```
mcgibbon, line 58:
```suggestion
```
Forgot to remove this, it duplicates the line above.

fv3core/stencils/divergence_damping.py
mcgibbon: We also discussed potentially doing this by refactoring to a write-with-region before reading with an offset.

mcgibbon:
Very curious that setting those regional points to zero makes only the numpy backend fail.

elynnwu (commented):
Thanks Jeremy for walking us through the code. Here's my first pass, mostly focused on differences between python & fortran. Will come back tomorrow/Friday for more.

fv3core/stencils/d_sw.py
```diff
@@ -323,53 +406,37 @@ def heat_source_from_vorticity_damping(
...
+        u_dx = u * dx
+        v_dy = v * dy
     # TODO(rheag). This computation is required because
     # ut and vt are API fields. If the distinction
     # is removed, so can this computation.
     # Compute the area mean relative vorticity in the z-direction
     # from the D-grid winds.
     with computation(PARALLEL), interval(...):
-        vorticity = rarea * (vt - vt[0, 1, 0] - ut + ut[1, 0, 0])
+        vorticity = rarea * (u_dx - u_dx[0, 1, 0] - v_dy + v_dy[1, 0, 0])
```
elynnwu, line 439: This is `wk` in fortran, which is volume_mean_relative_vorticity

mcgibbon: Yes, sounds good. Fortran reference is [here](https://github.com/VulcanClimateModeling/fv3gfs-fortran/blob/e42cd14e08d6b08910fdec7912a3cc4377145e8a/FV3/atmos_cubed_sphere/model/sw_core.F90#L1370).

fv3core/stencils/d_sw.py
```diff
@@ -85,24 +103,72 @@ def flux_capacitor(
         yflux = yflux + fy
 
 
+@gtscript.function
+def corner_ke(
```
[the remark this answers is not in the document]
mcgibbon: Oh I think I see what you're saying, it's possible this logic is used in multiple places but just not right now because of an unported option. I'll investigate the Fortran code and see if that's the case, if there are I'll return it back or place a comment here with a reference to the unported options.

rheacangeo (commented):
Whew! I'm not done looking at d_sw, but will cover that in the next round. Looking good so far.

fv3core/stencils/fvtp2d.py
```diff
@@ -228,25 +241,17 @@ def __call__(
...
-            self.stencil_transport_flux(
-                fx,
-                self._tmp_fx2,
-                fy,
-                self._tmp_fy2,
-                mfx,
-                mfy,
-            )
-            if (mass is not None) and self.delnflux is not None:
-                self.delnflux(q, fx, fy, mass=mass)
```
rheacangeo, line 241: The part about passing in aliases for xfx and yfx, assigning mfx and mfy to help reduce code duplication makes sense. But this particular part of the removed logic is a little more easy to miss, where the mass needs to be present to call delnflux when mfx and mfy are mass fluxes. This does warrant some explanation I think to help a user out.

fv3core/stencils/d_sw.py
```diff
@@ -449,48 +516,39 @@ def get_column_namelist(config: DGridShallowWaterLagrangianDynamicsConfig, npz):
...
+        )
+    with horizontal(region[i_start, :], region[i_end + 1, :]):
+        ub_contra = 0.5 * (uc_contra[0, -1, 0] + uc_contra)
     with horizontal(region[i_start, :], region[i_end + 1, :]):
-        vb = dt4 * (-vt[-2, 0, 0] + 3.0 * (vt[-1, 0, 0] + vt) - vt[1, 0, 0])
+        vb_contra = 0.25 * (
+            -vc_contra[-2, 0, 0]
+            + 3.0 * (vc_contra[-1, 0, 0] + vc_contra)
+            - vc_contra[1, 0, 0]
+        )
```
rheacangeo, line 547: By removing 'dt5' and dt4 from the way this is calculated, this is more difficult to line up what is happening here with the fortran code. I realize you did this so the words you used with xtp_u and ytp_v could be simple things like 'wind' instead of dt * wind, and making sense of xtp_u and ytp_v is the crux of this PR. But this is a lot of mental gymnastics to put on someone trying to port other options or compare the fortran code to the python code (e.g. if debugging a strange behavior present in python but not fortran). Can xtp_u and ytp_v not still be explained in a similar way, with the dt factor included? I  would be more comfortable with just removing dt4 and dt5 variables in favor of explicitly using '0.5 * dt', '0.25 * dt' in their places here, but that's not the point, I know.  Variables including or not including a 'dt' factor does not drastically change our physical understanding of what's happening.  I am surprised this didn't result in validation errors. It's a small thing, but I think we should be careful. At the very minimum the docstring should include something about how this differs from the calculation of ub and vb in the Fortran code.

fv3core/stencils/xtp_u.py
```diff
@@ -16,82 +16,126 @@
...
-        # Zero corners
-        with horizontal(
-            region[i_start - 1 : i_start + 1, j_start],
-            region[i_start - 1 : i_start + 1, j_end + 1],
-            region[i_end : i_end + 2, j_start],
-            region[i_end : i_end + 2, j_end + 1],
-        ):
-            bl = 0.0
-            br = 0.0
+    Named xtp_u in the original Fortran code.
```
rheacangeo, line 124: This is not true anymore because you've changed the units of what is passed in and what is calculated (specifically when dt gets applied). You should mention this change in the comment.

fv3core/stencils/d_sw.py
```diff
@@ -114,75 +180,92 @@ def pressure_and_vbke(
...
+        # vb_contra = interpolate_vc_to_cell_corners(vc, uc, cosa, rsina, vt)
+        ub_contra, vb_contra = interpolate_uc_vc_to_cell_corners(
+            uc, vc, cosa, rsina, uc_contra, vc_contra
+        )
+        advected_v = advect_v_along_y(vb_contra, v, dy, dya, rdy, dt)
+        advected_u = advect_u_along_x(ub_contra, u, dx, dxa, rdx, dt)
+        # TODO: we see here that ke is not kinetic energy, but kinetic energy * timestep
+        #       refactor or rename to avoid this confusion
+        ke = 0.5 * dt * (ub_contra * advected_u + vb_contra * advected_v)
+        ke = all_corners_ke(ke, u, v, uc_contra, vc_contra, dt)
```
rheacangeo, line 208: I do like seeing this all in one stencil, and am curious about whether it improves performance!

fv3core/stencils/d_sw.py
```diff
@@ -114,75 +180,92 @@ def pressure_and_vbke(
...
-        if dcon[0] > dcon_threshold:
-            # Creating a gtscript function for the ub/vb computation
-            # results in an "NotImplementedError" error for Jenkins
-            # Inlining the ub/vb computation in this stencil resolves the Jenkins error
-            with horizontal(region[local_is : local_ie + 1, local_js : local_je + 2]):
-                ub = vort - vort[1, 0, 0]
-            with horizontal(region[local_is : local_ie + 2, local_js : local_je + 1]):
-                vb = vort - vort[0, 1, 0]
+    # TODO: Why is this if statement here?
+    # Can/should it be moved outside this function?
```
rheacangeo, line 226: This doesn't need to be a function, it's only used once, might as well put it in the stencil and not add a TODO.

rheacangeo (commented):
some more comments. it's a lot to take in!

fv3core/stencils/d_sw.py
mcgibbon: > At the very minimum the docstring should include something about how this differs from the calculation of ub and vb in the Fortran code.

I think this is a very reasonable minimum, and I'll at least do this.

At the namelist options meeting this week, we should discuss our relative prioritization of a) making it easy to understand our model code, and b) making it easy to map between our model code and the Fortran code. We can satisfy both to a certain extent, but these two priorities are inherently at odds with each other in cases where the Fortran code is hard to follow. We could also get feedback from Lucas on this point.

fv3core/stencils/fxadv.py
```diff
@@ -432,76 +490,64 @@ def __call__(
...
+            uc: covariant x-velocity on the C-grid (in)
+            vc: covariant y-velocity on the C-grid (in)
             crx: Courant number, x direction(inout)
             cry: Courant number, y direction(inout)
             x_area_flux: flux of area in x-direction, in units of m^2 (inout)
             y_area_flux: flux of area in y-direction, in units of m^2 (inout)
-            ut: temporary x-velocity transformed from C-grid to D-grid equiv(?) (inout)
-            vt: temporary y-velocity transformed from C-grid to D-grid equiv(?) (inout)
+            uc_contra: contravariant x-velocity on C-grid (out)
+            vc_contra: contravariant y-velocity on C-grid (out)
```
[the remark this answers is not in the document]
mcgibbon: Thanks for catching these.

fv3core/stencils/d_sw.py
```diff
@@ -39,20 +39,38 @@ def k_bounds():
...
+
+    Args:
+        q: any scalar, is replaced with something in units of q * delp (inout)
+        delp: pressure thickness of layer (in)
+        gx: x-flux of q in units of q * Pa * area (in)
+        gy: y-flux of q in units of q * Pa * area (in)
+        rarea: 1 / area (in)
+    """
+    # TODO: this function changes the units and therefore meaning of q,
+    # is there any way we can void doing so?
```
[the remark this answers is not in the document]
elynnwu: The next time `w` and `qcon` are used after `flux_adjust` is `adjust_w_and_qcon`, and it first divides by `delp` to go back to the original unit. The problem is that `delp` does change in between. So if we want to have consistent names for the same unit, we will need to make extra storages to hold the intermediate data (q * delp) for `flux_adjust`.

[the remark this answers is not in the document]
elynnwu: Just occurred to me that the PR you just reviewed for physics also had the exact same pattern (` * delp` then ` / delp` later). So whatever we decide here, we should also be consistent on the physics side.

fv3core/stencils/d_sw.py
rheacangeo: I somewhat disagree, I don't think having dt folded in or not is more or less understandable, when it is documented what the variables are. It's super common in atmosphere science to move terms around in equations to help drop pieces off or solve them with different methods. It didn't seem impossible xtp_u/ytp_v could be stepping something forward in time, we just didn't have information on the intention of those subroutines, nor did we spend sufficient time exploring the context in which these subroutines were called when we did the first round of docstrings. But your in-depth investigation has helped us understand, so we do have the capability to document it in a way that makes sense.

mcgibbon:
I had a conversation with Lucas today and it cleared up a lot about what xppm/yppm are actually doing, and how fvtp2d uses them to determine a final flux. Commit ab179e5 has the fundamental core of these changes, without too much refactoring of the low-level routines used by xppm/yppm. I will do more refactors of xppm/yppm in a later commit/PR, since it seems nothing in them has anything to do with fluxes despite multiple variables like "flux" and "fx".

fv3core/stencils/fvtp2d.py
```diff
@@ -228,25 +241,17 @@ def __call__(
...
-                fx,
-                self._tmp_fx2,
-                fy,
-                self._tmp_fy2,
-                x_area_flux,
-                y_area_flux,
-            )
-            if self.delnflux is not None:
-                self.delnflux(q, fx, fy)
+        # up to here, fx and fy are in units of q * m^2
```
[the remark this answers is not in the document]
mcgibbon: As mentioned in a comment on the PR, this should be clearer now. A lot of the "flux" values were not actually fluxes, though the returned values from xppm/yppm are in fact in units of q (mean value of q in the advected area).

fv3core/stencils/fxadv.py
```diff
@@ -1,89 +1,107 @@
...
     rsin_u: FloatFieldIJ,
-    ut: FloatField,
+    uc_contra: FloatField,
 ):
+    """
+    Args:
+        uc: covariant c-grid x-wind (in)
+        vc: covariant c-grid y-wind (in)
+        cosa_u: ??? (in)
+        rsin_u: ??? (in)
```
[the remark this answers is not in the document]
mcgibbon: It is certainly hard to maintain copies of the descriptions for arguments used in many places, though I'm also nervous about not documenting arguments. I could make this easier by writing a linting test that reads the docstrings of all functions in `stencils/` and fails if any have the same arguments with different descriptions (ignoring the in/out at the end)? If we move to some kind of system where we're importing grid data instead of passing it as an argument, I think it might be more clear that you can look at some central place to see its description, and also less important for a coder to read if the grid data is automatically supplied (e.g. by our stencil wrapper).

Can we leave this however-it-gets-done for now, and unify the style after discussing what we want as a long term solution? This could happen as part of preparing the code to present at AMS, e.g. if we're going to set up Sphinx documentation.

fv3core/stencils/fxadv.py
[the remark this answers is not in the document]
mcgibbon: Added.

fv3core/stencils/xppm.py
```diff
@@ -17,12 +17,27 @@
 
 
 @gtscript.function
-def final_flux(courant, q, fx1, tmp):
-    return q[-1, 0, 0] + fx1 * tmp if courant > 0.0 else q + fx1 * tmp
+def apply_flux(courant, q, fx1, mask):
+    """
+    Args:
+        courant: any value whose sign is the same as the sign of
```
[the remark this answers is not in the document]
mcgibbon: This argument isn't really the courant number, it's any number (value) with the same sign as the x-wind on cell corners. It's used to determine wind direction only.

fv3core/stencils/xtp_u.py
```diff
@@ -16,82 +16,126 @@
 
 
 @gtscript.function
-def _get_flux(
+def advect_u(
```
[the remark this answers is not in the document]
mcgibbon: OK, did this refactor. Now bl_br are retrieved inside the nested function, and then since this function then becomes a one-liner calling another function, I deleted it and renamed advect_u to advect_u_along_x (commit b56f49cbb0a0ea4745c315438ca60014184eae65).

fv3core/stencils/xtp_u.py
mcgibbon: Added to comment.

fv3core/stencils/xtp_u.py
```diff
@@ -16,82 +16,126 @@
...
+        dt: timestep in seconds
+    """
+    # in the Fortran, dt is folded into ub_contra and called "courant"
+    bl, br = get_bl_br(u, dx, dxa)
+    # TODO: merge this function with advect_u by calling get_bl_br inside advect_u
+    updated_u = advect_u(u, ub_contra, rdx, bl, br, dt)
+    return updated_u
 
 
 class XTP_U:
```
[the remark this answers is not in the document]
mcgibbon: Moved XTP_U to the translate file (commit 098d46f1d4d2d9de83422c6acf175210e98258aa). I'll do the same for YTP_V along with all other xppm/xtp_u refactors when the reviews have converged

fv3core/stencils/xtp_u.py
```diff
@@ -115,7 +159,7 @@ def __init__(
         self._rdx = grid_data.rdx
         ax_offsets = axis_offsets(grid_indexing, origin, domain)
         self.stencil = FrozenStencil(
```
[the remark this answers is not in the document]
mcgibbon: Updated.

fv3core/stencils/xtp_u.py
```diff
@@ -131,7 +175,7 @@ def __call__(self, c: FloatField, u: FloatField, flux: FloatField):
         Compute flux of kinetic energy in x-dir.
 
         Args:
-            c (in): Courant number in flux form
+            c (in): product of x-dir wind on cell corners and timestep
```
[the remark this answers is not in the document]
mcgibbon: Yes, if I remember correctly this is commented in the Fortran code as being the "courant number in flux form". Nobody (including Lucas) could tell me what that meant, and it was one of my earliest questions about the code. I'm hoping this should be more clear, especially in adding that it's on cell corners which helps understand how this would be used to update the edge-center winds.

fv3core/stencils/ytp_v.py
```diff
@@ -16,79 +16,107 @@
...
+    dy: FloatFieldIJ,
+    dya: FloatFieldIJ,
+    rdy: FloatFieldIJ,
+    dt: float,
+):
+    """
+    Advect covariant y-wind on D-grid using contravariant y-wind on cell corners.
 
-            compile_assert(jord == 8)
+    Named xtp_u in the original Fortran code.
```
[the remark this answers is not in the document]
mcgibbon: At first I was doing updates to both the x and y routines but I stopped after a while, once we've finished reviewing the code via the xppm/xtp_u versions I'll completely re-update the y/v versions.

tests/savepoint/translate/translate_d_sw.py
```diff
@@ -66,7 +66,9 @@ def ubke(
     dt5: float,
 ):
     with computation(PARALLEL), interval(...):
-        ub = d_sw.ubke(uc, vc, cosa, rsina, ut, ub, dt4, dt5)
+        dt = 2.0 * dt5
+        ub, _ = d_sw.interpolate_uc_vc_to_cell_corners(uc, vc, cosa, rsina, ut, ut)
+        ub = ub * dt
```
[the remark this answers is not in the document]
mcgibbon: Yes, we want to make sure the tests are testing the stencil code, and not the test code. In this case the math isn't duplicated, it's only done here. This needed to be updated when I refactored ubke to take in only one timestep, and to return wind without the temperature folded in. The line before and after here are wrappers which restore the old behavior (since that's what the Fortran code provides data for).

fv3core/stencils/fvtp2d.py
```diff
@@ -33,13 +33,17 @@ def q_i_stencil(
...
-        fyy = y_area_flux * fy2
-        area_with_y_flux = apply_y_flux_divergence(area, y_area_flux)
-        q_i = (q * area + fyy - fyy[0, 1, 0]) / area_with_y_flux
+        fyy = y_area_flux * q_advected_along_y
+        # note the units of area cancel out, because area is present in all
+        # terms in the numerator and denominator of q_i
+        # corresponds to FV3 documentation eq 4.18, q_i = f(q)
+        q_i = (q * area + fyy - fyy[0, 1, 0]) / (
+            area + y_area_flux - y_area_flux[0, 1, 0]
+        )
```
rheacangeo, line 46: I'm glad this change in math still passes tests!

fv3core/stencils/fvtp2d.py
```diff
@@ -55,24 +59,36 @@ def q_j_stencil(
...
         with horizontal(region[:, :-1]):
-            fx = transport_flux(fx, fx2, mfx)
+            x_flux = (
+                0.5 * (q_advected_y_x_advected_mean + q_x_advected_mean) * x_unit_flux
+            )
         with horizontal(region[:-1, :]):
-            fy = transport_flux(fy, fy2, mfy)
+            y_flux = (
+                0.5 * (q_advected_x_y_advected_mean + q_y_advected_mean) * y_unit_flux
+            )
```
rheacangeo, line 91: this is a case where it makes sense to use a function (though albeit the name was probably not right). but you can leave as-is if this is more clear, it's not a huge difference either way

fv3core/stencils/fxadv.py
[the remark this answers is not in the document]
rheacangeo: ok sure, you can leave as-is for now.

fv3core/testing/validation.py
```diff
@@ -160,3 +169,11 @@ def enable_selective_validation():
         fv3core.stencils.tracer_2d_1l.TracerAdvection,
         get_compute_domain_k_interfaces,
     )
+
+    fv3core.stencils.divergence_damping.DivergenceDamping = get_selective_class(
+        fv3core.stencils.divergence_damping.DivergenceDamping,
+        {
+            "v_contra_dxc": get_domain_func([X_INTERFACE_DIM, Y_INTERFACE_DIM, Z_DIM]),
+            "vort": get_domain_func([X_INTERFACE_DIM, Y_INTERFACE_DIM, Z_DIM]),
+        },  # must include both function argument and savepoint names
```
rheacangeo, line 178: cool that you found another case we can selectively validate

tests/savepoint/translate/translate_d_sw.py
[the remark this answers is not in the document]
rheacangeo: ok true, it's not technically duplicated. I usually try to keep compute code out of the test area to make investigating errors easier, but with the changes made, we probably need to do this. so leave as-is to preserve the test functionality.

rheacangeo (approved):
I'm inclined to approve this, pending the updated to ytp matching xtp. huge effort! Great job. Might be a good idea to run the performance plan on this branch again, just to make sure there are no issues when running with c128.

fv3core/stencils/fvtp2d.py
mcgibbon: I'm good with this being a function if we can think of an explanatory name, but I don't think there is one. If I had to pick one I'd say "average y value instantaneously and after the x flux, and then multiply that average value by the unit weighting being advected to give the total amount of the scalar which is y-fluxed", but that's a bit of a mouthful.

mcgibbon:
Launched performance profile job https://jenkins.ginko.ch/job/fv3core-performance-profile/346/

mcgibbon:
launch jenkins

fv3core/stencils/d_sw.py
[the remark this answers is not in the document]
elynnwu: That makes sense, on the physics side the variables are divided/multiplied within the same stencil, so they can just be kept as temporaries.

FORTRAN_CHANGELOG.md
```diff
@@ -0,0 +1,15 @@
+Fortran Changelog
```
elynnwu, line 1: Thanks for including this changelog!

elynnwu (approved):

mcgibbon:
Submitted a new performance profile run after fixing the race condition failure on the last one: https://jenkins.ginko.ch/job/fv3core-performance-profile/347/

fv3core/stencils/d_sw.py
```diff
@@ -114,32 +184,39 @@ def pressure_and_vbke(
...
-                pt = pt / delp
-        vb = vbke(vc, uc, cosa, rsina, vt, vb, dt4, dt5)
-
-
-@gtscript.function
-def ke_from_bwind(ke, ub, vb):
-    return 0.5 * (ke + ub * vb)
+        # gx, gy read with 1-offset in respective direction
+        pt, delp = apply_pt_delp_fluxes(gx, gy, rarea, fx, fy, pt, delp)
+        # vb_contra = interpolate_vc_to_cell_corners(vc, uc, cosa, rsina, vt)
```
mcgibbon, line 204:
```suggestion
        pt, delp = apply_pt_delp_fluxes(gx, gy, rarea, fx, fy, pt, delp)
```

mcgibbon:
launch jenkins

mcgibbon:
launch jenkins

PR #611, opened 2021-09-29
Improve the translate tests

FlorianDeconinck:
## Purpose

Adds `--which_rank` in translate pytest and  verbose errors (adding the metric error)

## Code changes:

* `--which_rank`: only run one rank of the test (check all variables)
* Prints the metric error calculated in `compare_arr` which differs from an absolute error. Restrict print to 3 decimal in scientific notation

## Requirements changes:

N/A

## Infrastructure changes:

N/A

## Checklist

### DIFF
--- a/tests/savepoint/conftest.py
+++ b/tests/savepoint/conftest.py
@@ -168,6 +168,15 @@ def get_parallel_savepoint_names(metafunc, data_path):
     return parallel_names
 
 
+def get_ranks(metafunc, layout):
+    only_rank = metafunc.config.getoption("which_rank")
+    if only_rank is None:
+        total_ranks = 6 * layout[0] * layout[1]
+        return range(total_ranks)
+    else:
+        return [int(only_rank)]
+
+
 SavepointCase = collections.namedtuple(
     "SavepointCase",
     [
@@ -185,15 +194,13 @@ SavepointCase = collections.namedtuple(
 def sequential_savepoint_cases(metafunc, data_path):
     return_list = []
     layout = fv3core._config.namelist.layout
-    total_ranks = 6 * layout[0] * layout[1]
     savepoint_names = get_sequential_savepoint_names(metafunc, data_path)
+    ranks = get_ranks(metafunc, layout)
 
-    for rank in range(total_ranks):
+    for rank in ranks:
         serializer = get_serializer(data_path, rank)
         grid_savepoint = serializer.get_savepoint(GRID_SAVEPOINT_NAME)[0]
         grid = process_grid_savepoint(serializer, grid_savepoint, rank)
-        if rank == 0:
-            grid_rank0 = grid
         for test_name in sorted(list(savepoint_names)):
             input_savepoints = serializer.get_savepoint(f"{test_name}-In")
             output_savepoints = serializer.get_savepoint(f"{test_name}-Out")
@@ -209,7 +216,13 @@ def sequential_savepoint_cases(metafunc, data_path):
                     layout,
                 )
             )
+
+    # Set the grid to rank 0's data
+    serializer = get_serializer(data_path, 0)
+    grid_savepoint = serializer.get_savepoint(GRID_SAVEPOINT_NAME)[0]
+    grid_rank0 = process_grid_savepoint(serializer, grid_savepoint, 0)
     fv3core._config.set_grid(grid_rank0)
+
     return return_list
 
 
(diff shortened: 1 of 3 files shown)


ofuhrer:
Can one of the admins verify this patch?

tests/savepoint/conftest.py
```diff
@@ -209,7 +216,13 @@ def sequential_savepoint_cases(metafunc, data_path):
                     layout,
                 )
             )
+
+    # Set the grid to rank 0's data
+    serializer = get_serializer(data_path, 0)
+    grid_savepoint = serializer.get_savepoint(GRID_SAVEPOINT_NAME)[0]
+    grid_rank0 = process_grid_savepoint(serializer, grid_savepoint, 0)
     fv3core._config.set_grid(grid_rank0)
```
rheacangeo, line 224: You don't need to do this part if only running one rank. Basically this is done to reset the global state to rank 0 after processing all the ranks data so that when the test is run, you are using the correct grid. When we don't have global state, this will be removed anyways. But could actually cause some issues with 54 ranks and only testing one rank. For the moment, you could put this inside of `if len(ranks) > 1`.

tests/savepoint/conftest.py
```diff
@@ -401,8 +423,25 @@ def get_parallel_param(
...
+    savepoint_in_list,
+    savepoint_out_list,
+    call_count,
+    max_call_count,
+    only_one_rank,
 ):
+    dependency = (
+        pytest.mark.dependency()
+        if only_one_rank
+        else pytest.mark.dependency(
```
rheacangeo, line 437: nice haha!

rheacangeo, line 437: this might produce strange results with the mock parallel option. Most of these do halo updates (just in loops), so running only one rank will definitely give the wrong answer. But it's not a huge deal to leave it as is.

FlorianDeconinck:
launch jenkins

tests/savepoint/test_translate.py
```diff
@@ -24,12 +24,24 @@
...
     return compare
 
 
+def compare_scalar(computed_data: np.float64, ref_data: np.float64) -> np.float64:
+    """Smooth error near zero values. Scalar versions."""
+    denom = np.abs(ref_data) + np.abs(computed_data)
+    if denom == 0:
+        return np.float64(0.0)
+    else:
+        return 2.0 * np.abs(computed_data - ref_data) / denom
```
rheacangeo, line 42: rather than duplicating the method for scalars, you could alternatively turn the scalars into np arrays of length 1

rheacangeo (commented):
Awesome, thanks for working on this!

tests/savepoint/test_translate.py
FlorianDeconinck: I though about that and got a micro-optimization reflex to not have a local allocation :D I'll change it

tests/savepoint/test_translate.py
```diff
@@ -94,25 +106,46 @@ def sample_wherefail(
...
-    sample = [f[0] for f in found_indices]
-    fullcount = len(ref_data.flatten())
 
+    # Determine worst result
+    worst_err = 0
+    worst_idx = 0
+    worst_full_idx = 0
+    for b in range(0, bad_indices_count, failure_stride):
+        err = compare_scalar(computed_failures[b], reference_failures[b])
+        if worst_err < err:
```
elynnwu, line 130: (optional) you can also use `numpy.argmax` and `numpy.unravel_index` to find the largest error index instead of having a loop over the failures.
```
err =  compare_arr(computed_data, ref_data)
worst_err = err.max()
worst_full_idx = np. unravel_index(np.argmax(err, axis=None), err.shape)
```
I think you won't need the scalar version?

elynnwu (commented):
Thanks for adding this!

tests/savepoint/conftest.py
```diff
@@ -168,6 +168,15 @@ def get_parallel_savepoint_names(metafunc, data_path):
     return parallel_names
 
 
+def get_ranks(metafunc, layout):
+    only_rank = metafunc.config.getoption("which_rank")
+    if only_rank is None:
+        total_ranks = 6 * layout[0] * layout[1]
+        return range(total_ranks)
+    else:
+        return [int(only_rank)]
```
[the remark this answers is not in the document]
FlorianDeconinck: Yes only one rank for now, it makes it easy for the logic later on but it can be extended

tests/savepoint/conftest.py
FlorianDeconinck: I am going to raise a `RuntimeError` on mock_parallel and parallel if `--which_rank` is used

FlorianDeconinck:
launch jenkins

rheacangeo (approved):

PR #619, opened 2021-10-05
Reduce number of copy corners kernel calls for FiniteVolumeTransport

mcgibbon:
## Purpose

Corner copying consists of a significant portion of our total runtime (tens of percent). Many of these copies occur when calling FiniteVolumeAdvection in fvtp2d.py. This PR attempts to reduce total runtime by merging stencil calls for copy corners operations when FiniteVolumeAdvection is called, reducing these stencil calls by a factor of 3.

## Code changes:

- FiniteVolumeAdvection is refactored to take in a `CopiedCorners` dataclass for the variable to be advected, which provides attributes for x-differenceable and y-differenceable forms of the variable to be advected
- New class CopyCornersXY will copy both x and y corners in one stencil, and does not do a full copy of the storage as a separate stencil call, reducing stencil calls by a factor of 3
- New class PreallocatedCopyCornersFactory creates CopiedCorners instances using CopyCornersXY, and is now used wherever FiniteVolumeAdvection is used

## Checklist


fv3core/stencils/fvtp2d.py
```diff
@@ -75,6 +76,62 @@ def transport_flux_xy(
...
+        """
+        Args:
+            grid_indexing: information about grid size
+            dims: dimensionality of data to be copied
+            y_temporary: if given, storage to use for y-differenceable field
+                (x-differenceable field uses same memory as base storage),
+                if None then a storage is initialized based on max shape
+        """
+        if y_temporary is None:
+            y_temporary = utils.make_storage_from_shape(
```
eddie-c-davis, line 120: Could the `y_temporary` also use the same storage? Probably triple-aliasing is not a good idea, just curious.

fv3core/utils/corners.py
```diff
@@ -368,6 +425,124 @@ def copy_corners_y_stencil_defn(q_in: FloatField, q_out: FloatField):
             q_out = q_in[-3, -2, 0]
 
 
+def copy_corners_xy_stencil_defn(
```
eddie-c-davis, line 428: Nice stencil fusion!

fv3core/utils/corners.py
```diff
@@ -70,6 +72,61 @@ def __call__(self, field: FloatField):
...
+    """
+    Helper-class to copy corners corresponding to the Fortran functions
+    copy_corners_x and copy_corners_y
+    """
+
+    def __init__(
+        self,
+        grid_indexing: GridIndexing,
+        dims: Sequence[str],
+        y_field,
```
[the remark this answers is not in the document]
mcgibbon: I removed the type hint because I can't put Optional[FloatFIeld] because FloatField is not a real type. We talked about turning FloatField into a real type a while back, but haven't gotten around to it. I should update the docstring to convey more of the information that would be conveyed by this type hint.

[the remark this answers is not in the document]
mcgibbon: Updated.

fv3core/utils/corners.py
```diff
@@ -54,6 +56,61 @@ def __call__(self, field: FloatField):
...
+        Fills cell quantity field using corners from itself.
+
+        Args:
+            field: field to fill corners
+
+        Returns:
+            x_differenceable: input field, updated so it can be differenced
+                in x-direction
+            y_differenceable: copy of input field which can be differenced
+                in y-direction
```
rheacangeo, line 105: this is just copying the corners, how the resulting field is used after doesn't necessarily have to be tied to x and y being differenceable, though that is the purpose in fvtp2d

fv3core/utils/corners.py
```diff
@@ -352,6 +409,124 @@ def copy_corners_y_stencil_defn(q_in: FloatField, q_out: FloatField):
...
+):
+    from __externals__ import i_end, i_start, j_end, j_start
+
+    with computation(PARALLEL), interval(...):
+        q_out_x = q_in
+        q_out_y = q_in
+        with horizontal(
+            region[i_start - 3, j_start - 3], region[i_end + 3, j_start - 3]
+        ):
+            q_out_x = q_in[0, 5, 0]
```
rheacangeo, line 423: I thought we had found this pattern to not be ok/not work. But I see it was changed to this in the single x/y  variables so I'll follow up elsewhere

